PR #3, opened 2020-03-16
dependencies: remove deprecated extensions

vince-fugnitto:
Removes the `@theia/typescript` extension from the examples
which leads to a multi-injector error from inversify.

Signed-off-by: vince-fugnitto <vincent.fugnitto@ericsson.com>

### DIFF
--- a/browser-app/package.json
+++ b/browser-app/package.json
@@ -14,7 +14,6 @@
     "@theia/languages": "next",
     "@theia/markers": "next",
     "@theia/monaco": "next",
-    "@theia/typescript": "next",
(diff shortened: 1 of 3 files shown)


marcdumais-work:
@bhufmann  I do not remember if supporting TypeScript/JavaScript was important for you. If so we can add-back support using so-called `built-in VS Code extensions`. We have multiple examples if interested.

bhufmann:
Hi Marc

I don't need in the trace extension at the moment. Only if we support some scripting using JavaScript it might be interesting to add.

Thanks
Bernd

________________________________________

PR #6, opened 2020-03-24
timegraph-output-component: fix synchronization bug

chertyb:
Fix visible in trace explorer with analysis "Thread Status", entries are same height as lines and both components (tree and chart) scroll at the same time.
Works for Chrome but Firefox has resize problem. 
Also experienced line height problems (lines too big and blurry) with a few browsers but don't know the cause.
height-problem
normal-view

### DIFF
--- a/viewer-prototype/src/browser/style/output-components-style.css
+++ b/viewer-prototype/src/browser/style/output-components-style.css
@@ -46,6 +46,8 @@
 #timegraph-main {
     width: 100%;
     display: flex;
+    overflow-y: scroll;
+    position: relative;
 }
 
 #main {
--- a/viewer-prototype/src/browser/trace-viewer/components/abstract-tree-output-component.tsx
+++ b/viewer-prototype/src/browser/trace-viewer/components/abstract-tree-output-component.tsx
@@ -6,16 +6,34 @@ import { Entry, EntryHeader } from 'tsp-typescript-client/lib/models/entry';
 
 
 export abstract class AbstractTreeOutputComponent<P extends AbstractOutputProps, S extends AbstractOutputState> extends AbstractOutputComponent<P, S> {
+    protected treeRef = React.createRef<any>();
+    protected chartRef = React.createRef<any>();
+    
     renderMainArea(): React.ReactNode {
         const treeWidth = this.props.style.width - this.props.style.chartWidth - this.getHandleWidth();
         return <React.Fragment>
-            <div className='output-component-tree' style={{ width: treeWidth, height: this.props.style.height }}>
+            <div 
+                ref={this.treeRef} 
+                className='output-component-tree' 
+                style={{ width: treeWidth, height: this.props.style.height }}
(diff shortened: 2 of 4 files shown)


tahini:
Just adding myself to this review. With this patch, the scroll now works. Thanks!

delislesim:
Using `TimeGraphRowController` or something like that from the timeline-chart library might be an other option since the timegraph canvas is using his own vertical scrollbar

chertyb:
Updated commit message and signed-off

chertyb:
@delislesim We looked at timegraphRowController but weren't able to get scroll value from that and it seems that the vertical scroll bar is never called (getVerticalScrollbar() in viewer-prototype/src/browser/traceviewer/components/timegraph-output-component.tsx)

bhufmann:
Hi Lea, Thanks for updating the patch. I tried the updates and for the kernel Thread status view the height of the states are not correct. See attached picture which is similar that you pointed out earlier. I happens on chrome on Linux (81.0.4044.138). Please note that the line heights are correct when opening the same trace without this patch. Could you please look into this?

image

tahini:
I have the exact same browser as you Bernd, but I can't reproduce this bug on my side. I don't know what could be the problem here. A linux/Windows thing? But Léa was on Windows and had no problem iirc...

On the other hand, this patch makes the time graph at least minimally usable for most cases, so either we could accept this patch with some problematic browsers, or keep the current not scrollable implementation.

bhufmann:
Hi Genvieve

I'm running on Linux. I tried with 3 different traces and all are the same. Chrome is the default browser and it should work with it. Lea also experienced this issue with Firefox. So it sounds like an issue.

I'd like what the patch fixes and I'm looking forward having it in. But I think we should investigate and understand what is going on. Then find a solution before merging this.

delislesim:
Jumping in again, @bhufmann @tahini it might be interesting to investigate a bit further on the scrolling mechanism that is used in the timegraph library. This patch is adding an overflow scroll to a canvas and modify the behaviour of catching the scroll event. The rendered canvas should never "overflow" that means there is a problem in the way the canvas is drawn. I think right now the drawing is not optimized and render all the row in a giant canvas. At some point this will cause problem since the library might not receive the scroll event to fetch more data.The scroll should happen in the canvas not scrolling the canvas.

PS: This might not be the case, so don't take this as the only source of truth 😛

bhufmann:
Thanks @delislesim for the feedback. It sounds reasonable. 
I investigated a bit. I'm slowly getting some understanding. I noticed that the vscrollbar (TimeGraphVerticalScrollbar) is not hooked up properly in timegraph-output-component and hence it's not rendered at the moment. I would suggest, first, to render the scrollbar, second have a scrollbar component in the tree and then synchronize both.

bhufmann:
Any news on this patch?

tahini:
Closing this PR as there's another one for the same topic

PR #10, opened 2020-04-09
Add support of experiments consisting of 1 or more traces

bhufmann:
- Added support of opening single and multiple traces in an experiment
- Upload each trace and create experiment using ExperimentManager
- Implemented simple trace detection algorithm. If the selection is a
directory, then the files system is recursively searched for CTF traces
by finding the metadata file. Otherwise a single file is opened as trace
- The detection of traces is done in the front-end. This needs to be
replaced by a backend service that avoids querying the file system
directly from the frontend.
- Updated class, method and variable names to reflect support of
experiments
- Update Trace Explorer to show available experiments and it's traces.
This visualized list is very basic and needs to be improved.

### DIFF
--- a/viewer-prototype/package.json
+++ b/viewer-prototype/package.json
@@ -17,6 +17,7 @@
   "dependencies": {
     "@theia/core": "next",
     "@theia/editor": "next",
+    "@theia/filesystem": "next",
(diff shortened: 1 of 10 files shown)


MatthewKhouzam:
Tested, it works

PR #63, opened 2020-05-25
Handle conflicting experiments when opening

tahini:
When opening a trace with the same name as another experiment, the experiment
that contains it will have a conflict on the server. The client should
retry posting it with a different name until it succeeds or another
error occurs.

The title of the trace viewer should also be updated to represent the
actual name of the experiment.

### DIFF
--- a/viewer-prototype/src/common/experiment-manager.ts
+++ b/viewer-prototype/src/common/experiment-manager.ts
@@ -5,6 +5,7 @@ import { Query } from 'tsp-typescript-client/lib/models/query/query';
 import { injectable, inject } from 'inversify';
 import { OutputDescriptor } from 'tsp-typescript-client/lib/models/output-descriptor';
 import { Experiment } from 'tsp-typescript-client/lib/models/experiment';
+import { TspClientResponse } from 'tsp-typescript-client/lib/protocol/tsp-client-response';
 
 @injectable()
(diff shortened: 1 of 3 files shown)


tahini:
This goes with Trace server commit: https://git.eclipse.org/r/#/c/163272/

viewer-prototype/src/common/experiment-manager.ts
```diff
@@ -87,11 +88,32 @@ export class ExperimentManager {
...
+                let suffix = '(' + tryNb + ')';
+                return await tspClient.createExperiment(new Query({
+                    'name': name + suffix,
+                    'traces': traceURIs
+                }))
+            }
+            let conflictResolutionResponse = experimentResponse;
+            let i = 1;
+            while (conflictResolutionResponse.getStatusCode() === 409) {
+                conflictResolutionResponse = await handleConflict(this.tspClient, i);
```
tahini, line 107: forgot i++

bhufmann: Ahh I missed that. I only tried having one conflict. I'm waiting for an update.

tahini:
I updated the patch, and also added the management of the trace conflict, even though there is no such conflicts anymore from the trace server, but if we add some, it will be handled here.

PR #66, opened 2020-06-01
xy chart: checkbox tree view

chertyb:


bhufmann:
Thanks for this pull request. I started with testing the pull request and it's a nice improvement. I'm wondering, is there an open source react tree (with columns) that we could just use? Of course implementing our own tree, we have the freedom on design it in a way that it's most beneficial for this project's use cases. 

When testing it though, I realized that the parent selection state can be get out of sync with the children states, e.g. the parent is unselected and some children are selected and some are not (or vice versa). I think that there should be a third state for nodes with children, partially selected. With this you can select a parent node to select all, and unselect some children. The checkbox should also indicate that it's partially selected. Any change of selection, basically all the parents have to be checked and updated, moreover the tree traversed to see if parent should be checked, uncheck or partially. 

Another side note, The tree component should be eventually outside the theia-trace-extension, like the timeline chart. However, this is not part of the scope of this pull request

viewer-prototype/src/browser/trace-viewer/components/utils/filter-tree-component.tsx
```diff
@@ -0,0 +1,230 @@
...
+            checkedIDs: nextProps.checkedSeries
+        })
+    }
+
+    getRootNodes = () => {
+        const nodes = this.state.treeNodes;
+        return nodes.filter((node: TreeNode) => node.isRoot === true);
+    }
+
+    getChildNodes = (node: TreeNode) => {
```
tahini, line 44: While this is a perfectly legit typescript statement, I wonder if there should be a return type somewhere, like getChildNodes: something?

viewer-prototype/src/browser/trace-viewer/components/utils/filter-tree-component.tsx
```diff
@@ -0,0 +1,230 @@
...
+        if (!node.children) return [];
+        return node.children;
+    }
+
+    getNode: any = (node: TreeNode, id: number) => {
```
tahini, line 49: Is the return type really any, or Node | undefined?

viewer-prototype/src/browser/trace-viewer/components/utils/filter-tree-component.tsx
```diff
@@ -0,0 +1,230 @@
...
+                result = this.getNode(node.children[i], id);
+            }
+            return result;
+        }
+        return null;
+    }
+
+    onToggle = (node: TreeNode) => {
+        const nodes = this.state.treeNodes;
+        const toggledNode = this.getNode(nodes[0], node.id);
```
tahini, line 64: This should probably be a for loop, when we have experiments, there can be multiple root nodes

Also, you should maybe call a setState on some components if you really want it to work?

viewer-prototype/src/browser/trace-viewer/components/utils/filter-tree-component.tsx
```diff
@@ -0,0 +1,230 @@
...
+                                checkedSeries={this.props.checkedSeries}
+                                onChecked={this.handleCheck}
+                            />
+                        )
+                        
+                    })}
+                </div>
+            )
+        }
+        return (<div></div>)
```
tahini, line 126: Here, you could return <EmptyTree> and create a component for it so we can edit it later when we want something specific

viewer-prototype/src/browser/trace-viewer/components/utils/filter-tree-component.tsx
```diff
@@ -0,0 +1,230 @@
...
+}
+
+class TreeNodeComponent extends React.Component<TreeNodeComponentProps> {
+    constructor(props: TreeNodeComponentProps) {
+        super(props);
+    }
+
+    getPaddingLeft = (level: number, type: string) => {
+        let paddingLeft = level * 15;
+        if (type === 'leaf') paddingLeft += 15;
```
tahini, line 146: type should be either an enum or extracted to some constants. Also is it necessary or does the child count give the answer if it's a leaf or a node?

viewer-prototype/src/browser/trace-viewer/components/utils/filter-tree-component.tsx
```diff
@@ -0,0 +1,230 @@
...
+        return paddingLeft;
+    }
+
+    render() {
+        const paddingLeft = this.getPaddingLeft(this.props.level, this.props.node.type);
```
tahini, line 151: Could padding  come from the parent as a prop? To avoid recalculation at each render

viewer-prototype/src/browser/trace-viewer/components/utils/checkbox-component.tsx
```diff
@@ -4,34 +4,18 @@ type CheckboxProps = {
...
-    onChecked: (id: number) => void
+    onChecked: (id: number) => void;
 }
 
-type CheckboxState = {
-    checked: boolean;
-}
-
-export class CheckboxComponent extends React.Component<CheckboxProps, CheckboxState> {
+export class CheckboxComponent extends React.Component<CheckboxProps> {
```
tahini, line 10: If this file is not used anywhere else, you may want to move it in the filter-tree folder

viewer-prototype/src/browser/trace-viewer/components/utils/filtrer-tree/indent.tsx
```diff
@@ -0,0 +1,31 @@
...
+export class Indent extends React.Component<IndentProps> {
+    constructor(props: IndentProps) {
+        super(props);
+    }
+
+    render() {
+        if (!this.props.level) return null;
+        
+        let { level } = this.props;
+        level = this.props.isLastLevel ? level+1 : level;
```
tahini, line 17: Why a +1 for the last level?

viewer-prototype/src/browser/trace-viewer/components/utils/filtrer-tree/tree-node.tsx
```diff
@@ -0,0 +1,70 @@
...
+    children: [];
+    isRoot: boolean;
+}
+
+type TreeNodeComponentProps = {
+    node: TreeNode;
+    level: number;
+    isCheckable: boolean;
+    checkedSeries: number[];
+    getChildNodes: (node: TreeNode) => TreeNode[];
```
tahini, line 22: Do you need the getChildNodes() method or can you just use node.children whenever required?

viewer-prototype/src/browser/trace-viewer/components/utils/filtrer-tree/tree-node.tsx
```diff
@@ -0,0 +1,70 @@
...
+                            id={this.props.node.id}
+                            name={this.props.node.name}
+                            checked={this.props.checkedSeries.includes(this.props.node.id) || false}
+                            onChecked={this.props.onChecked}
+                        />
+                        : <span style={{marginLeft: 5}}>{this.props.node.name}</span> 
+                    } 
+                </div>
+    
+                {this.props.node.expanded && this.props.getChildNodes(this.props.node).map((childNode: any) => {
```
tahini, line 59: any -> TreeNode?

viewer-prototype/src/browser/trace-viewer/components/utils/filtrer-tree/xy-tree.tsx
```diff
@@ -0,0 +1,29 @@
+import * as React from 'react';
+import { Entry } from 'tsp-typescript-client/lib/models/entry';
+import { listToTree } from './utils';
+import { FilterTree } from './tree'
+
+type XYTreeProps = {
+    nodes: Entry[];
```
tahini, line 7: name this entries to not confuse with the nodes of the other components

viewer-prototype/src/browser/trace-viewer/components/utils/filtrer-tree/indent.tsx
chertyb: to have an extra indent since there is no arrow icon

viewer-prototype/src/browser/trace-viewer/components/utils/filtrer-tree/tree.tsx
```diff
@@ -0,0 +1,186 @@
...
+    }
+    
+    constructor(props: FilterTreeProps) {
+        super(props);
+        this.state = {
+            treeNodes: this.props.nodes,
+        }
+    }
+
+    static getDerivedStateFromProps = (nextProps: FilterTreeProps, prevState: FilterTreeState) => {
```
tahini, line 29: So this function is really necessary? why?

From this blog post, it appears it should be used sparingly https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html 
So if you think this is one of the use cases where it is needed, maybe you can add a comment to explain why.

viewer-prototype/src/browser/trace-viewer/components/utils/filtrer-tree/tree.tsx
```diff
@@ -0,0 +1,186 @@
+import * as React from 'react';
+import { TreeNode, TreeNodeComponent, defaultTreeNode } from './tree-node';
+import { Message } from './message';
+
+type FilterTreeProps = {
+    nodes: TreeNode[];
+    showCheckboxes: boolean;    
+    checkedSeries: number[];                //Optional
```
tahini, line 8: Should this be a prop or a state? just asking...

viewer-prototype/src/browser/trace-viewer/components/utils/filtrer-tree/tree.tsx
```diff
@@ -0,0 +1,186 @@
...
+        });
+        return (
+            <React.Fragment>
+                {treeNodes}
+            </React.Fragment>
+        )
+    }
+
+    render() {
+        if (!this.props.nodes || !this.props.checkedSeries) return null;
```
tahini, line 178: You said checkedSeries should be optional above, this check is making it mandatory?

viewer-prototype/src/browser/trace-viewer/components/xy-output-component.tsx
```diff
@@ -63,15 +63,11 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
-                return <CheckboxComponent key={entry.id}
-                    id={entry.id}
-                    name={entry.labels[0]}
-                    checked={this.state.checkedSeries.find(id => entry.id === id) ? true : false}
-                    onChecked={this.onSeriesChecked} />
-            })}
-        </React.Fragment>;
+        return <XYTree
+            entries = {this.state.XYTree }
+            checkedSeries={this.state.checkedSeries}
```
tahini, line 68: Indeed, if it's part of the state of the caller, it should be in the props of the tree.

viewer-prototype/src/browser/trace-viewer/components/utils/filtrer-tree/tree.tsx
```diff
@@ -0,0 +1,186 @@
...
+
+type FilterTreeProps = {
+    nodes: TreeNode[];
+    showCheckboxes: boolean;    
+    checkedSeries: number[];                //Optional
+    onChecked: (ids: number[]) => void;     //Optional
+}
+
+type FilterTreeState = {
+    treeNodes: TreeNode[];
```
tahini, line 13: From what  I understand, the part of the TreeNode that is really part of the state of the tree vs the parent is the expanded state, the checked one is managed by the parent, is my understanding correct?

viewer-prototype/src/browser/trace-viewer/components/utils/filtrer-tree/tree.tsx
babyadoresorange: As discussed, the solution to just keep the Expand in FilterTree will work. But imho, I think it's better to keep it in the same place with the tree itself, since the whole thing is a state, it makes sense for them to be together :)

In that case, the function to update Expand will need to be held the parent component. That function will later be passed as props to FilterTree for updating.

tahini:
Thanks for all the fixes. IMHO, this is ready to be merged. It is a great improvement and it works well.

As future work for later pull requests:

1- Fix the todo for the state, to avoid using the getDerivedStateFromProps method, that is for rare use cases only.

2- For the items, use a ul-li syntax instead of multiple indentation spans.

@bhufmann Do you want to review this?

bhufmann:
I tried a several XY plots and it work well. Thank you for the patch. Thanks @tahini for the thorough review.

PR #72, opened 2020-06-02
Add warning when trace server is offline

tahini:

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -15,7 +15,7 @@
                 "viewer-prototype:///js/*": "${workspaceRoot}/viewer-prototype/lib/**/*.js"
             },
             "webRoot": "${workspaceRoot}",
-            "url":"http://localhost:3000"
+            "url": "http://localhost:3000"
         },
         {
             "type": "chrome",
@@ -28,6 +28,20 @@
                 "viewer-prototype:///js/*": "${workspaceRoot}/viewer-prototype/lib/**/*.js"
             },
             "webRoot": "${workspaceRoot}"
+        },
+        {
+            "type": "firefox",
+            "request": "launch",
+            "name": "Start in Firefox",
+            "reAttach": true,
+            "webRoot": "${workspaceRoot}",
+            "url": "http://localhost:3000",
+            "pathMappings": [
+                {
+                    "url": "webpack:///viewer-prototype",
+                    "path": "${workspaceFolder}/viewer-prototype"
+                }
+            ]
         },
         {
             "type": "node",
@@ -75,7 +89,7 @@
                 "${workspaceRoot}/electron-app/src-gen/frontend/electron-main.js",
                 "${workspaceRoot}/electron-app/src-gen/backend/main.js",
                 "${workspaceRoot}/electron-app/lib/**/*.js",
-                "${workspaceRoot}/node_modules/@theia/*/lib/**/*.js",
+                "${workspaceRoot}/node_modules/@theia/*/lib/**/*.js"
             ],
             "smartStep": true,
             "internalConsoleOptions": "openOnSessionStart",
--- a/viewer-prototype/src/browser/style/status-bar.css
+++ b/viewer-prototype/src/browser/style/status-bar.css
@@ -1,0 +1,22 @@
+/*
+ * SPDX-License-Identifier: MIT
+ *
+ * Copyright (C) 2020 École Polytechnique de montréal
+ */
+
+
+.traceserver-mod-offline #theia-statusBar {
+    background-color: var(--theia-statusBar-offlineBackground) !important;
+}
+
+.traceserver-mod-offline #theia-statusBar .area .element {
+    color: var(--theia-statusBar-offlineForeground) !important;
+}
+
+.traceserver-mod-offline #theia-statusBar .area .element.hasCommand:hover {
+    background-color: var(--theia-statusBarItem-offlineHoverBackground) !important;
+}
+
+.traceserver-mod-offline #theia-statusBar .area .element.hasCommand:active {
+    background-color: var(--theia-statusBarItem-offlineActiveBackground) !important;
+}
--- a/viewer-prototype/src/browser/trace-explorer/trace-explorer-widget.tsx
+++ b/viewer-prototype/src/browser/trace-explorer/trace-explorer-widget.tsx
@@ -42,6 +42,7 @@ export class TraceExplorerWidget extends ReactWidget {
 
     private openedExperiments: Array<Experiment> = new Array();
     private selectedExperimentIndex: number = 0;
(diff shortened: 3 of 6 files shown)


viewer-prototype/src/browser/trace-server-status.ts
```diff
@@ -0,0 +1,101 @@
...
+        }
+        this.scheduledPing = this.setTimeout(async () => {
+            try {
+                await this.tspClient.fetchExperiments();
+                this.updateStatus(true);
+            } catch (e) {
+                this.logger.trace(e);
+            }
+            this.schedulePing();
+        }, this.options.offlineTimeout * 0.8);
```
tahini, line 53: This means a call every like 4.5 seconds... Maybe too much? this.options.offlineTimeout *2.0 maybe?

.vscode/launch.json
```diff
@@ -29,6 +29,20 @@
             },
             "webRoot": "${workspaceRoot}"
         },
+        {
+            "type": "firefox",
```
bhufmann, line 33: Had to install firefox debugger. Maybe it would be good to have a section in the README to describe on how to setup development environment, debug etc. (not in this patch)

.vscode/launch.json
```diff
@@ -29,6 +29,20 @@
...
+            "request": "launch",
+            "name": "Start in firefox",
```
bhufmann, line 35: Firefox instead of firefox

viewer-prototype/src/browser/style/status-bar.css
```diff
@@ -0,0 +1,31 @@
...
+    color: var(--theia-statusBar-offlineForeground) !important;
+}
+
+.traceserver-mod-offline #theia-statusBar .area .element.hasCommand:hover {
+    background-color: var(--theia-statusBarItem-offlineHoverBackground) !important;
+}
+
+.traceserver-mod-offline #theia-statusBar .area .element.hasCommand:active {
+    background-color: var(--theia-statusBarItem-offlineActiveBackground) !important;
+}
```
bhufmann, line 31: I get a warning about no newline at the end

viewer-prototype/src/browser/trace-server-status.ts
```diff
@@ -0,0 +1,101 @@
...
+    ) {
+        super();
+    }
+
+    @postConstruct()
+    protected init(): void {
+        this.schedulePing();
+    }
+
+    protected schedulePing(): void {
```
bhufmann, line 41: Detecting lost connection did work for me when I stopped the server. But after restarting the server, then it didn't detect that the server is running again. I had to click into the "available analysis" area so that it got deblocked. Not sure what is the reason. I debugged a bit and added breakpoints in the methods below, but the breakpoint didn't hit until me clicking in the available analysis area.

viewer-prototype/src/browser/trace-explorer/trace-explorer-widget.tsx
```diff
@@ -315,7 +317,11 @@ export class TraceExplorerWidget extends ReactWidget {
                 outputDescription = outputDescriptors[props.index].description;
             }
         }
-        return <div className='outputs-list-container' key={props.key} style={props.style} onClick={this.outputClicked.bind(this, props.index)}>
+        let traceContainerClassName = 'outputs-list-container';
+        if (props.index == this.lastSelectedOutputIndex) {
+            traceContainerClassName = traceContainerClassName + ' theia-mod-selected';
+        }
+        return <div className={traceContainerClassName} key={props.key} style={props.style} onClick={this.outputClicked.bind(this, props.index)}>
```
bhufmann, line 324: I wonder if we should synchronize the graph with the selection (like link with editor in Eclipse). I miss that in Trace Compass mainline. Maybe not part of this patch.

viewer-prototype/src/browser/style/status-bar.css
```diff
@@ -0,0 +1,31 @@
...
+ * terms of the Eclipse Public License v. 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the Eclipse
+ * Public License v. 2.0 are satisfied: GNU General Public License, version 2
+ * with the GNU Classpath Exception which is available at
+ * https://www.gnu.org/software/classpath/license.html.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
```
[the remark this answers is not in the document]
tahini: Is it staying that way? Given that theia is EPL 2 (I copied those files originally from Theia)

viewer-prototype/src/browser/trace-server-status.ts
tahini: I got the same with the Theia server connection status (I tried all combinations of both server going offline), and only when I got back to the file explorer did the Theia server go back online... It's as if once offline, the ping doesn't get called again until the service is "required" by a view. I don't know where this behavior comes from. If you have exact steps to reproduce it always, please let me know, I'll try to figure it out in the meantime.

viewer-prototype/src/browser/trace-explorer/trace-explorer-widget.tsx
tahini: What do you mean by graph? See this issue here, I think it covers various use cases. That will be for another patch though. It's another issue.

https://github.com/theia-ide/theia-trace-extension/issues/73

viewer-prototype/src/browser/trace-server-status.ts
bhufmann: It's already an improvement and gives feedback to the user.

viewer-prototype/src/browser/trace-explorer/trace-explorer-widget.tsx
bhufmann: I meant selecting a graph (e.g. CPU usage) on the right, should it then select the corresponding analysis in the list of available analysis. Yes, that could be another issue.

viewer-prototype/src/browser/trace-server-status.ts
tahini: *2.0 makes the trace server unavailable one in 2 requests, I guess it compares with the timeout. Having a ping on the server will make make it more efficient

viewer-prototype/src/browser/trace-server-status.ts
tahini: Leaving it as is for now, I'll investigate if/when I can reproduce with more consistency

tahini: No wait, there's room for improvement here...

tahini:
Shall I split this PR, I'd like to get the firefox debug option in ;-)

PR #77, opened 2020-06-10
Tooltips as overlay boxes when clicking on a row element in the chart

smouss:
In timegraph-output-component.tsx :
 - Created states to check if the element is clicked and if the tooltip is opened;
 - In the renderChart(), the tooltip is rendered only if the previous states are true;
 - The tooltip content is automatically updated when clicking on an element.
 - The tooltip box is closed when clicking on the close (x) button.

The attached video shows how the features works.

[tooltipsFeature.zip](https://github.com/theia-ide/theia-trace-extension/files/4756434/tooltipsFeature.zip)

### DIFF
--- a/viewer-prototype/src/browser/trace-viewer/components/timegraph-output-component.tsx
+++ b/viewer-prototype/src/browser/trace-viewer/components/timegraph-output-component.tsx
@@ -18,6 +18,7 @@ import { StyleProvider } from './data-providers/style-provider';
 import { TspDataProvider } from './data-providers/tsp-data-provider';
 import { ReactTimeGraphContainer } from "./utils/timegraph-container-component";
 import { OutputElementStyle } from 'tsp-typescript-client/lib/models/styles';
+import { TooltipComponent } from './utils/tooltip-component'
 
 type TimegraphOutputProps = AbstractOutputProps & {
     addWidgetResizeHandler: (handler: () => void) => void;
@@ -25,6 +26,8 @@ type TimegraphOutputProps = AbstractOutputProps & {
 
 type TimegraohOutputState = AbstractOutputState & {
     timegraphTree: TimeGraphEntry[];
+    isElementSelected: boolean;
+    isTooltipOpened: boolean;
 }
 
 export class TimegraphOutputComponent extends AbstractTreeOutputComponent<TimegraphOutputProps, TimegraohOutputState> {
@@ -33,18 +36,22 @@ export class TimegraphOutputComponent extends AbstractTreeOutputComponent<Timegr
     private chartLayer: TimeGraphChart;
     private vscrollLayer: TimeGraphVerticalScrollbar;
     private horizontalContainer: React.RefObject<HTMLDivElement>;
-
+    private toolTipObject: { [key: string]: string } = {};
     private tspDataProvider: TspDataProvider;
     private styleMap = new Map<string, TimeGraphRowElementStyle>();
-
     private selectedElement: TimeGraphRowElement | undefined;
 
     constructor(props: TimegraphOutputProps) {
         super(props);
         this.state = {
             outputStatus: ResponseStatus.RUNNING,
-            timegraphTree: []
+            timegraphTree: [],
+            isElementSelected: false,
+            isTooltipOpened: false,
         };
+
+        this.closeTooltip = this.closeTooltip.bind(this);
+        this.openTooltip = this.openTooltip.bind(this);
         this.tspDataProvider = new TspDataProvider(this.props.tspClient, this.props.traceId, this.props.outputDescriptor.id);
         this.rowController = new TimeGraphRowController(this.props.style.rowHeight, this.totalHeight);
         this.horizontalContainer = React.createRef();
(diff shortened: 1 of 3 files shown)


viewer-prototype/src/browser/trace-viewer/components/utils/timegraph-container-component.tsx
```diff
@@ -29,6 +29,7 @@ export class ReactTimeGraphContainer extends React.Component<ReactTimeGraphConta
     }
 
     render() {
-        return <canvas ref={ref => this.ref = ref || undefined} onWheel={e => e.preventDefault()}></canvas>
+        /*return <canvas id="canvas" ref={ref => this.ref = ref || undefined} onScroll={ev => this.props.scrollHandler(this.ref, ev.persist()) } ></canvas>*/
+        return <canvas id="canvas" ref={ref => this.ref = ref || undefined} ></canvas>
```
tahini, line 33: Remove the id, why change this line?

viewer-prototype/src/browser/trace-viewer/components/timegraph-output-component.tsx
```diff
@@ -104,30 +148,69 @@ export class TimegraphOutputComponent extends AbstractTreeOutputComponent<Timegr
...
+
             {this.state.outputStatus === ResponseStatus.COMPLETED ?
-                <div id='timegraph-main' className='ps__child--consume' onWheel={ev => { ev.preventDefault(); ev.stopPropagation(); }} >
+                <div id='timegraph-main' className='ps__child--consume' onWheel={ev => { ev.preventDefault(); ev.stopPropagation(); }}>
                     {this.renderTimeGraphContent()}
+                    {
+                        this.state.isClicked && this.state.isTooltipOpened &&
+                        <div id='tooltip-box' ref={this.tooltipRef} >
+                            <button id='close' onClick={ev => { this.closeTooltip() }}>x</button>
+                            {this.renderTooltip()}
```
tahini, line 172: I would put the tooltip as its own react component. I think there's nothing there specific to the time graph, so the XY chart's tooltip could use the same tooltip component. So you could create a new react component, TooltipBox and send it props like `<TooltipBox onClose={this.closeTooltip} tooltip=this.TooltipObject/>`

From what I understand of react, we must not hesitate to have as many small component classes as required. And I think the tooltip one can make for an easy-ish stateless and re-usable function component. 

Here's a simple article explaining how to make a typescript functional react component: https://medium.com/@ethan_ikt/react-stateless-functional-component-with-typescript-ce5043466011 Maybe it does not apply directly, but there should be tons of examples out there.

viewer-prototype/src/browser/trace-viewer/components/timegraph-output-component.tsx
```diff
@@ -168,7 +251,8 @@ export class TimegraphOutputComponent extends AbstractTreeOutputComponent<Timegr
         ></ReactTimeGraphContainer>;
     }
 
-    private async onElementSelected(element: TimeGraphRowElement | undefined) {
+
+    async getToolTipObject(element: TimeGraphRowElement | undefined): Promise<any> {
```
tahini, line 255: should it be private?

viewer-prototype/src/browser/trace-viewer/components/timegraph-output-component.tsx
```diff
@@ -19,20 +19,29 @@ import { TspDataProvider } from './data-providers/tsp-data-provider';
...
 }
 
 export class TimegraphOutputComponent extends AbstractTreeOutputComponent<TimegraphOutputProps, TimegraohOutputState> {
+    [x: string]: any;
     private totalHeight: number = 0;
     private rowController: TimeGraphRowController;
     private chartLayer: TimeGraphChart;
     private vscrollLayer: TimeGraphVerticalScrollbar;
     private horizontalContainer: React.RefObject<HTMLDivElement>;
+    private toolTipObject: { [key: string]: string } = {};
```
tahini, line 42: Here I'm not sure, but should the toolTipObject be part of the state? Maybe a more react expert can tell, but fwiu, anything mutable that has effect on how the component will be rendered should be part of the state. And calling setState with the new tooltip set wil automatically re-render the component, so you don't have to call forceUpdate after setting this field, just setting the state does it. But I may be wrong...

viewer-prototype/src/browser/trace-viewer/components/timegraph-output-component.tsx
```diff
@@ -76,12 +92,40 @@ export class TimegraphOutputComponent extends AbstractTreeOutputComponent<Timegr
...
+        }
+    }
+
+    toggleCloseTooltip() {
+        return {
+            isTooltipOpened: false,
+        }
+    }
+
+
```
tahini, line 124: Remove extra blank lines, here and below

tahini:
Functionality-wise, it works well, the right tooltip is shown. And the fact that the box is fixed in one location and does not disappear as the mouse moves is not a bad thing! I'd let the UX expert tell us how it should behave, but leave it as is for now.

smouss:
> I don't want to slow down the work here, I just want to say that `TimeGraphChart` (`chartLayer` in this case) has a `registerRowElementMouseInteractions` that can be use to register mouseover. I think you just need to pass an handler like:
> 
> ```ts
> chartLayer.registerRowElementMouseInteractions({
>     mouseover: el => {
>         // Do tooltip related calls
>     }
> });
> ```

smouss:
@delislesim Thank you for your tip. I'm going to try this out!

tahini:
Will you be looking at Simon's suggestion in this PR? Otherwise, please look at the new methods you added/changed and make sure that if they can be private, then they are private.

viewer-prototype/src/browser/trace-viewer/components/utils/tooltip-component.tsx
```diff
@@ -0,0 +1,38 @@
...
+        }
+        return <React.Fragment>
+            {tooltipArray.map(element => {
+                return element;
+            })}
+        </React.Fragment>
+    }
+
+    render() {
+        return <div id='tooltip-box' ref={this.tooltipRef}>
```
tahini, line 34: Avoid using a div ID, as it implies there is one and only one such div in the whole application, now and forever. Using a class would be better.

tahini:
I'll close and open a new rebased PR from my repo

PR #88, opened 2020-06-22
Add linter support to the trace viewer prototype

tahini:


tahini:
@marechal-p  @marcdumais-work  I'm starting to add linter support. I only auto-fixed for now, will work on it again later this week, but could you take a look at the configuration please? 

I took a bit from the theia project and from this post: https://www.robertcooper.me/using-eslint-and-prettier-in-a-typescript-project

I may add prettier eventually too. Do you use it in theia?

marcdumais-work:
cc: @vince-fugnitto

marcdumais-work:
> @marechal-p @marcdumais-work I'm starting to add linter support. I only auto-fixed for now, will work on it again later this week, but could you take a look at the configuration please?
> 

Yes. @vince-fugnitto @marechal-p  I think you one of you updated the main repo to go from `TSLint` to `ESLint`, not that long ago? And I think @vince-fugnitto that you added similar configs when we split the `cpp` extensions in their own repo? Can one of you do a preliminary review? 

> I may add prettier eventually too. Do you use it in theia?

I do not think so, but sounds cool. Give it a try if you want for sure.

tahini:
Turns out prettier is not prettier at all... or at least many rules conflict with what we have configured for eslint, so I'll leave it out and configure eslint rules for code formatting instead like the keyword-spacing on from the last patch.

tahini:
To be rebased after Lea's branch is in

tahini:
This patch is ready for review.

@chertyb @smouss it touches pretty much all the project, so some of the files you are working on too. Maybe you can base your future work on this branch to avoid rebase hells later, unless there's an ideological objection to this.

babyadoresorange:
@tahini: eventually I think if you want to add Prettier you will need an [integration](https://prettier.io/docs/en/integrating-with-linters.html) to resolve those conflicts.

.vscode/settings.json
```diff
@@ -0,0 +1,5 @@
+{
+    "editor.codeActionsOnSave": {
```
marcdumais-work, line 2: Not sure that this works. If I add an easily fixed linter issue, nothing seems to happen after I save. Also potentially related, there is a warning related to this entry, when I open in Gitpod: "Property editor.codeActionsOnSave is not allowed"

marcdumais-work: Seems to work locally if I use VS Code. Maybe that's good enough? Though ideally it would work in Theia-based applications, like Gitpod.

tahini: It works in vscode and it's in the .vscode directory fwiw. Could it be a bug/missing feature in theia? Or is it expected that it should work?

marcdumais-work: Yes, it should work I think. Theia apps will use the .vscode folder, if present.

marcdumais-work: I noticed that the `eslint` VS Code extension is not part of the app. But even adding it is not enough. Maybe we can look at making linting work in Theia apps afterwards, if this was not originally wanted for this PR.

viewer-prototype/.eslintrc.js
```diff
@@ -0,0 +1,37 @@
+/** @type {import('eslint').Linter.Config} */
```
marcdumais-work, line 1: ATM I think only the `viewer-prototype` folder is covered, since there is no `.eslintrc.js` in the repo root. Probably not too important since there is little JS/TS code elsewhere in the repo?

tahini: Indeed, I only implemented it in the viewer-prototype for now. There is nothing much elsewhere, so we'll add it as we need it I think.

PR #90, opened 2020-06-25
Timegraph: Tree and chart synchronization

smouss:
The tree and the chart are synchronized when scrolling and wheeling.
The tree's rows are equalized with the chart's rows.

### DIFF
--- a/viewer-prototype/src/browser/trace-viewer/components/abstract-tree-output-component.tsx
+++ b/viewer-prototype/src/browser/trace-viewer/components/abstract-tree-output-component.tsx
@@ -9,7 +9,7 @@ export abstract class AbstractTreeOutputComponent<P extends AbstractOutputProps,
     renderMainArea(): React.ReactNode {
         const treeWidth = this.props.style.width - this.props.style.chartWidth - this.getHandleWidth();
         return <React.Fragment>
-            <div className='output-component-tree' style={{ width: treeWidth, height: this.props.style.height }}>
+            <div ref={this.treeRef} className='output-component-tree' onScroll={ev => { this.synchronizeTreeScroll() }} style={{ width: treeWidth, height: this.props.style.height }}>
                 {this.renderTree()}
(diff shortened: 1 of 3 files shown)


viewer-prototype/src/browser/trace-viewer/components/abstract-tree-output-component.tsx
```diff
@@ -9,19 +9,24 @@ export abstract class AbstractTreeOutputComponent<P extends AbstractOutputProps,
...
     }
 
+    treeRef = React.createRef<any>();
+    chartRef = React.createRef<any>();
+
     abstract renderTree(): React.ReactNode;
 
     abstract renderChart(): React.ReactNode;
 
+    abstract ScrollSync(): void;
```
tahini, line 28: Can you not make it abstract and just override it in the timegraph? Instead of requiring to implement an empty function in all sub-classes.

Also, to make clearer that this method is called when the tree is scrolled, not the chart, you could call it treeScrolled() (start the method name with lower case)

viewer-prototype/src/browser/trace-viewer/components/timegraph-output-component.tsx
```diff
@@ -66,6 +66,10 @@ export class TimegraphOutputComponent extends AbstractTreeOutputComponent<Timegr
         };
         this.chartLayer = new TimeGraphChart('timeGraphChart', providers, this.rowController);
         this.vscrollLayer = new TimeGraphVerticalScrollbar('timeGraphVerticalScrollbar', this.rowController);
+        
+        this.rowController.onVerticalOffsetChangedHandler((offset)=>{ 
+            this.treeRef.current.scrollTop=this.rowController.verticalOffset;
```
tahini, line 71: Is this.rowController.verticalOffset == offset? If so, use the offset in parameter. Otherwise, what is the offset parameter?

PR #92, opened 2020-07-02
Timegraph: implement tree view

chertyb:
Works fine for tree part but chart is not updating with visible rows

Signed-off-by: Lea Carlier <lea.carlier@polymtl.ca>


chertyb:
@delislesim is there a way to update timegraph chart rows?

delislesim:
> @delislesim is there a way to update timegraph chart rows?

I did not check the entire PR, but I don't think you should create a new chartLayer each time.

To answer your question I think it might need a few changes in the [timeline-chart library](https://github.com/theia-ide/timeline-chart), right now only the change in time range is causing an update to the chartLayer. Maybe the `TimeGraphRowController` should be responsible to keep track of what is visible and have the appropriate method to update the rows. Maybe look at: https://github.com/theia-ide/timeline-chart/blob/a005d9b9b36ba72671222fa629344a874bcc53ba/timeline-chart/src/layer/time-graph-chart.ts#L125 this is how the unitController is "updating" the chart layer.

Out of curiosity if zoom or time range is changed do you see if the row are updated as well?

bhufmann:
Thank you very for the updates. I will be great addition to the time graphs and I'm looking forward having it integrated. I have some initial comments though. 

I run the change and I noticed that the tree label wrap to new lines if the the text is too long. The wrapping will cause the the alignment of tree entries and time graph states to be off.  (see picture below)

When collapsing a tree node then the time graph on the right is not updated. 

Additionally, from code point of view, the tree component is called XYTree and this name is not fitting to the usage in the time graph. I suggest to refactor the tree implementation and have a independent name.

image

tahini:
It would seem that the state rows don't match the entries on the left. See screenshot below. The states part is exactly the same. In green is the original entries, while in red, it's the same state configuration, but associated with the wrong entries. 

The list to tree conversion probably changes the order of the ids that are sent to fetch the rows and this order is important when drawing the results?

BeforeAfterTgTree

viewer-prototype/src/browser/trace-viewer/components/utils/filtrer-tree/utils.tsx
```diff
@@ -10,15 +10,15 @@ const entryToTreeNode = (entry: Entry) => ({
         children: []
     } as TreeNode);
 
-export const listToTree = (list: Entry[]): TreeNode[] => {
+export const listToTree = (list: Entry[], rootId: number): TreeNode[] => {
```
tahini, line 13: You could add a default value for rootId. Also, why do you need the rootId parameter? I see it's just for the current workaround for empty rows... oh well!

chertyb:
@bhufmann did you pull latest commit ? Collapse behaviour should be disabled and no line wrapping
image

bhufmann:
> @bhufmann did you pull latest commit ? Collapse behaviour should be disabled and no line wrapping

Yes, you are right. I checked-out an older commit.

When using the correct commit, however, the tree and timegraph is not populating anymore. Left and right side are empty.  Hmm, I need to trouble-shoot a bit more to find out what is going on. 

image

viewer-prototype/src/browser/trace-viewer/components/timegraph-output-component.tsx
```diff
@@ -100,24 +107,39 @@ export class TimegraphOutputComponent extends AbstractTreeOutputComponent<Timegr
...
-                        {entry.labels[0] + '\n'}
-                    </p>;
-                }
-            })}
-        </React.Fragment>;
+        this.onCollapse = this.onCollapse.bind(this);
+        return  <EntryTree
+            collapsedNodes={this.state.collapsedNodes}
+            padding={12}
+            rootId={0}
```
bhufmann, line 137: I don't think you can assume that the root ID is 0. The root ID could be any number. In method listToTree the root nodes should be auto discovered.

viewer-prototype/src/browser/trace-viewer/components/data-providers/tsp-data-provider.ts
```diff
@@ -27,31 +26,20 @@ export class TspDataProvider {
...
+    async getData(ids: number[], entries: TimeGraphEntry[], viewRange?: TimelineChart.TimeGraphRange, resolution?: number): Promise<TimelineChart.TimeGraphModel> {
+        this.timeGraphEntries = [...entries];
         this.totalRange = this.timeGraphEntries[0].endTime - this.timeGraphEntries[0].startTime; // 1332170682540133097 - starttime
-        const selectedItems = new Array<number>();
-        this.timeGraphEntries.forEach(timeGraphEntry => {
-            selectedItems.push(timeGraphEntry.id);
-        });
-
-        let statesParameters = QueryHelper.selectionTimeQuery(QueryHelper.splitRangeIntoEqualParts(1332170682440133097, 1332170682540133097, 1120), selectedItems);
+        let statesParameters = QueryHelper.selectionTimeQuery(QueryHelper.splitRangeIntoEqualParts(1332170682440133097, 1332170682540133097, 1120), ids);
```
tahini, line 32: by the way, what are those numbers? where do they come from?

tahini:
By the way, I also had the problem of an empty view, but it got populated when I zoomed in. Is it related to this patch or was it there all along, I just opened more views now, I don't know. But can be investigated later, as far as I'm concerned

bhufmann:
> By the way, I also had the problem of an empty view, but it got populated when I zoomed in. Is it related to this patch or was it there all along, I just opened more views now, I don't know. But can be investigated later, as far as I'm concerned

The populating of the tree works now and the latest version fixed that. It's a nice UX improvement. However, the timegraph on the right side is not populated unless I open another view or zoom. It seems that a refresh is missing or fails. This didn't happen before this patch and it's seems to be related. I think we should fix it before merging this PR.

viewer-prototype/src/browser/trace-viewer/components/timegraph-output-component.tsx
```diff
@@ -43,13 +46,25 @@ export class TimegraphOutputComponent extends AbstractTreeOutputComponent<Timegr
...
+            if (this.treeRef.current) {
+                this.treeRef.current.scrollTop=this.rowController.verticalOffset;
+            }
+        });
+    }
+
+    createChartLayer(orderedTreeIds: number[]): void {
         const providers: TimeGraphChartProviders = {
-            dataProvider: async (range: TimelineChart.TimeGraphRange, resolution: number) => this.fetchTimegraphData(range, resolution),
+            dataProvider: async (viewRange: TimelineChart.TimeGraphRange, resolution: number) => this.fetchTimegraphData(orderedTreeIds, viewRange, resolution),
```
tahini, line 67: Talking to Soukaina here:

here should be

async (orderedTreeIds: number[], viewRange: TimelineChart.TimeGraphRange, resolution: number)

and the chart layer should be created only once. It's the data provider function call that needs to add a parameter I think.

So in the time-graph-chart.d.ts of the timeline-chart library, the TimeGraphChartProviders should be a function that also adds an array of entry ids.

viewer-prototype/src/browser/trace-viewer/components/timegraph-output-component.tsx
```diff
@@ -196,14 +231,14 @@ export class TimegraphOutputComponent extends AbstractTreeOutputComponent<Timegr
         }
     }
 
-    private async fetchTimegraphData(range: TimelineChart.TimeGraphRange, resolution: number) {
+    private async fetchTimegraphData(ids: number[], range: TimelineChart.TimeGraphRange, resolution: number) {
         const length = range.end - range.start;
```
tahini, line 235: Talking to Soukaina again:

In this method, you could get the ids using this.state.timegraphEntry and then with the 2 lines from componentDidUdate:

 const treeNodes = listToTree(treeResponse.model.entries);
 const orderedTreeIds = getAllVisibleNodeIds(treeNodes, []);

You have the orderedTreeIds. This way, you don't need the ids in parameter, the chart layer can be created in the constructor.

The only thing remaining is to trigger a refresh of the time graph chart on collapse (state change)

tahini:
This PR has been replaced with Soukaina's PR

PR #95, opened 2020-07-08
add filter to tree

chertyb:
Signed-off-by: Lea Carlier <lea.carlier@polymtl.ca>


viewer-prototype/src/browser/trace-viewer/components/xy-output-component.tsx
```diff
@@ -70,13 +70,16 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
         this.onSeriesChecked = this.onSeriesChecked.bind(this);
         this.onCollapse = this.onCollapse.bind(this);
-        return <XYTree
-            entries={this.state.XYTree}
-            collapsedNodes={this.state.collapsedNodes}
-            checkedSeries={this.state.checkedSeries}
-            onChecked={this.onSeriesChecked}
-            onCollapse={this.onCollapse}
-        />;
+        return this.state.XYTree.length
```
tahini, line 73: Is this change needed for this patch here?

chertyb: yes because I'm using the parent tree data to initialize filteredNodes in tree. If no data is sent, the result of filter shows No entries found at the beginning

viewer-prototype/src/browser/trace-viewer/components/utils/filtrer-tree/tree.tsx
```diff
@@ -74,14 +84,17 @@ export class FilterTree extends React.Component<FilterTreeProps> {
 
     handleCheck = (id: number): void => {
         let checkedIds: number[] = [];
-        const checkedNode = this.getNode(id);
+        const checkedNode = this.getNode(this.props.nodes, id);
         if (checkedNode) {
             if (checkedNode.children.length) {
                 const childrenIds = this.getAllChildrenIds(checkedNode, []);
+                const visibleChildrenIds = childrenIds.filter((id: number) => {
+                    return this.getNode(this.state.filteredNodes, id) !== undefined;
```
tahini, line 92: What about collapsed nodes? Now, if an entry is collapsed, but has children that would be visible with the filter, the children are checked when the parent is checked (for example, in the Disks I/O view, collapse a few devices (like loopX or sdX), filter by "read", and check the trace name at the top, it checks the read entry of the collapsed.

I think I would expect them not to be checked if they are not visible?

viewer-prototype/src/browser/trace-viewer/components/utils/filtrer-tree/tree.tsx
```diff
@@ -118,7 +131,49 @@ export class FilterTree extends React.Component<FilterTreeProps> {
...
+
+    filterTree = (nodes: TreeNode[], matchedIds: number[]): TreeNode[] => {
+        return nodes.filter((node: TreeNode) => matchedIds.indexOf(node.id) > -1)
+                    .map((node: TreeNode) => ({
+                        ...node,
+                        children: node.children ? this.filterTree(node.children, matchedIds) : []
+                    }));
+    }
+
+    renderFilterTree = (): JSX.Element | undefined => {
```
tahini, line 165: This is never undefined, so you can remove it from the return type, no?

viewer-prototype/src/browser/trace-viewer/components/xy-output-component.tsx
tahini: This method though now returns undefined, so you should add it to the return type

chertyb:
demo
Problem in question: when a parent node is collapsed (loop0 here) and the child is in the filter (read), if we check the parent, should the child also be checked or not?

MatthewKhouzam:
@PatrickTasse could you please check this out?

PatrickTasse:
From some examples I have seen, the behavior of a checkbox is the same whether the item is collapsed or not. I don't know if that's industry standard.

tahini:
I guess I was wrong... at least from what Trace Compass does.

filterTreeTraceCompass

viewer-prototype/src/browser/trace-viewer/components/utils/filtrer-tree/tree.tsx
```diff
@@ -118,7 +131,49 @@ export class FilterTree extends React.Component<FilterTreeProps> {
...
 
-    renderTreeNodes = (nodes: TreeNode[], _parent: TreeNode = defaultTreeNode, level = 0): JSX.Element | undefined => {
+    handleFilterChanged = (filter: string) : void => {
+        let filteredTree: TreeNode[] = [];
+        const matchedIds: number[] = [];
+        const rootNodes = this.getRootNodes();
+        rootNodes.forEach((node: TreeNode) => this.getMatchingIds(node, filter, matchedIds));
+        filteredTree = this.filterTree(this.props.nodes, matchedIds);
+        this.setState({filteredNodes: filteredTree});
+    }
```
tahini, line 141: There are many issues here that should be auto-fixed when saving the file is vscode. Maybe you need to add eslint plugin to your vscode?

tahini:
Behavior-wise, it's ok now

tahini:
Looks good, except when checking the root element after filter, grand-parent selection icon is checked while parent is half checked. From the Trace Compass screencast behavior, if all filtered children are checked, the parent should be checked and become partially checked when removing the filter.

treeSelection

We could accept this patch as is, then write the unit test that will detect this error and fix it, or fix it now.

bhufmann:
> I guess I was wrong... at least from what Trace Compass does.
> 
> filterTreeTraceCompass

I wonder if we'd like to do the auto-expansion of the tree when typing a filter like in Trace Compass? Could be done afterwards separately. Pls, let me know.

Otherwise it looks good to me.

viewer-prototype/src/browser/trace-viewer/components/utils/filtrer-tree/tree.tsx
```diff
@@ -72,6 +72,20 @@ export class FilterTree extends React.Component<FilterTreeProps, FilterTreeState
...
     };
 
+    getParentIdsToCheck = (parentId: number, ids: number[], toCheck: boolean): number[] => {
+        ids.push(parentId);
+        const parentNode = this.getNode(this.props.nodes, parentId);
+        if (parentNode) {
+            if (toCheck && this.areAllSiblingsChecked(parentNode)) {
+                this.getParentIdsToCheck(parentNode.parentId, ids, toCheck);
+            }
+            if (!toCheck && this.isNodeChecked(parentNode.parentId)) {
```
tahini, line 82: could be else if here?

viewer-prototype/src/browser/trace-viewer/components/utils/filtrer-tree/tree.tsx
```diff
@@ -86,19 +100,28 @@ export class FilterTree extends React.Component<FilterTreeProps, FilterTreeState
...
         const checkedNode = this.getNode(this.props.nodes, id);
         if (checkedNode) {
-            if (checkedNode.children.length) {
-                const childrenIds = this.getAllChildrenIds(checkedNode, []);
-                const visibleChildrenIds = childrenIds.filter((childId: number) =>this.getNode(this.state.filteredNodes, childId) !== undefined);
-                if (this.isNodeChecked(id)) {
-                    checkedIds = checkedIds.concat(visibleChildrenIds);
-                } else {
+            const childrenIds = this.getAllChildrenIds(checkedNode, []);
+            const visibleChildrenIds = childrenIds.filter((childId: number) =>this.getNode(this.state.filteredNodes, childId) !== undefined);
```
tahini, line 104: This line could be extracted to getVisibleChildren(). Line 150 has a similar call

viewer-prototype/src/browser/trace-viewer/components/utils/filtrer-tree/tree.tsx
```diff
@@ -86,19 +100,28 @@ export class FilterTree extends React.Component<FilterTreeProps, FilterTreeState
...
+            if (!this.isNodeChecked(id)) {
```
tahini, line 105: I observe the correct behavior in the application, but reading the code, I fail to understand! In both the if and the else here, there's an if (checkedNode.children.lengt) and a else that does checkedIds = checkedIds.concat(id); Why does it work??

chertyb: behaviour is different if you want to check or uncheck a node so I verify that first. Next is if the node has children or not. I thought it was easier to define it this way

tahini: I understand, but in both cases, if the node has no children, it gets pushed in the checkedIds? That's the part I don't understand...

chertyb: we need to check or uncheck the leaf, then verify the related nodes status

viewer-prototype/src/browser/trace-viewer/components/utils/filtrer-tree/tree.tsx
```diff
@@ -74,20 +98,30 @@ export class FilterTree extends React.Component<FilterTreeProps> {
...
+                    checkedIds = checkedIds.concat(visibleChildrenIds);
+                } else {
+                    checkedIds = checkedIds.concat(id);
+                }
+                if (this.isNodeChecked(checkedNode.parentId)) {
+                    const parentsToCheck = this.getParentIdsToCheck(checkedNode.parentId, [], false);
+                    checkedIds = checkedIds.concat(parentsToCheck);
                 }
             }
             this.props.onChecked(checkedIds);
```
tahini, line 127: As discussed offline, rename the onChecked method to onToggleCheck to make everything much clearer! ;-)

viewer-prototype/src/browser/trace-viewer/components/utils/filtrer-tree/tree-node.tsx
```diff
@@ -55,7 +55,7 @@ export class TreeNodeComponent extends React.Component<TreeNodeComponentProps> {
                     { this.isLeaf()
                         ? <span style={{paddingLeft:this.props.padding}}></span>
                         : <span
-                            onClick={() => this.props.onCollapsed(this.props.node.id)}
+                            onClick={() => this.props.onToggleCollapse(this.props.node.id)}
```
bhufmann, line 58: `According` to the coding guidelines under the react section shouldn't this line something like the following:

    render(): JSX.Element {
        return <div onClick={this.onClickDiv} >
            <span style={{padding: 5}}>
                {this.renderCheckbox(this.props.checkedStatus)}
            </span>
            {this.props.name}
        </div>;
    }

    protected onClickDiv = () => {
        this.props.onToggleCheck(this.props.id)
    }


[1] https://github.com/eclipse-theia/theia/wiki/Coding-Guidelines

viewer-prototype/src/browser/trace-viewer/components/utils/filtrer-tree/checkbox-component.tsx
```diff
@@ -27,7 +27,7 @@ export class CheckboxComponent extends React.Component<CheckboxProps> {
     };
 
     render(): JSX.Element {
-        return <div onClick={() => this.props.onChecked(this.props.id)} >
+        return <div onClick={() => this.props.onToggleCheck(this.props.id)} >
```
bhufmann, line 30: Sorry, I missed that in my previous review. I have same comment than last review:
```
render(): JSX.Element {
        return <div onClick={this.onClickDiv} >
            <span style={{padding: 5}}>
                {this.renderCheckbox(this.props.checkedStatus)}
            </span>
            {this.props.name}
        </div>;
    }

    private onClickDiv = () => {
        this.props.onToggleCheck(this.props.id)
    }
```

Issue #102, opened 2020-07-20
Use latest Theia (@theia/latest) instead of @theia/next


bhufmann:
Right now the theia-trace-extension is using the @theia/next. Instead it should be base on the latest release to avoid unexpected behaviours.

tahini:
Fixed by commit 4df221c169d1e84cddf05b7e4d886852e9a9dd2a

PR #104, opened 2020-07-27
#102: Use use latest @theia/latest release instead of @theia/next

bhufmann:


electron-app/package.json
```diff
@@ -23,26 +23,26 @@
...
+    "@theia/terminal": "latest",
+    "@theia/editor": "latest",
+    "@theia/languages": "latest",
+    "@theia/markers": "latest",
+    "@theia/monaco": "latest",
+    "@theia/messages": "latest",
+    "@theia/vsx-registry": "latest",
+    "@theia/keymaps": "latest",
+    "@theia/getting-started": "latest",
     "@theia/electron": "next",
```
marcdumais-work, line 41: I think this one needs to be `latest` as well?

marcdumais-work: I had a quick look and this might work as-is, only because `@theia/electron` is not itself pulling other Theia extensions. But still preferable to have all Theia extensions the same: `latest` or `next`.

marcdumais-work: I think I see why you might have kept this one on `next`. We've updated Electron recently on master, from 4.x to 9.x. Since you've been using `next`, that change has already landed in this repo and in `yarn.lock`. By adopting `latest` right now, you're rolling-back Electron to 4.x, against your current `yarn.lock`. 

I would suggest to keep things as they are now, and wait until Thursday for the Theia release. At that point, `@theia/electron:latest` will also be using `Electron 9.x`, so you can switch and I think everything should fall back on its feet.

bhufmann: yes, I  missed that

bhufmann: I'll update this PR on Thursday when @theia/electron/latest will use Electron 9.x

marcdumais-work: I'll give it another try after the update

bhufmann:
@marcdumais-work could you please have another look at this patch?

marcdumais-work:
@bhufmann there is an issue preventing testing on Gitpod: https://github.com/theia-ide/theia-trace-extension/issues/106

So I proceeded locally on my laptop instead.

marcdumais-work:
> @bhufmann there is an issue preventing testing on Gitpod: #106
> 
> So I proceeded locally on my laptop instead.

PR: https://github.com/theia-ide/theia-trace-extension/pull/107

marcdumais-work:
@bhufmann not a huge deal but ATM you have 2 commits - you could squash them into one.

bhufmann:
> @bhufmann not a huge deal but ATM you have 2 commits - you could squash them into one.

Oh, not sure how it happened. I'll fix it.

marcdumais-work:
Awesome. Ready to merge AFAIK. I'll let you do so in case you're waiting for Matthew to review.

marcdumais-work:
Oops - @bhufmann I just noticed your commit had no message :(

bhufmann:
> just to prevent merge until you have a chance to correct commit message

I'll fix it.

bhufmann:
Thanks for the reviews!

Issue #60, opened 2020-05-25
Package theia-trace-extension for linux distros and other


tahini:
It should be easy for users to install and run the theia-trace-extension and trace server.

We should look at how other nodejs applications do, like vscode. For ubuntu, snap can be used

marcdumais-work:
With https://github.com/theia-ide/theia-trace-extension/pull/74 we have a base: an Electron builder config was added, that can be used to produce end-user packages for the example application. ATM it supports Linux packaging (AppImage and Debian packages). 

Windows and Mac packaging are also possible, but IIUC require signing so the end-users can install them without security warnings. 

See the following theia-apps PR, where Travis CI is used to produce and make available packages for an Electron example application. The plan for the main Eclipse Theia project is to use the Eclipse Foundation signing service. 

https://github.com/theia-ide/theia-apps/pull/370
Corresponding wiki entry: https://github.com/theia-ide/theia-apps/wiki/Publishing-a-release
Released packages: https://github.com/theia-ide/theia-apps/releases (note: I am not 100% happy with a name such as "Theia-1.2.0.AppImage" - I'll correct at some point to add "example" in there)

marcdumais-work:
Note: I think this applies to the example app - the issue title seems to refer to the extension, not the app. The extension should also be packaged and published, but in the form of a `npm package`.

MatthewKhouzam:
@arfio Is this what you progressed on by making the appimage?

PR #105, opened 2020-07-28
#60: Add github action to publish next version

bhufmann:

### DIFF
--- a/.github/workflows/publish-next.yml
+++ b/.github/workflows/publish-next.yml
@@ -1,0 +1,22 @@
+name: NPM package (next)
+on:
+  push:
+    branches: [ master ]
+#  release:
+#    types: [created]
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    # Setup .npmrc file to publish to npm
+    - uses: actions/setup-node@v1
+      with:
+        node-version: '12.x'
+        registry-url: 'https://registry.npmjs.org'
+        # Defaults to the user or organization that owns the workflow file
+        scope: '@theia-ide' 
+    - run: yarn
+    - run: yarn publish:next
+      env:
+        NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
--- a/.npmignore
+++ b/.npmignore
@@ -1,0 +1,17 @@
+.DS_Store
+node_modules
+build
+*.log
+.idea
+.metadata
+jdt.ls-java-project
+lerna-debug.log
+.nyc_output
+coverage
+.browser_modules
+download
+*ui-spec.ts
+*slow-spec.ts
+test-resources
+browser-app
+electron-app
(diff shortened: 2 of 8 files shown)


tahini:
What does this publish exactly, the viewer prototype? Are we ready for this? Is it mature enough yet? Or we go for release-early/release-often approach?

bhufmann:
> What does this publish exactly, the viewer prototype? Are we ready for this? Is it mature enough yet? Or we go for release-early/release-often approach?

The publish will only publish the viewer prototype and not the electron nor the browser app. Electron/browser app are only examples to easily test during development. Applications (IDEs) should be deployed differently (for example under theia-apps).

The Github action added in this patch would publish a "next" version at every merged pull request in master. It's similar to nightly builds that are providing bleeding edge. I also added a command in the root package.json file for providing releases (latest), but it's not yet connected to a Github action. 

It's really for simplification. The reason having it as npm package is to be able to integrate it more easily into applications, like a Theia based IDE. Users can include in their IDE build to try. We were asked to provide an npm package for the trace extension. We already publish tsp-typescript-client (https://www.npmjs.com/package/tsp-typescript-client) and the timeline-chart (https://www.npmjs.com/package/timeline-chart) the same way as npm packages.

> Are we ready for this? Is it mature enough yet?
These are a good questions. Probably not. For sure not for releases. But it could be ok providing "next" versions. However, the name that it would be published as would be "trace viewer" and it's probably not the name that we'd like. Once published the package will be there basically forever.

We can hold of a bit with this PR, but sooner than later we need to think about the deployment so that it can be included in user's IDEs.

tahini:
Do you have a need to publish this now for internal purposes? My main concern is that we're investigating if [part of] the trace viewer could be a vscode plugin, in which case, our publishing platform would change. I'd wait a bit for the investigation to go a little further before publishing anything since, as you put it, the package will be there basically forever...

I think we can have a sensible answer by september about vscode or not vscode. And if so, we can easily port everything that is there to a vscode plugin and publish there instead.

bhufmann:
Yes, publishing as npm package would make it easier for our stakeholders, but could be useful for others as well that want to integrate with it. 

I agree that there are some unknowns regarding the platform and maybe it's better to hold off with this PR for now, until we know if we switch to a vscode plugin or not. I'll check for some alternatives for our stakeholders in meanwhile.

tahini:
I'm just raising concerns here, but I don't want to stop progress or be an impediment to collaboration. If it's easier for you guys, we can go ahead with this patch. Just let me know if it's going to be used right now by your stakeholders, I'll positively review this patch. But if the timeline is a month or so, it could wait for more certainty.

bhufmann:
> I'm just raising concerns here, but I don't want to stop progress or be an impediment to collaboration. If it's easier for you guys, we can go ahead with this patch. Just let me know if it's going to be used right now by your stakeholders, I'll positively review this patch. But if the timeline is a month or so, it could wait for more certainty.

These are valid concerns. I'll clarify on my side to see if we could wait with this until we have some more certainty.

PR #108, opened 2020-08-06
Table tree

chertyb:
Show entries data in a table with tree structure


viewer-prototype/src/browser/trace-viewer/components/utils/filtrer-tree/table-cell.tsx
```diff
@@ -0,0 +1,25 @@
...
+    constructor(props: TableCellProps) {
+        super(props);
+    }
+
+    render(): React.ReactNode {
+        const content: React.ReactNode = (this.props.nodeKey !== 'Legend')
+            ? this.props.node[this.props.nodeKey as keyof TreeNode]
+            : undefined;
+        return (
+            <td key={this.props.nodeKey+'-'+this.props.node.id}>
```
tahini, line 19: Is the key parameter required? I don't see it in the cell

tahini:
Also, a bug that may have been present previously too, but went unnoticed: Try opening the Memory Usage view for the pacman trace. In all appearance, the view is empty and everything is ok, but in the console, I get an infinite number of these:

Uncaught (in promise) TypeError: treeModel is null
    updateTree xy-output-component.tsx:211
    componentDidUpdate xy-output-component.tsx:57
    React 6
    unstable_runWithPriority scheduler.development.js:653
    React 5

delislesim:
I don't want to slow down the work here, but have you considered using [react-virtualized](https://github.com/bvaughn/react-virtualized) or [react-window](https://github.com/bvaughn/react-window) to make a virtualized table in order to maybe reduce the code you have to right and increase the performance for rendering large tree?

tahini:
@delislesim from what I understand of react-window, this patch is not incompatible with it, we can later wrap the table tree (without the top filter part) into a react-window's grid component.

tahini:
@chertyb do you plan to continue working on this patch, or I can take over your branch, fix the bugs and start a new PR.

chertyb:
I will have access to my computer tomorrow to look at this bug. It seemed to work on my computer but let me check again. I should be able to fix it.

viewer-prototype/src/browser/trace-viewer/components/utils/filtrer-tree/tree.tsx
```diff
@@ -59,7 +66,17 @@ export class FilterTree extends React.Component<FilterTreeProps, FilterTreeState
     };
 
     handleCollapse = (id: number): void => {
-        this.props.onToggleCollapse(id);
+        const nodes = sortNodes(this.state.filteredNodes, this.state.sortConfig);
+        this.props.onToggleCollapse(id, nodes);
```
tahini, line 70: Why do you need the sorted nodes in the toggleCollapse, what is this parameter exactly? I'm not sure how to "javadoc" typescript stuff so users know the meaning of the parameter, but just document it somewhere and we'll make it official at some point.

chertyb: because to get visible ids in xy-chart-component to update orderedIds, I need the tree nodes and a list of collapsed nodes. Toggle collapse will change this list, so I send the nodes to parent component since it doesnt have it in its state. There may be a better way to do this, I'm open to suggestions.

tahini:
Thanks for this PR, I rebase on master and fixed the timegraph to use the table as well. So, I'll close this PR and open another one from my repo

PR #113, opened 2020-08-28
Events table infinite scrolling.

mad-season:
Uses ag-grid's 'infinite' row model type.
Added a switch to display event index (very useful for testing).
Tested on Chrome Version 84.0.4147.135 (Official Build) (64-bit) using sample traces ls-ust and yum.

Signed-off by: Akshay Sharma <akshay.c.sharma@ericsson.com>


mad-season:
Screwed up history. New PR time.

PR #116, opened 2020-09-04
filter-tree: Use a table layout in filter tree

tahini:
This allows to display many columns in the tree. Columns are sortable.

Signed-off-by: Lea <lea.carlier@polymtl.ca>


viewer-prototype/src/browser/trace-viewer/components/utils/filtrer-tree/table-row.tsx
```diff
@@ -0,0 +1,91 @@
...
+    };
+
+    render(): React.ReactNode | undefined {
+        if (!this.props.node) {return undefined;}
+        const children = this.renderChildren();
+
+        return (
+            <React.Fragment>
+                <tr>{this.renderRow()}</tr>
+                {children}
```
chertyb, line 87: we can call this.renderChildren() directly

viewer-prototype/src/browser/trace-viewer/components/utils/filtrer-tree/table.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+    initializeSortConfig = (): void => {
+        const config: SortConfig[] = [];
+        this.sortableColumns.forEach((column: string) => {
+            config.push({column: column, sortState: sortState.default});
+        });
+        this.props.onSortConfigChange(config);
+    };
+
+    getTableHeaderKeys = (): string[] => {
```
chertyb, line 41: add TODO: remove hardcoded values when trace-server can send column headers

chertyb (changes requested):
table works fine, only add TODO and I'll approve the request

chertyb (approved):

bhufmann:
Just to let you know that I've started looking at the patch and review is ongoing.

bhufmann (changes requested):
Thanks for adding the table. It's a nice addition to the Time Graph and the XY charts. It will be useful when implementing a statistics view.

The columns are quite undefined. They depend on the extra attributes TimeGraphEntryModel implementation of each data provider. We'll need to use the columns headers and labels for each rows, that are currently defined in the TSP, but no time graph nor xy data provider has implemented it yet. Also, the trace server and tsp-typescript-client should ignore attributes that are not in the TSP. This is outside the scope of this PR.

tahini:
I added the todo, for the table headers and label, since changes are required in the tsp-client and trace server, it can be for later patches?

bhufmann:
> I added the todo, for the table headers and label, since changes are required in the tsp-client and trace server, it can be for later patches?

I think so. Maybe have the trace server patch [1] that only serializes fields that are in the TSP and related tsp-client patch should be merged first. Then, the time graphs won't unnecessary columns.

[1] https://git.eclipse.org/r/c/tracecompass.incubator/org.eclipse.tracecompass.incubator/+/167256

bhufmann:
Another question, the Legend column doesn't really make sense for time graphs, right?

bhufmann:
> Another question, the Legend column doesn't really make sense for time graphs, right?

After testing, I realized that the Legend column is only for XY charts and not for time graph.

bhufmann (approved):
It looks good to me

tahini:
Simply rebased, I'll merge for now so the update with the new tsp patches will be more obvious.

tahini:
Thanks @chertyb for this great improvement!

Issue #117, opened 2020-09-10
Exception in tsp-data-provider when opening 'Resources Status' view


mad-season:
RHEL7 Linux VM, Chrome Version 75.0.3770.100 (Official Build) (64-bit)

![exception](https://user-images.githubusercontent.com/68356105/92680727-3e086880-f2f1-11ea-8fab-d8eb37a463c5.gif)

Occurs in multiple traces in TraceCompassTutorialTraces.

Uncaught (in promise) TypeError: Cannot read property 'endTime' of undefined
    at TspDataProvider.getData (tsp-data-provider.ts:31)
    at TimegraphOutputComponent.fetchTimegraphData (timegraph-output-component.tsx:227)
    at Object.dataProvider (timegraph-output-component.tsx:57)
    at TimeGraphChart.<anonymous> (time-graph-chart.ts:156)
    at step (time-graph-chart-selection-range.ts:4)
    at Object.next (time-graph-chart-selection-range.ts:4)
    at time-graph-chart-selection-range.ts:4
    at new Promise (<anonymous>)
    at push.../node_modules/timeline-chart/lib/layer/time-graph-chart.js.__awaiter (time-graph-chart-selection-range.ts:4)
    at TimeGraphChart.push.../node_modules/timeline-chart/lib/layer/time-graph-chart.js.TimeGraphChart.maybeFetchNewData (time-graph-chart.ts:146)

PR #119, opened 2020-09-11
Timegraph: Fixed issue #117 (undefined variable exception)

smouss:
Fixed undefined variable (timeGraphEntries) exception in tsp-data-provider.ts

Signed-off-by: soukaina <soukaina.moussaoui@polymtl.ca>


tahini (changes requested):
I can't reproduce the problem, but seeing this patch, I'm pretty convinced this solves it! Thanks!

Just a little comment to re-order and refactor the if and else statements and it should be ok. Also, please cherry-pick your patch on master instead of merging the branches.

bhufmann (approved):
I was able to test it with a local changed trace server.

tahini (approved):
I'll approve, we can refactor later

tahini:
I'll rebase the PR and push again. Thanks @smouss

PR #124, opened 2020-09-21
Vscode migration

tahini:


tahini:
Current status: working great!

TODO before acceptance:
- Trace explorer is a cheat, it's just the root level folders of the workspace. Need a proper trace open or just a context menu in the file explorer to directly open trace for now.
- Right now, we need to manually build the extension and react-app before pressing F5. We should make it automatic in vscode, at least to call the build when F5 is pressed.
- The theia apps should include this new vscode plugin and build correctly with it.

tahini:
@bhufmann I think this is ready to be tried and reviewed and make sure it's not a case of "works-on-my-machine".

For now, it only runs as a vscode-plugin (f5 from vscode after running yarn build everywhere), I tried to add the browser and electron apps, but even though the package.json file and the directory are identical to what is there now, except that the dependency to the theia-trace-extension is removed, it doesn't build! But maybe if we release the vscode extension soon, we won't be needing those.

tahini:
Since this is a big patch whose rebases may not always be trivial, please let me know when you plan to review this, so I can rebase it on the latest master. Thanks!

marcdumais-work:
It's a bit hard to follow right now where things changed (not being very aware of recent changes generally). The build errors I see are probably not directly related to the vscode extension -  they should be able to build independently (contrary to Theia extensions).

About the vscode extension:

I suggest you build and package the extension and then just add a `file://` URI to the app's package.json, [along with other vscode extensions](https://github.com/theia-ide/theia-trace-extension/blob/master/electron-app/package.json#L64-L132) we consume. 

To package, add `'vsce': 'latest'` as a dev-dependency and use it in the vscode extension's directory: `vsce package`. Make the changes it complains about, like add a `publisher` - this will be needed to publish the extension later. Once it passes, you obtain a `.vsix` extension.

package.json
```diff
@@ -1,25 +1,29 @@
...
+    "test": "lerna run test"
   },
   "devDependencies": {
-    "lerna": "2.4.0"
+    "lerna": "2.4.0",
+    "rimraf": "^3.0.2"
   },
   "workspaces": [
-    "viewer-prototype",
-    "browser-app",
```
marcdumais-work, line 22: Is the intention to drop the example applications?

marcdumais-work: Add entries for the new app folders if you want them built along with the rest:
```json
"app/browser",
"app/electron"
```

package.json
```diff
@@ -1,25 +1,29 @@
...
   "private": true,
   "scripts": {
+    "reinstall": "rimraf ./**/node_modules && yarn install",
     "prepare": "lerna run prepare",
-    "rebuild:browser": "theia rebuild:browser",
-    "rebuild:electron": "theia rebuild:electron",
-    "start:browser": "yarn rebuild:browser ; yarn --cwd browser-app start",
-    "start:electron": "yarn rebuild:electron ; yarn --cwd electron-app start",
+    "build": "lerna run build",
+    "build:browser": "theia rebuild:browser",
```
marcdumais-work, line 7: I would advise to keep the original names: `rebuild:*` - it's kind of a convention and does not build the app, just rebuilds the native npm dependencies for either `browser` or `electron`.

packages/vscode-extension/package.json
```diff
@@ -0,0 +1,263 @@
...
+    "@types/mocha": "^7.0.2",
+    "@types/node": "^13.11.0",
+    "@types/react": "^16.9.43",
+    "@types/react-dom": "^16.9.8",
+    "@types/vscode": "^1.47.0",
+    "@typescript-eslint/eslint-plugin": "^2.30.0",
+    "@typescript-eslint/parser": "^2.30.0",
+    "babel-eslint": "10.1.0",
+    "babel-jest": "^24.9.0",
+    "babel-loader": "8.1.0",
```
marcdumais-work, line 135: I see this dependency not being found when building the `browser example app`. Maybe it should be a runtime dependency. but most likely it's not a good idea to mix Theia and vscode stuff under the same monorepo, with shared hoisted dependencies. 

Since the Theia example app and the vscode trave viewer extension have no build-time relation, they should be be built separately. They can still be in the same repo though, just not in the same yarn workspace.

packages/vscode-extension/package.json
```diff
@@ -0,0 +1,263 @@
+{
+  "name": "@trace-viewer/vscode-extension",
```
danieltomasku, line 2: When trying to package the extension with `vsce package` it complains that this is an invalid extension name. I think it doesn't liked the scoped naming.

packages/vscode-extension/react-app/src/logo.svg
```diff
@@ -0,0 +1,7 @@
...
+        <circle cx="420.9" cy="296.5" r="45.7"/>
+        <path d="M520.5 78.1z"/>
+    </g>
+</svg>
```
danieltomasku, line 7: You can remove this file (unless you want to use the react logo 🙂)

packages/vscode-extension/react-app/src/App.css
```diff
@@ -0,0 +1,28 @@
...
+}
+
+.App-intro {
+  font-size: large;
+}
+
+@keyframes App-logo-spin {
+  from { transform: rotate(0deg); }
+  to { transform: rotate(360deg); }
+}
```
danieltomasku, line 28: Seems like you may be still in the process of cleaning things up, but I know `create-react-app` tends to generate a lot of cruft. Might be worth going through the `react-app` package and cleaning up anything unused.

danieltomasku (commented):
This is a nice start @tahini and @babyadoresorange ! I did a quick high-level overview and have some minor comments below, but I will do a deeper code dive and test the extension further this week. I think it will be easier to review when some of the generated `create-react-app` things are cleaned up.

bhufmann:
Great to see so much action on the PR. I'm going to try it out as well.

marcdumais-work:
@tahini is it desired to keep the old Theia extension for now? (other than in the repo's history)

Ideally the repo's structure would be simplified if the main content is a vscode extension. The example applications can still have entries in the main `package.json` to build/start them, but they are just a convenience to the extension's developers, e.g. for PR reviews, providing (with Gitpod's help) a pre-built app that has the extension installed and ready to test. 

But it would also work to just locally clone the repo, open it in vscode and press `F5`. 

Repos that host a vscode extensions tend to be simple and follow a similar pattern. Generally the extension is built at repo root, with the root `package.json` being the extension's. Typically, the build steps are something like:
```
$> git clone <entension repo>; cd <repo>
# look for yarn.lock or package-lock.json to know which of yarn or npm to use to install dependencies
$> npm install 
or
$> yarn install
$> vsce package
```
If this repo follows similar, we can submit a PR [here](https://github.com/open-vsx/publish-extensions/pulls) to [add an entry](https://github.com/open-vsx/publish-extensions/blob/master/extensions.json) for our extension, and it will be automatically published on `open-vsx.org`, I think both nightly and solid revisions, whenever we add a release tag.

marcdumais-work:
Q: I see that the vscode extension seems to be split in a few components/packages:
 - @trace-viewer/base
 - @trace-viewer/react-components
 - @trace-viewer/vscode-extension


Are these the seeds of a group of vscode extensions that would, together, provide trace viewing capabilities? Or more the internal breakdown of components of, ultimately, a single trace viewing extension?

bhufmann:
I was able to run it and open a trace in VsCode. Nice!

I noticed that the events table is not working properly. I tried to debug in the react-components, but the breakpoints don't hit. I see a warning message "Unbound breakpoint". However, breakpoints work in the vscode-extension package. 

I wonder if something is missing in the launch configuration, but I haven't figured out what. Anybody any idea?

tahini:
> I was able to run it and open a trace in VsCode. Nice!
> 
> I noticed that the events table is not working properly. I tried to debug in the react-components, but the breakpoints don't hit. I see a warning message "Unbound breakpoint". However, breakpoints work in the vscode-extension package.
> 
> I wonder if something is missing in the launch configuration, but I haven't figured out what. Anybody any idea?

Thanks for the feedback. Unfortunately, the issue of debugging the webview (react components) bugs me. I don't know how to do it! It runs in an iframe and the webview developer tools of vscode does not have a debugger. Besides, our current way of building the react app builds a production app, so it wouldn't be meaningful debugging, but that can be fixed. 

I asked on the theia-help channel on discord and @marechal-p pointed me to https://stackoverflow.com/questions/3275816/debugging-iframes-with-chrome-developer-tools but it applies to the browser (chrome or firefox) and we are not running in the context of a real browser. Anyway, if you figure it out, let me know, but that's one thing I still have to look into.

danieltomasku:
> Thanks for the feedback. Unfortunately, the issue of debugging the webview (react components) bugs me. I don't know how to do it! It runs in an iframe and the webview developer tools of vscode does not have a debugger. Besides, our current way of building the react app builds a production app, so it wouldn't be meaningful debugging, but that can be fixed.
> 
> I asked on the theia-help channel on discord and @marechal-p pointed me to https://stackoverflow.com/questions/3275816/debugging-iframes-with-chrome-developer-tools but it applies to the browser (chrome or firefox) and we are not running in the context of a real browser. Anyway, if you figure it out, let me know, but that's one thing I still have to look into.

Hi, I had looked into debugging webviews for another extension and found this `Open Webview Developer Tools` command you can run in the command palette in VS Code:
image
It's not very well documented, but it opens the browser developer tools specifically for the webview. This should be what you're looking for. As for the React production app--you should be able to keep the source maps and then the dev tools will automatically translate between the compiled source and the original source so you can set breakpoints more intuitively. Hope this helps!

bhufmann:
> Hi, I had looked into debugging webviews for another extension and found this `Open Webview Developer Tools` command you can run in the command palette in VS Code:
> image
> It's not very well documented, but it opens the browser developer tools specifically for the webview. This should be what you're looking for. As for the React production app--you should be able to keep the source maps and then the dev tools will automatically translate between the compiled source and the original source so you can set breakpoints more intuitively. Hope this helps!

Thanks @danieltomasku! I found that too and I tried it. You have to be in the webview and execute that command and developer tools will open in a browser. Otherwise it won't open. Unfortunately we cannot debug in vscode itself, but browser is also ok. 

I was not able to setup the source maps to find the typescript source code.  Any ideas how to setup the source map?

tahini:
> Hi, I had looked into debugging webviews for another extension and found this `Open Webview Developer Tools` command you can run in the command palette in VS Code:
> image
> It's not very well documented, but it opens the browser developer tools specifically for the webview. This should be what you're looking for. As for the React production app--you should be able to keep the source maps and then the dev tools will automatically translate between the compiled source and the original source so you can set breakpoints more intuitively. Hope this helps!

And you guys have a debugger in those developer tools? I don't! I can see the console, network, profiler, etc. but no debugger. Maybe I didn't install something?

As for the typescript source map, maybe you need to build the react app in development mode. I did it manually by updating the build script, replacing production by development.

tahini:
> Q: I see that the vscode extension seems to be split in a few components/packages:
> 
>     * @trace-viewer/base
> 
>     * @trace-viewer/react-components
> 
>     * @trace-viewer/vscode-extension
> 
> 
> Are these the seeds of a group of vscode extensions that would, together, provide trace viewing capabilities? Or more the internal breakdown of components of, ultimately, a single trace viewing extension?

The 'base' and 'react-components' packages are meant to be reusable components that are consumed by the vscode extension but can also be consumed by other apps, like eventual grafana or jenkins plugins. This repo is thus a monorepo that contains among other things a vscode extension. I'm not sure how this can work out with the vscode extension release. We could create another repo for the vscode extension if needed.

bhufmann:
> > Hi, I had looked into debugging webviews for another extension and found this `Open Webview Developer Tools` command you can run in the command palette in VS Code:
> > image
> > It's not very well documented, but it opens the browser developer tools specifically for the webview. This should be what you're looking for. As for the React production app--you should be able to keep the source maps and then the dev tools will automatically translate between the compiled source and the original source so you can set breakpoints more intuitively. Hope this helps!
> 
> And you guys have a debugger in those developer tools? I don't! I can see the console, network, profiler, etc. but no debugger. Maybe I didn't install something?
> 
> As for the typescript source map, maybe you need to build the react app in development mode. I did it manually by updating the build script, replacing production by development.

bhufmann:
> And you guys have a debugger in those developer tools? I don't! I can see the console, network, profiler, etc. but no debugger. Maybe I didn't install something?
> 
> As for the typescript source map, maybe you need to build the react app in development mode. I did it manually by updating the build script, replacing production by development.

I was able to get source maps working (kind of). We have to do the following steps:

Add the following line to the webpack-config.js:
devtool: "eval-source-map"

Add the following line to the launch.json:
"sourceMaps": true

Above I said it works "kind of", because I'm only able set breakpoints to some lines, e.g. in tsp-data-provider.ts, other lines are disabled. Any ideas?

marcdumais-work:
> The 'base' and 'react-components' packages are meant to be reusable components that are consumed by the vscode extension but can also be consumed by other apps, like eventual grafana or jenkins plugins. 

Thanks for the explanation. So it makes sense for these two components to each be published on npm. But probably not for `@trace-viewer/vscode-extension`, which I expect an extender would modify/extend by forking the repo, and which built form (`.vsix`) will be ideally published on open-vsx.org and if desired the Microsoft Marketplace. 

> We could create another repo for the vscode extension if needed.

I think that's the best way forward, instead of inventing a twisted way top make all co-habitate. The extension can have its own simple repo and consume the two other components from npm. If we add it to the open-vsx build/publish service's config, a new `next` version will be built and published max 24 h after a fix is merged (assuming CIs publish a `next` automatically) in either the shared component or extension.

marcdumais-work:
> We could create another repo for the vscode extension if needed.

I can do that, BTW, in the `theia-ide` org. Let me know.

tahini:
Ok, so I found the debugger under the "Sources" tab. I'm used to firefox where there is a Debugger tab. I can now officially debug!

> Above I said it works "kind of", because I'm only able set breakpoints to some lines, e.g. in tsp-data-provider.ts, other lines in the same file are disabled. Any ideas?

Not sure why, but the file that we see in the debugger is some kind of pre-compiled version of the original typescript. I don't know where it got this. The lines where you put the breakpoints actually correspond to the original source code! I'll investigate how to have the proper files

tahini:
I updated and did some cleanup.

I tried to build the vscode extension, but it would seem that the extension really should be at the root of the repo, as @marcdumais-work  suggested, not as a workspace (vsce package does not find the modules that are in the root node_modules, just those in its own node_modules).

What still needs to be done before merging and going forward

1- The vscode extension will move to its own new repo, as well as the 2 npm packages base and react-components, to their own repos as well.

2- Before doing so, we need to figure out under which github team/account (can't be theia-ide, or eclipse-theia, maybe tracecompass?), under which umbrella (ideally Eclipse Foundation to keep the Trace Compass naming, so we can publish as "tracecompass-community") and under what name (traceviewer??). It is being investigated.

marcdumais-work:
> 2- Before doing so, we need to figure out under which github team/account (can't be theia-ide, or eclipse-theia, maybe tracecompass?), under which umbrella (ideally Eclipse Foundation to keep the Trace Compass naming, so we can publish as "tracecompass-community") and under what name (traceviewer??). It is being investigated.

I advise not to stress too much about that part, or let you be blocked by these choices. If needed, go with clearly temporary names, that will need to be changed (I'm sure @MatthewKhouzam would have suggestions). It's not a big deal, and you can aim to find proper permanent names for v1.0, once all is clear and determined.

tahini:
> I advise not to stress too much about that part, or let you be blocked by these choices. If needed, go with clearly temporary names, that will need to be changed (I'm sure @MatthewKhouzam would have suggestions). It's not a big deal, and you can aim to find proper permanent names for v1.0, once all is clear and determined.

I agree we/I don't want this PR to be blocked here for too long, we have to begin somewhere and start using it. So we could start a new repo under theia-ide (ironic) for the vscode-trace-extension that we won't publish yet on openvsx. 

But how about the 2 packages (base and react-components) that would need to go on npm for the trace-extension to use? Do we start publishing them under the same branding as the tsp-typescript-client and timeline-chart. What's your temporary thought on this?

marcdumais-work:
> I agree we/I don't want this PR to be blocked here for too long, we have to begin somewhere and start using it. 

:+1: 

>So we could start a new repo under theia-ide (ironic) for the vscode-trace-extension that we won't publish yet on openvsx.

Why not start publishing it now or as soon as we think it's useful, even if only for early testing? It will make it so much easier to consume that way - any Theia app that uses `open-vsx.org` will be able to install it with one click. Consider the `theia-ide` GH org as a sandbox we can use, until a given component is ready, if ever, for the big leagues (e.g. an open-source Foundation). We just need to be careful picking a license that gives flexibility to re-license later (MIT) and be mindful of our dependencies' licenses, and we're good. 

> But how about the 2 packages (base and react-components) that would need to go on npm for the trace-extension to use? 

I think these two components could wait before being published on `npm`, if the immediate need is to consume them to build the extension. In the meantime they can e.g. be fetched in source form and built-in-place, before the extension. Not quite as convenient as adding them as dependencies in the extension's `package.json` but not a show-stopper. 


> Do we start publishing them under the same branding as the tsp-typescript-client and timeline-chart. What's your temporary thought on this?

I think the current names for [tsp-typescript-client](https://github.com/theia-ide/tsp-typescript-client/blob/master/package.json#L2) and [timeline-chart](https://github.com/theia-ide/timeline-chart/blob/master/timeline-chart/package.json#L2) are fine temporary names and would not necessarily need to change after an eventual move the the Eclipse Foundation. If you can find similarly good names for the two new packages, it would work. 

We've had a very similar situation with the `sprotty` library, that started under `theia-ide` and was moved to the Eclipse Foundation later. They continued to use the [same npm package name](https://www.npmjs.com/package/sprotty) after the move.

tahini:
> > I agree we/I don't want this PR to be blocked here for too long, we have to begin somewhere and start using it.
> 
> +1
> 
> > So we could start a new repo under theia-ide (ironic) for the vscode-trace-extension that we won't publish yet on openvsx.
> 
> Why not start publishing it now or as soon as we think it's useful, even if only for early testing? It will make it so much easier to consume that way - any Theia app that uses `open-vsx.org` will be able to install it with one click. Consider the `theia-ide` GH org as a sandbox we can use, until a given component is ready, if ever, for the big leagues (e.g. an open-source Foundation). We just need to be careful picking a license that gives flexibility to re-license later (MIT) and be mindful of our dependencies' licenses, and we're good.
> 
> > But how about the 2 packages (base and react-components) that would need to go on npm for the trace-extension to use?
> 
> I think these two components could wait before being published on `npm`, if the immediate need is to consume them to build the extension. In the meantime they can e.g. be fetched in source form and built-in-place, before the extension. Not quite as convenient as adding them as dependencies in the extension's `package.json` but not a show-stopper.
> 
> > Do we start publishing them under the same branding as the tsp-typescript-client and timeline-chart. What's your temporary thought on this?
> 
> I think the current names for [tsp-typescript-client](https://github.com/theia-ide/tsp-typescript-client/blob/master/package.json#L2) and [timeline-chart](https://github.com/theia-ide/timeline-chart/blob/master/timeline-chart/package.json#L2) are fine temporary names and would not necessarily need to change after an eventual move the the Eclipse Foundation. If you can find similarly good names for the two new packages, it would work.
> 
> We've had a very similar situation with the `sprotty` library, that started under `theia-ide` and was moved to the Eclipse Foundation later. They continued to use the [same npm package name](https://www.npmjs.com/package/sprotty) after the move.

@bhufmann what's your opinion on @marcdumais-work 's last comment? Shall we start a new repo called vscode-trace-extension in theia-ide for now and keep the current one for the 2 packages? I'd update this PR to reflect that.

tahini:
Since the vscode extension will be moved to its own repo, only the split of the packages will remain in this repo. PR #172 replaces this one

PR #127, opened 2020-09-24
repo: misc cleanup

vince-fugnitto:
**Description**

The pull-request includes the following changes:
- [x] updates the `engines` field similarly to [upstream theia](https://github.com/eclipse-theia/theia/blob/9a46d0ab5ad0065eb686efd2af3d38e5063828b0/package.json#L5-L8)\
 (ideally the `engines` field should be at the root `package.json` and not defined for each example app)
- [x] removes the deprecated `@theia/languages` extension which will lead to build issues regarding mix-match of `@theia` dependencies.
- [x] update `@theia` dependencies present in the `yarn.lock`.

---

cc @bhufmann the following changes should help resolve the issues you were experiencing regarding building the repo and including newer versions of `@theia`.

### DIFF
--- a/browser-app/package.json
+++ b/browser-app/package.json
@@ -42,7 +42,8 @@
     "download:plugins": "theia download:plugins -p=true"
   },
   "engines": {
-    "node": ">=10.2.0 < 13"
+    "yarn": "1.0.x || >=1.2.1",
+    "node": ">=12.14.1 <13"
   },
   "theiaPluginsDir": "plugins",
   "theiaPlugins": {
--- a/browser-app/package.json
+++ b/browser-app/package.json
@@ -22,7 +22,6 @@
     "@theia/process": "latest",
     "@theia/terminal": "latest",
     "@theia/editor": "latest",
-    "@theia/languages": "latest",
     "@theia/markers": "latest",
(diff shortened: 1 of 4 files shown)


bhufmann (approved):
Looks good to me

bhufmann:
We just have to figure out the failure of the "License Compliance" check before merging it.

browser-app/package.json
```diff
@@ -42,7 +41,8 @@
     "download:plugins": "theia download:plugins -p=true"
   },
   "engines": {
-    "node": ">=10.2.0 < 13"
+    "yarn": "1.0.x || >=1.2.1",
+    "node": ">=12.14.1 <13"
```
tahini, line 45: Does it mean we don't support version 10 anymore at all?

vince-fugnitto: > Does it mean we don't support version 10 anymore at all?

Correct, the [framework dropped](https://github.com/eclipse-theia/theia/blob/master/CHANGELOG.md#1_5_0_drop_node_10_support) support for `node 10` in the `1.5.0` release as 12 is the LTS.

tahini (approved):

PR #128, opened 2020-09-25
Handle missing labels in entry gracefully

bhufmann:


tahini (approved):
I'm not getting the exception, but sounds good!

bhufmann:
> I'm not getting the exception, but sounds good!

It only happens with a modified trace server that doesn't send the labels.

Issue #129, opened 2020-09-25
Order of entries returned by time graph data providers not kept


bhufmann:
Time Graph tree entries are sorted by id in the UI and the order returned by the time data provider is not kept. Since the ids, are arbitrary, it should not be used for sorting.

In my custom trace I have a data provider that returns the following order:

parentA (parentId: -1, id: 3)
childA (parentId:3, id 10)
childB (parentId:3, id 5)   
parentB (parentId: -1, id: 2)
childC (parentId:2, id 20)
childD (parentId:2, id 15)   

In the view it looks like:

parentB (parentId: -1, id: 2)
   childD (parentId:2, id 15)  
   childC (parentId:2, id 20)
parentA (parentId: -1, id: 3)
   childB (parentId:3, id 5)   
   childA (parentId:3, id 10)

PR #131, opened 2020-09-25
#129: Keep order of tree entries returned by the data provider

bhufmann:

### DIFF
--- a/viewer-prototype/src/browser/trace-viewer/components/utils/filtrer-tree/utils.tsx
+++ b/viewer-prototype/src/browser/trace-viewer/components/utils/filtrer-tree/utils.tsx
@@ -11,17 +11,25 @@ const entryToTreeNode = (entry: Entry) => ({
 export const listToTree = (list: Entry[]): TreeNode[] => {
     const rootNodes: TreeNode[] = [];
     const lookup: { [key: string]: TreeNode } = {};
+    // Fill-in the lookup table
     list.forEach(entry => {
         lookup[entry.id] = entryToTreeNode(entry);
     });
-    Object.keys(lookup).forEach(id => {
-        const entry = lookup[id];
-        if (entry.parentId === -1) {
-            entry.isRoot = true;
-            rootNodes.push(entry);
-        } else if (entry.parentId in lookup) {
-            const p = lookup[entry.parentId];
-            p.children.push(entry);
+    // Create the tree in the order it has been received
+    list.forEach(entry => {
+        const node = lookup[entry.id];
+        if ((entry.parentId !== undefined) && (entry.parentId !== -1)) {
+            const parent: TreeNode = lookup[entry.parentId];
+            if (parent) {
+                parent.children.push(node);
+            } else {
+                // no parent available, treat is as root node
+                node.isRoot = true;
+                rootNodes.push(node);
+            }
+        } else {
+            node.isRoot = true;
+            rootNodes.push(node);
         }
     });
     return rootNodes;


viewer-prototype/src/browser/trace-viewer/components/utils/filtrer-tree/utils.tsx
```diff
@@ -12,17 +12,26 @@ const entryToTreeNode = (entry: Entry) => ({
...
     list.forEach(entry => {
-        lookup[entry.id] = entryToTreeNode(entry);
+        const node = entryToTreeNode(entry);
+        lookup[entry.id] = node;
+        if (entry.parentId !== -1) {
+            const parent: TreeNode = lookup[entry.parentId];
+            if (parent) {
+                parent.children.push(node);
+            } else {
+                noParentList.push(node);
```
tahini, line 24: So basically, if a child comes before its parent in the entry list, you add it to a list of entries with undefined parent and after all entries have been processed, you push those children at the end of the parent? If other children for that parent came later, then the child that first came will be after them in the list, that doesn't preserve the order of entries, no? Or am I mistaken?

Shouldn't you first prepare the lookup array, then fill the entries by taking the list order again?

tahini (changes requested):
I'm realizing that the data providers don't send the entries in their expected order, so I can't appreciate this patch as much as I'd like. Do you know any of the view with an example trace that would show the difference this patch makes?

viewer-prototype/src/browser/trace-viewer/components/utils/filtrer-tree/utils.tsx
bhufmann: Ok, if a child comes before the parent then the order is not kept. I'll need to fix that.

I hope to get the order right without preparing the lookup array first which would avoid looping twice. I'll see what I can do.

bhufmann:
> I'm realizing that the data providers don't send the entries in their expected order, so I can't appreciate this patch as much as I'd like. Do you know any of the view with an example trace that would show the difference this patch makes?

In Trace Compass the ordering is done in the UI code. See for example, Resources View or Flame Chart View. I don't think the id should be the ordering key. In the theia-trace-extension we'd like to avoid that, because each data provider would need a corresponding view implementation. So, I think the initial order should come from the data provider.

bhufmann (commented):
I'll need to some updates to the ordering algorithm.

tahini:
> In Trace Compass the ordering is done in the UI code. See for example, Resources View or Flame Chart View. I don't think the id should be the ordering key. In the theia-trace-extension we'd like to avoid that, because each data provider would need a corresponding view implementation. So, I think the initial order should come from the data provider and the generic view should keep it.

Data providers will need a few updates: entry ordering is one, labels and headers are another. Also see this PR in the protocol about the parentId: https://github.com/theia-ide/trace-server-protocol/pull/34

viewer-prototype/src/browser/trace-viewer/components/utils/filtrer-tree/utils.tsx
```diff
@@ -11,17 +11,26 @@ const entryToTreeNode = (entry: Entry) => ({
...
-        if (entry.parentId === -1) {
-            entry.isRoot = true;
-            rootNodes.push(entry);
-        } else if (entry.parentId in lookup) {
-            const p = lookup[entry.parentId];
-            p.children.push(entry);
+
+    // Create the tree in the order it has been e
+    list.forEach(entry => {
+        const node = entryToTreeNode(entry);
```
tahini, line 21: const node = lookup[entry.id] ? entryToTreeNode has already been called on line 15

It's causing the hierarchy to fail since you're adding the child to the parent in the lookup list, but the root node you are adding is the one just created here.

tahini (changes requested):
The TC patch pre-sorting the entries in the data provider is now in, so the resources view now should show the entries properly sorted.

viewer-prototype/src/browser/trace-viewer/components/utils/filtrer-tree/utils.tsx
bhufmann: I'll update it. BTW, I have noticed that there is some issues for data providers in Experiments, e.g. experiment with 2 kernel traces, where in that case the Thread status or Resource status view only shows one trace. I'm looking into fixes in the Trace Compass server for that.

tahini (approved):
Looks good! Thanks!

viewer-prototype/src/browser/trace-viewer/components/utils/filtrer-tree/utils.tsx
```diff
@@ -11,17 +11,25 @@ const entryToTreeNode = (entry: Entry) => ({
...
     });
-    Object.keys(lookup).forEach(id => {
-        const entry = lookup[id];
-        if (entry.parentId === -1) {
-            entry.isRoot = true;
-            rootNodes.push(entry);
-        } else if (entry.parentId in lookup) {
-            const p = lookup[entry.parentId];
-            p.children.push(entry);
+    // Create the tree in the order it has been e
```
PatrickTasse, line 18: received?

bhufmann: Thanks for finding the typo

PatrickTasse (approved):

bhufmann:
thanks!

PR #134, opened 2020-09-27
Set Focus on Canvas to handle keypress events

### LINKED ISSUE #135: State Navigation using Right/Left Arrow doesn't work correctly
Pressing Right/Left Arrow button skips few states.

Navigation

muddana-satish:
This pull request is dependent on https://github.com/theia-ide/timeline-chart/pull/45

fixes #135 

### DIFF
--- a/viewer-prototype/src/browser/trace-viewer/components/utils/timegraph-container-component.tsx
+++ b/viewer-prototype/src/browser/trace-viewer/components/utils/timegraph-container-component.tsx
@@ -30,7 +30,7 @@ export class ReactTimeGraphContainer extends React.Component<ReactTimeGraphConta
 
     shouldComponentUpdate(nextProps: ReactTimeGraphContainer.Props): boolean {
         return nextProps.options.height !== this.props.options.height
-               || nextProps.options.width !== this.props.options.width ;
+            || nextProps.options.width !== this.props.options.width;
     }
 
     componentDidUpdate(prevProps: ReactTimeGraphContainer.Props): void {
@@ -40,6 +40,6 @@ export class ReactTimeGraphContainer extends React.Component<ReactTimeGraphConta
     }
 
     render(): JSX.Element {
-        return <canvas ref={ref => this.ref = ref || undefined} onWheel={e => e.preventDefault()}></canvas>;
+        return <canvas ref={ ref => this.ref = ref || undefined } onWheel={ e => e.preventDefault() } tabIndex={ 1 }></canvas>;
     }
 }
--- a/yarn.lock
+++ b/yarn.lock
@@ -13459,9 +13459,9 @@ timed-out@^4.0.0, timed-out@^4.0.1:
   integrity sha1-8y6srFoXW+ol1/q1Zas+2HQe9W8=
 
 timeline-chart@next:
-  version "0.2.0-next.e3c754b0"
-  resolved "https://registry.yarnpkg.com/timeline-chart/-/timeline-chart-0.2.0-next.e3c754b0.tgz#49934ed5d2c9d5d2987c996324edee386c9ee419"
-  integrity sha512-zMcDFmeTkc+t1c6rACAgfTaw9LH0ZQOXOGKxXdhxzmmn3muFd4IeIC5I7dCtU+IMJ4nvp3iZ9JHB7j5ZGlxpUQ==
+  version "0.2.0-next.924c35d7"
+  resolved "https://registry.yarnpkg.com/timeline-chart/-/timeline-chart-0.2.0-next.924c35d7.tgz#d08a9aa10fad1470ed2849fbc859304b0e74d797"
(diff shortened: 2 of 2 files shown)


muddana-satish:
This is how it works after the changes. Please review.

Navigation FIxed

tahini:
I tried it and don't see that it fixes the issue #135. I tried with 2 time graphs opened, and tried zooming in the view and still the states are skipped when navigating with left/right arrows. I also had the impression that when selecting a row in a view after having selected one in another view, the row is not really selected. But maybe I'm mistaken.

muddana-satish:
> I tried it and don't see that it fixes the issue #135. I tried with 2 time graphs opened, and tried zooming in the view and still the states are skipped when navigating with left/right arrows. I also had the impression that when selecting a row in a view after having selected one in another view, the row is not really selected. But maybe I'm mistaken.

You're right. I didn't try with multiple charts. But looks like the root cause is different now. The way we select the next state is based on the current selection range(the blue line). Since, the blue line is shared across the charts the behavior isn't consistent when we have multiple charts opened.

tahini:
ok, so what you're saying is that there are 2 different issues, one with one view (the one you're fixing here), and another one with more views? So this patch (and the one in timeline-chart) are valid for the first issue? Or do you think there's a way to solve both issues with a same fix?

muddana-satish:
> ok, so what you're saying is that there are 2 different issues, one with one view (the one you're fixing here), and another one with more views? So this patch (and the one in timeline-chart) are valid for the first issue? Or do you think there's a way to solve both issues with a same fix?

Yes. The issue I fixed happens when we open and close the chart multiple times. Each time you open the chart, it registers a eventlistener on document object. But when we close it, we don't remove them. This causes the event to register multiple times. Because of this, each eventlistener is triggered multiple times which causes it to skip few states.

At this point I do not know how to fix the issue that you pointed. If you want that to be included in this PR, then I would like to put this on hold till I am ready with the proper fix for the second issue.

tahini (changes requested):
The final changes of this branch is a one-liner that adds the tabIndex. Is this what fixes the #135 issue, as mentioned originally? And is this required for the matching timeline-chart modification to work? In any case, I have no objection. But before approving, could you squash your commits so we have the best commit message possible. Thanks!

muddana-satish:
> The final changes of this branch is a one-liner that adds the tabIndex. Is this what fixes the #135 issue, as mentioned originally? And is this required for the matching timeline-chart modification to work? In any case, I have no objection. But before approving, could you squash your commits so we have the best commit message possible. Thanks!

Yes, the change is required. Without setting tabIndex, the keypress events doesn't work on the canvas. I squashed all the commits. Could you please verify them now.

tahini (changes requested):
Works great! You'll have to update the yarn.lock once the timeline-chart change is in

muddana-satish:
@tahini  I added a yarn.lock file
This is what I did. I removed yarn.lock from my project directory (rm -rf yarn.lock) and then ran 'npx yarn' to generate it again. I then committed the newly generated file to this branch. Looks like the changes are huge. 

Let me know if I did it correctly.

tahini:
@muddana-satish You basically updated the whole project by removing yarn.lock and building it again. There is a bit of testing required if we do this. You can go back to the original yarn.lock and just remove the ~5 lines of the timeline-chart block and then do a yarn. The block will be recreated with the latest.

muddana-satish:
> @muddana-satish You basically updated the whole project by removing yarn.lock and building it again. There is a bit of testing required if we do this. You can go back to the original yarn.lock and just remove the ~5 lines of the timeline-chart block and then do a yarn. The block will be recreated with the latest.

I just committed yarn.lock. Could you please review it now. Thanks.

PatrickTasse:
I can only get it to work if I first click on the timeline chart, thus losing its previous time selection.
I would expect mouse wheel scroll/zoom to work whenever the mouse is over the timeline chart.
In that case the timeline chart should *not* have keyboard focus (if the focus was on another widget) but the mouse wheel event should contain some indication of which key modifiers were pressed at the time of the mouse scroll, that will indicate whether to scroll or zoom.

PatrickTasse:
Another thing that happens in Eclipse is that whenever you give focus to a tab (view) the application decides to which of its children widget should be given the focus.
In our case I would suggest that clicking on a trace's tab should give focus to its first (top) analysis widget.
Subsequently, an analysis widget receiving focus should give focus to one of its child widgets. In the case of the events table, it would be the table. In the case of a time analysis, it should be the timeline widget (not the tree).
This should also happen when the user clicks on a specific analysis name tab (the left part with the title), so you could click on it and immediately can navigate with the arrow keys without needing to click the timeline first.

tahini (approved):
I'm fine with this, given the scope and the bug to fix. Patrick's suggestion may be for another issue?

muddana-satish:
Thanks @PatrickTasse for pointing out these issues. I agree the mouse scroll/zoom should work without the focus on the chart. I can add those changes as part of another PR. 

I am not sure about setting the automatic focus to the widgets. Setting focus to the first or upper most widget sounds reasonable to me.

MatthewKhouzam (approved):

MatthewKhouzam:
Looks great to me!

PR #136, opened 2020-10-05
TimeGraph Rows and States deselects when zooming in/out

muddana-satish:
Remove state selection

This logic is already handled in timeline-chart. And validating states with id is incorrect. Hence removed it from this file.

This PR is dependent on https://github.com/theia-ide/timeline-chart/pull/49


tahini (changes requested):
Works well. Again please update the yarn.lock file when the timeline-chart patch is in.

muddana-satish:
> Works well. Again please update the yarn.lock file when the timeline-chart patch is in.

I updated the yarn.lock file. Could you please review it now.

bhufmann (approved):
Looks good to me

tahini (approved):
Works well, thanks!

PR #137, opened 2020-10-05
Rows do not render because of a runtime error

muddana-satish:

### DIFF
--- a/viewer-prototype/src/browser/trace-viewer/components/timegraph-output-component.tsx
+++ b/viewer-prototype/src/browser/trace-viewer/components/timegraph-output-component.tsx
@@ -145,7 +145,7 @@ export class TimegraphOutputComponent extends AbstractTreeOutputComponent<Timegr
     renderChart(): React.ReactNode {
         return <React.Fragment>
             {this.state.outputStatus === ResponseStatus.COMPLETED ?
-                <div id='timegraph-main' className='ps__child--consume' onWheel={ev => { ev.preventDefault(); ev.stopPropagation(); }} style={{ height:this.props.style.height }} >
+                <div id='timegraph-main' className='ps__child--consume' onWheel={ev => { ev.preventDefault(); ev.stopPropagation(); }} style={{ height: this.props.style.height }} >
                     {this.renderTimeGraphContent()}
                 </div> :
                 'Analysis running...'}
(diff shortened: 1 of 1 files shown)


viewer-prototype/src/browser/trace-viewer/components/timegraph-output-component.tsx
```diff
@@ -326,7 +326,7 @@ export class TimegraphOutputComponent extends AbstractTreeOutputComponent<Timegr
             }
             style = this.styleMap.get(value);
             if (!style) {
-                style = backupStyles[(value % backupStyles.length)];
+                style = backupStyles[(Math.abs(value) % backupStyles.length)];
                 this.styleMap.set(value, style);
             }
             return {
```
muddana-satish, line 332: I noticed a run time error because 'style' object can be undefined because for few traces, the 'value' could be negative. Hence we need to make sure we always get a valid 'style' object.

I couldn't reproduce this with any of the traces frequently. But saw this error a few times.

tahini (approved):
I don't have this bug, but it sounds fair. Anyway, values and styles will be revisited soon.

viewer-prototype/src/browser/trace-viewer/components/timegraph-output-component.tsx
ankusht-work: I was able to reproduce it after several attempts. Below is the error 

timegraph-output-component.tsx:333 Uncaught (in promise) TypeError: Cannot read property 'color' of undefined
    at TimegraphOutputComponent.getDefaultElementStyle (timegraph-output-component.tsx:333)
    at TimegraphOutputComponent.getElementStyle (timegraph-output-component.tsx:269)
    at Object.rowElementStyleProvider (timegraph-output-component.tsx:58)
    at TimeGraphChart.push.../node_modules/timeline-chart/lib/layer/time-graph-chart.js.TimeGraphChart.createNewRowElement (time-graph-chart.ts:320)
    at time-graph-chart.ts:300

Nevertheless, its good to have this check.

muddana-satish: Thanks. Do you know the workflow which reproduces this consistently? Or you found it in a random workflow?

ankusht-work: It was random while navigating states.

bhufmann (approved):
Looks good to me. Thanks!

Issue #144, opened 2020-10-07
Support  DATA_TREE output type of data providers


bhufmann:
The list of available analysis list all data providers outputs from the selected experiment. There are different provider types available in the TSP and currently implemented in the Trace Compass Trace server:

public enum ProviderType {
        /**
         * A provider for a table data structure implemented as virtual table.
         */
        TABLE,
        /**
         * A provider for a tree, whose entries have XY series. The x-series is time.
         */
        TREE_TIME_XY,
        /**
         * A provider for a Time Graph model, which has entries with a start and end
         * time, each entry has a series of states, arrows link from one series to
         * another
         */
        TIME_GRAPH,
        /**
         * A provider for a data tree, which has entries (rows) and columns.
         * @since 6.1
         */
        DATA_TREE
    }

The theia-trace-extension will list all available types in the list of available data provider. Providers of type TABLE, TREE_TIME_XY and TIME_GRAPH will open successfully and contribute a widget the the trace container. However, when selecting an analysis of type DATA_TREE then it fails. Moreover all open views disappear due to the usage of unsupported type.

The DATA_TREE providers should not be listed in the list of available analysis until the support is available. Hence I see 2 sets of contributions for this issue tracker:

1. Remove DATA_TREE providers from the list of available analysis
1. Implement support for DATA_TREE providers. When opening such a widget, it should not be added to the timeline container.

bhufmann:
DATA_TREE are now supported. However, the data is fetched using the fetchXY() tree, because the corresponding endpoint hasn't been implemented in Trace Compass server.

I'm closing this one ticket and open a new issue to use the data tree endpoint when available.

PR #146, opened 2020-10-08
#144: Show 'Not implemented yet!' for unsupported data provider type

bhufmann:
For example, provider type DATA_TREE is currently not implemented and
when selecting such provider (e.g. Function Duration Statistics) the
trace container crashes and the trace has to be closed first before
being able to continue.

This patch shows avoids the crash and shows a warning message instead.

### DIFF
--- a/viewer-prototype/src/browser/trace-viewer/components/null-output-component.tsx
+++ b/viewer-prototype/src/browser/trace-viewer/components/null-output-component.tsx
@@ -1,0 +1,27 @@
+import * as React from 'react';
+import { AbstractOutputComponent, AbstractOutputProps, AbstractOutputState } from './abstract-output-component';
+
+type NullOutputState = AbstractOutputState & {
+};
+
+type NullOutputProps = AbstractOutputProps & {
+};
+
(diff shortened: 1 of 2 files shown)


tahini:
I'll review this soon, but just for your information in general, in case you did not know, React has error boundaries to catch any uncaught errors in component (https://reactjs.org/docs/error-boundaries.html). I used it on another project, it's like a try-catch per component, we could add that to wrap our components and limit the error to the component that had it.

tahini (approved):
Nice addition! Thanks!

bhufmann:
> I'll review this soon, but just for your information in general, in case you did not know, React has error boundaries to catch any uncaught errors in component (https://reactjs.org/docs/error-boundaries.html). I used it on another project, it's like a try-catch per component, we could add that to wrap our components and limit the error to the component that had it.

Thanks for the information.

bhufmann:
@MatthewKhouzam are you ok with this change?

MatthewKhouzam (approved):
looks fine to me!

bhufmann:
Thanks!

Issue #35, opened 2020-05-25
Time Graph: Support display styles


tahini:
Some data provider support styles. They should be used when available

MatthewKhouzam:
To be clear, style in this case would be 

- Color
- Size
- Shape (Marker)
- image (one can dream)

Anything else?

ebugden:
⚠️**This issue is related to the Time Graph which has been moved to its own library: [theia-ide/timeline-chart](https://github.com/theia-ide/timeline-chart)**. The code related to this issue can be found in the timeline-chart library and PRs for this issue should be submitted there.

mirsky-work:
Fixed by #344?

bhufmann:
The support for timeline-charts is implemented. We might want to verify if there are still styles that can't be applied by the library. 

For the xy-charts, the styles are not supported yes, afaik.

mirsky-work:
> The support for timeline-charts is implemented. We might want to verify if there are still styles that can't be applied by the library.
> 
> For the xy-charts, the styles are not supported yes, afaik.

Right, so can we close this issue and create a separate one for xy-charts?

PR #147, opened 2020-10-08
Align code with the TSP specification

tahini:
The trace server protocol and the tsp-typescript-client have been
updated. This makes the trace extension match the newer versions.

It updates the filter-tree (left part of xy and timegraphs) to use
labels and column headers instead of extra keys.

Signed-off-by: Geneviève Bastien <gbastien@versatic.net>

### DIFF
--- a/viewer-prototype/src/browser/trace-viewer/components/data-providers/tsp-data-provider.ts
+++ b/viewer-prototype/src/browser/trace-viewer/components/data-providers/tsp-data-provider.ts
@@ -1,6 +1,5 @@
-/* eslint-disable @typescript-eslint/no-explicit-any */
 import { TspClient } from 'tsp-typescript-client/lib/protocol/tsp-client';
-import { TimeGraphEntry, TimeGraphRow, TimeGraphModel, TimeGraphState } from 'tsp-typescript-client/lib/models/timegraph';
+import { TimeGraphEntry, TimeGraphRow, TimeGraphState } from 'tsp-typescript-client/lib/models/timegraph';
 import { TimelineChart } from 'timeline-chart/lib/time-graph-model';
 import { QueryHelper } from 'tsp-typescript-client/lib/models/query/query-helper';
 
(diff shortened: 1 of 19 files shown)


tahini:
This uses the following PR in Trace Server Protocol: https://github.com/theia-ide/trace-server-protocol/pull/37

Requires the Trace Server patches: https://git.eclipse.org/r/c/tracecompass.incubator/org.eclipse.tracecompass.incubator/+/170409 and previous

And needs the TSP-typescript-client PR: https://github.com/theia-ide/tsp-typescript-client/pull/4

Also, for the XY columns, since no data provider implements it correctly yet on the server side, there's no dp that will have columns (as of today october 8th 2020)

tahini:
Also, this chain of PRs has focused on the data models returned for the various outputs. There should be another round of patches for query parameters (I did some for the table queries though) and the traces/experiment endpoints.

viewer-prototype/src/browser/trace-viewer/components/table-output-component.tsx
```diff
@@ -140,16 +139,12 @@ export class TableOutputComponent extends AbstractOutputComponent<TableOutputPro
             colIds.push(0);
         }
 
-        columnEntries.forEach(entry => {
-            const id = this.showIndexColumn ? ++entry.id : entry.id;
+        columnEntries.forEach(columnHeader => {
+            const id = this.showIndexColumn ? ++columnHeader.id : columnHeader.id;
```
MatthewKhouzam, line 143: later patch... can we split this... I really hate counting on pre/post increments.

viewer-prototype/src/browser/trace-viewer/components/data-providers/tsp-data-provider.ts
```diff
@@ -100,35 +99,34 @@ export class TspDataProvider {
...
+                        id: row.entryId + '-' + idx,
                         label: 'GAP',
                         range: {
                             start: end,
-                            end: nextState.startTime - chartStart
+                            end: nextState.start - chartStart
                         },
                         data: {
-                            stateValue: -1
+
```
MatthewKhouzam, line 129: could you comment on why a null state isn't -1?

viewer-prototype/src/browser/trace-viewer/components/timegraph-output-component.tsx
```diff
@@ -305,29 +320,31 @@ export class TimegraphOutputComponent extends AbstractTreeOutputComponent<Timegr
...
         if (modelData) {
-            const value = modelData.stateValue;
-            const elementStyle = styles[value.toString()];
-            if (elementStyle) {
+            const outputStyle = modelData.style;
+            if (!outputStyle) {
                 return {
-                    color: parseInt(elementStyle.color, 16),
-                    height: this.props.style.rowHeight * elementStyle.height,
+                    color: 0xCACACA,
```
MatthewKhouzam, line 326: Ok, so we check if there's no output style instead?

Also... should we extract all the colors as constants (separate patch)

MatthewKhouzam (commented):
I can confirm this works with updated TC.

viewer-prototype/src/browser/trace-viewer/components/data-providers/tsp-data-provider.ts
tahini: the concept of stateValue is now gone from the TSP!!... 

though as I did this patch, I'm ambivalent... stateValue was small number, the style is more verbose. But having both style and value was confusing, as it's either one or the other. And style offers more possibility. 

For now in the Trace Server, when the style is not set in the state, it's using the value's string value as a style... And if it's a null state, then there's no style at all.

viewer-prototype/src/browser/trace-viewer/components/timegraph-output-component.tsx
tahini: One of my next task will be https://github.com/theia-ide/theia-trace-extension/issues/35

I'll look into the default values and everything then. I just made it work for now.

viewer-prototype/src/browser/trace-viewer/components/data-providers/tsp-data-provider.ts
tahini: I think we could fix the style vs value using mixed type in the protocol, as described here: https://swagger.io/docs/specification/data-models/data-types/

bhufmann (approved):
It looks good to me. It works well with related changes in trace server and tsp-typescript-client.

tahini:
I just saw the approval here, thanks! So missing is only the incubator's trace server patches.

bhufmann:
@MatthewKhouzam are you ok with the change? The dependent patches (tsp-typescript-client and trace server) are merged.

MatthewKhouzam (approved):

tahini:
Merging this now, don't forget to yarn download:server again!

PR #150, opened 2020-10-13
Handle onClick event on timegraph row

muddana-satish:
Feature was working with earlier versions of PIXI.
Current PIXI is not able to handle events for the elements with opacity 0.
Changed the Opacity as a workaround fix.

Fixes https://github.com/theia-ide/timeline-chart/issues/52

This PR is dependent on https://github.com/theia-ide/timeline-chart/pull/57

### DIFF
--- a/viewer-prototype/src/browser/trace-viewer/components/timegraph-output-component.tsx
+++ b/viewer-prototype/src/browser/trace-viewer/components/timegraph-output-component.tsx
@@ -58,7 +58,8 @@ export class TimegraphOutputComponent extends AbstractTreeOutputComponent<Timegr
             rowElementStyleProvider: (model: TimelineChart.TimeGraphRowElementModel) => this.getElementStyle(model),
             rowStyleProvider: (row: TimelineChart.TimeGraphRowModel) => ({
                 backgroundColor: 0x979797,// 0xaaaaff,
-                backgroundOpacity: row.selected ? 0.1 : 0,
+                // PIXI 5 doesn't handle events on elements with opacity 0. As a workaround set it to 0.001
+                backgroundOpacity: row.selected ? 0.1 : 0.001,
                 lineColor: 0xdddddd, // hasStates ? 0xdddddd : 0xaa4444, // row.data && row.data.hasStates
                 lineThickness: 1, // hasStates ? 1 : 3 // row.data && row.data.hasStates
             })
@@ -145,7 +146,7 @@ export class TimegraphOutputComponent extends AbstractTreeOutputComponent<Timegr
     renderChart(): React.ReactNode {
         return <React.Fragment>
             {this.state.outputStatus === ResponseStatus.COMPLETED ?
-                <div id='timegraph-main' className='ps__child--consume' onWheel={ev => { ev.preventDefault(); ev.stopPropagation(); }} style={{ height:this.props.style.height }} >
+                <div id='timegraph-main' className='ps__child--consume' onWheel={ev => { ev.preventDefault(); ev.stopPropagation(); }} style={{ height: this.props.style.height }} >
                     {this.renderTimeGraphContent()}
                 </div> :
                 'Analysis running...'}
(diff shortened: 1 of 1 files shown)


muddana-satish:
This may not be the right fix. This should be handled by PIXI even for the elements with opacity 0. But as a work around I changed the opacity to 0.01. 

This is how it works after the changes:
Row select





Visually the chart looks the same.Attaching the screenshots for comparison.

**Current Version:**
(without selecting row)
image
(selecting row)
image

**After changing Opacity to 0.01:**
(without selecting row)
image
(selecting row)
image

muddana-satish:
Changed Opacity to 0.001 which is almost same as 0. Thanks @MatthewKhouzam for the suggestion.

MatthewKhouzam (approved):

MatthewKhouzam (changes requested):
Please update the timeline-chart

tahini (changes requested):
My main objection here is that we even need this patch. The day the default opacity is changed again in timeline-chart, it will have to be updated again... I'd expect the default value to come from the library and not have to specify it here, no? Like specify it only if/because the values from the library are not to our liking.

muddana-satish:
> My main objection here is that we even need this patch. The day the default opacity is changed again in timeline-chart, it will have to be updated again... I'd expect the default value to come from the library and not have to specify it here, no? Like specify it only if/because the values from the library are not to our liking.

I agree with the idea of setting all the default behaviors within the timeline-chart library. I believe this should be extended to other elements like states, labels, arrows, lines etc. Thank you for the suggestion.

The theia-ide/timeline-chart#57 is already merged and to set the default behaviors I have to make changes in timeline-chart repo. So is it okay if I open a new issue on that or do you want me to work as part of this issue?

tahini:
> I agree with the idea of setting all the default behaviors within the timeline-chart library. I believe this should be extended to other elements like states, labels, arrows, lines etc. Thank you for the suggestion.
> 
> The [theia-ide/timeline-chart#57](https://github.com/theia-ide/timeline-chart/pull/57) is already merged and to set the default behaviors I have to make changes in timeline-chart repo. So is it okay if I open a new issue on that or do you want me to work as part of this issue?

Does using the defaults from the timeline-chart involve changes to the timeline-chart repo, or just this one? What happens if you simply don't set the backgroundOpacity? Will it fallback to the timeline-chart's?

muddana-satish:
> > I agree with the idea of setting all the default behaviors within the timeline-chart library. I believe this should be extended to other elements like states, labels, arrows, lines etc. Thank you for the suggestion.
> > The [theia-ide/timeline-chart#57](https://github.com/theia-ide/timeline-chart/pull/57) is already merged and to set the default behaviors I have to make changes in timeline-chart repo. So is it okay if I open a new issue on that or do you want me to work as part of this issue?
> 
> Does using the defaults from the timeline-chart involve changes to the timeline-chart repo, or just this one? What happens if you simply don't set the backgroundOpacity? Will it fallback to the timeline-chart's?

With the current code base of timeline-chart, if we don't set any backgroundOpacity, then the library sets the default value to 1 which is incorrect because 'opacity: 1' would hide the lines inside the row. Attached screenshots of example app.

image

We should handle the default behaviors correctly. So it would require some changes to timeline-chart.

tahini:
> We should handle the default behaviors correctly. So it would require some changes to timeline-chart.

Ok, I think the best would be to handle it in timeline-chart and then fix it here so we don't need the same hack. @MatthewKhouzam  any other opinion for this?

MatthewKhouzam (approved):

MatthewKhouzam:
I agree with @tahini  and @muddana-satish 

> We should handle the default behaviors correctly. So it would require some changes to timeline-chart.

but if we're going with the hack, then this is one I can live with. :+1:

muddana-satish:
Thanks for the feedback.

I will work on setting all the default behaviors within the timeline-chart library. Since I have to make sure that I don't set any default behavior explicitly in theia-trace-extension, I would need to make changes in this repo as well. Hence keeping this PR open.

muddana-satish:
No longer needed since https://github.com/theia-ide/timeline-chart/pull/70 is merged.

PR #164, opened 2020-10-14
Added default style to GAP states 

### LINKED ISSUE #148: Gaps should have default style
The gap states (that are inferred to exist between the real states returned in the row model) are all drawn in a generic color.
In Eclipse Trace Compass, they are drawn using the style of their entry (that was returned in the tree model).

However, although right now this style is not used for any other purpose in time graphs (it is used in XY charts), the color of this style should probably refer to the tree section of the time graph (e.g. the tree text color?), and perhaps a different parameter (ie. defaultStateStyle) could be added to the protocol and/or model for this purpose. This could maybe be added to the TimeGraphRow model (state endpoint) instead of the TreeDataModel (tree endpoint).

ankusht-work:
fixes #148

### DIFF
--- a/viewer-prototype/src/browser/trace-viewer/components/data-providers/tsp-data-provider.ts
+++ b/viewer-prototype/src/browser/trace-viewer/components/data-providers/tsp-data-provider.ts
@@ -2,6 +2,7 @@ import { TspClient } from 'tsp-typescript-client/lib/protocol/tsp-client';
 import { TimeGraphEntry, TimeGraphRow, TimeGraphState } from 'tsp-typescript-client/lib/models/timegraph';
 import { TimelineChart } from 'timeline-chart/lib/time-graph-model';
 import { QueryHelper } from 'tsp-typescript-client/lib/models/query/query-helper';
+import { OutputElementStyle } from 'tsp-typescript-client/lib/models/styles';
 
 export class TspDataProvider {
 
@@ -79,7 +80,7 @@ export class TspDataProvider {
             if (timeGraphRow) {
                 newTimeGraphRows.push(timeGraphRow);
             } else {
-                const emptyRow: TimeGraphRow = {states: [{start: 0, end: 0, label: '', tags: 0}], entryId: id};
+                const emptyRow: TimeGraphRow = { states: [{ start: 0, end: 0, label: '', tags: 0 }], entryId: id };
                 newTimeGraphRows.push(emptyRow);
             }
         });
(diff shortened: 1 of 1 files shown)


viewer-prototype/src/browser/trace-viewer/components/timegraph-output-component.tsx
```diff
@@ -318,8 +318,8 @@ export class TimegraphOutputComponent extends AbstractTreeOutputComponent<Timegr
 
             if (value === -1) {
                 return {
-                    color: 0xCACACA,
-                    height: this.props.style.rowHeight * 0.5,
+                    color: style.color,
+                    height: style.height,
                     borderWidth: element.selected ? 2 : 0,
                     borderColor: 0xeef20c
```
muddana-satish, line 324: I don't think we need this 'if' block any more. Lines 345 - 350 should take care of this?

tahini:
See my comment in #148 . I'm not sure this is an improvement over current situation, but that's just my personal opinion ;-)

ankusht-work:
> See my comment in #148 . I'm not sure this is an improvement over current situation, but that's just my personal opinion ;-)

So not sure if this is an improvement per se,  but what am doing here is that instead of assigning a default color (light grey) for the gap states, I am retrieving the style object for each entry from the Tree Data provider and assigning it to the state's color which is exactly what we are doing in trace compass as well.

viewer-prototype/src/browser/trace-viewer/components/timegraph-output-component.tsx
ankusht-work: Yes, thanks for pointing this out. Updated the code

PatrickTasse:
> See my comment in #148 . I'm not sure this is an improvement over current situation, but that's just my personal opinion ;-)

I'm not sure I understand, are you saying that it's better to use gray for gaps because it makes obvious that we don't know the real state? It could still flicker between gray and the real state colors as the user zooms in.

Gray could still be the 'default' default color if none is provided by the application. But there are some cases where all the states on a particular row are necessarily of the same color/style (in that particular application), and then providing it as the default state color/style will prevent the flicker.

If I don't understand correctly, please explain the advantages of the current situation.

tahini:
> > See my comment in #148 . I'm not sure this is an improvement over current situation, but that's just my personal opinion ;-)
> 
> I'm not sure I understand, are you saying that it's better to use gray for gaps because it makes obvious that we don't know the real state? It could still flicker between gray and the real state colors as the user zooms in.

Yes, that's what I'm saying. Of course, if the entry has a color, then it can use that color, but for now, we don't have any entries with color. And the current grey color can be used as a fallback style.

bhufmann (approved):
Looks good to me. Thanks

ankusht-work:
> > > See my comment in #148 . I'm not sure this is an improvement over current situation, but that's just my personal opinion ;-)
> > 
> > 
> > I'm not sure I understand, are you saying that it's better to use gray for gaps because it makes obvious that we don't know the real state? It could still flicker between gray and the real state colors as the user zooms in.
> 
> Yes, that's what I'm saying. Of course, if the entry has a color, then it can use that color, but for now, we don't have any entries with color. And the current grey color can be used as a fallback style.

Updated the code based on your suggestion. If the entry has a color then it will apply that and if not it will set it to the default grey color with the default height for Gap states. Please review the changes and let me know your thoughts

PatrickTasse (approved):

PR #170, opened 2020-10-23
Implement Tooltip

maria-ericsson:
Signed-off-by: maria-ericsson <maria.chowdhury@ericsson.com>

### DIFF
--- a/viewer-prototype/src/browser/style/output-components-style.css
+++ b/viewer-prototype/src/browser/style/output-components-style.css
@@ -1,96 +1,94 @@
 /* Output component styling */
 /* Main container*/
 .output-container {
-    display: flex;
-    color: var(--theia-ui-font-color0)
+   display: flex;
+   color: var(--theia-ui-font-color0)
 }
 
 .widget-handle {
-    background-color: var(--theia-layout-color4);
-    display: grid;
-    grid-template-rows: 25px 1fr;
+   background-color: var(--theia-layout-color4);
+   display: grid;
+   grid-template-rows: 25px 1fr;
 }
 .title-bar-label {
-    text-align: center;
-    writing-mode: vertical-rl;
-    height: 50%;
-    min-height: 150px;
-    transform: rotate(180deg);
-    user-select: none;
-    align-self: center;
-    justify-self: center;
+   text-align: center;
+   writing-mode: vertical-rl;
+   height: 50%;
+   min-height: 150px;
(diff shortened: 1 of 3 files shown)


mirsky-work:
tooltips

Nice, but there are some issues that are demonstrated in the attached Gif taken from Gitpod:
- The tooltip seems to open up at a fixed position relative to the state, and not where the cursor is.
- It doesn't disappear when the cursor is moved. It does disappear when moving away from the chart.

Not an issue, but rather a question - what would it take to make the tooltip text "copyable", as it's already "selectable"?

maria-ericsson:
> tooltips
> 
> Nice, but there are some issues that are demonstrated in the attached Gif taken from Gitpod:
- Thank you. 
> * The tooltip seems to open up at a fixed position relative to the state, and not where the cursor is.
- Tooltip position is updated.
> * It doesn't disappear when the cursor is moved. It does disappear when moving away from the chart.
> 
> Not an issue, but rather a question - what would it take to make the tooltip text "copyable", as it's already "selectable"?

Tooltip disappears 

- When the mouse leaves the chart

- Between selection of two state elements. There are two reasons for that

1. If tooltip contains html links, then the user should be able to click link to open the relevant web-page
2. Tooltip content should be copyable

maria-ericsson:
> > tooltips
> > Nice, but there are some issues that are demonstrated in the attached Gif taken from Gitpod:
> 
> * Thank you.
> 
> > * The tooltip seems to open up at a fixed position relative to the state, and not where the cursor is.
> 
> * Tooltip position is fixed for a specific state element.
> * Tooltip position changes when a different state element is selected.
> 
> > * It doesn't disappear when the cursor is moved. It does disappear when moving away from the chart.
> > 
> > Not an issue, but rather a question - what would it take to make the tooltip text "copyable", as it's already "selectable"?
> 
> Tooltip disappears
> 
> * When the mouse leaves the chart
> * Between selection of two state elements. There are two reasons for that
> 
> 1. If tooltip contains html links, then the user should be able to click link to open the relevant web-page
> 2. Tooltip content should be copyable

--- tooltip close operation is added with a timer. 
---The tooltip content is copy-able and any URL that may be present in the tooltip is clickable. 
--- tooltip closes when the mouse leaves the timeline chart
--- tooltip closes when the mouse leaves the selected element.

bhufmann:
Thanks. Feature implemented by #283.

PR #174, opened 2020-10-27
Upgrade react-grid-layout to ^1.1.0

tahini:
The version of react-grid-layout used by the application used the
renamed componentWillReceiveProps method and resulted in a warning when
running the extension. The warning suggests upgrading the package.

Signed-off-by: Geneviève Bastien <gbastien@versatic.net>


bhufmann (approved):
Looks good to me

PR #177, opened 2020-11-02
Configure tsp url

tahini:


.gitpod.yml
```diff
@@ -21,7 +22,9 @@ tasks:
     yarn download:sample-traces &&
     yarn
   command: >
-    yarn start:electron ../TraceCompassTutorialTraces/ --hostname=0.0.0.0 --ignore-gpu-blacklist --no-sandbox
+    export GITPOD_URL=$(gp url) &&
+    export TRACE_SERVER_URL=$(echo "$GITPOD_URL" | sed -E "s/(http[s]:\/\/)(.*)/\18080-\2\/tsp\/api/") &&
+    yarn start:browser ../TraceCompassTutorialTraces/ --hostname=0.0.0.0 --ignore-gpu-blacklist --no-sandbox
```
marcdumais-work, line 27: `--ignore-gpu-blacklist` and ` --no-sandbox` are Electron-specific and can be removed

.gitpod.yml
```diff
@@ -2,6 +2,7 @@ image:
   file: .gitpod.dockerfile
 ports:
 - port: 3000 # Trace viewer example application
+  onOpen: ignore
```
marcdumais-work, line 5: This prevents the pop-up, that you can click to open the example application. I was wondering why it did not show.

.gitpod.yml
```diff
@@ -21,7 +22,9 @@ tasks:
     yarn download:sample-traces &&
     yarn
   command: >
-    yarn start:electron ../TraceCompassTutorialTraces/ --hostname=0.0.0.0 --ignore-gpu-blacklist --no-sandbox
+    export GITPOD_URL=$(gp url) &&
+    export TRACE_SERVER_URL=$(echo "$GITPOD_URL" | sed -E "s/(http[s]:\/\/)(.*)/\18080-\2\/tsp\/api/") &&
```
marcdumais-work, line 26: Nicely done!

marcdumais-work: `18080` - trace server port? Maybe parameterise this too?

marcdumais-work:
@tahini ATM I do not see much in the Trace views, but maybe I am not looking at the correct trace / correct view(s). Can you add info in the description about how to minimally test?

.gitpod.yml
tahini: actually, gp url 8080 gives the url directly, I'll use that instead.

tahini (commented):
Thanks for the feedback. Did you open the trace in gitpod? What do you or don't you see exactly?

MatthewKhouzam (approved):
I tried this on a different port, that works, does it also work on different machines? I didn't install yet remote. I'm reading the code, I see no reason it would _not_ work on a different machine.

Unrelated question. should we pick another port than 8080? everyone takes 8080

marcdumais-work:
@MatthewKhouzam let the [power of the Cloud](https://gitpod.io/#https://github.com/theia-ide/theia-trace-extension/pull/177) (*) guide you. 

(*) provided by Gitpod



> Thanks for the feedback. Did you open the trace in gitpod? What do you or don't you see exactly?

Like graphs and stuff. I just open a trace chosen randomly and then add a few analysis carefully chosen the same way. I had not been lucky so far, but I just retried and got a couple of things working:

image

Nice to be able to do this with a remote backend and trace server!

The resource usage looks more normal too this time:
image

Nice work!

marcdumais-work (approved):
Works nicely!

PR #179, opened 2020-11-03
Hide hidden columns from trace viewer

MatthewKhouzam:
Use the "type" parameter to encode hidden columns.

### DIFF
--- a/viewer-prototype/src/browser/trace-viewer/components/table-output-component.tsx
+++ b/viewer-prototype/src/browser/trace-viewer/components/table-output-component.tsx
@@ -5,7 +5,10 @@ import { AgGridReact } from 'ag-grid-react';
 import { ColDef, IDatasource, GridReadyEvent } from 'ag-grid-community';
 import { QueryHelper } from 'tsp-typescript-client/lib/models/query/query-helper';
 import { cloneDeep } from 'lodash';
-
+interface ColumnHeaderType {
+    hidden: boolean;
+    tag: string;
+}
 type TableOuputState = AbstractOutputState & {
     tableColumns: ColDef[];
 };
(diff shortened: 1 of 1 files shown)


viewer-prototype/src/browser/trace-viewer/components/table-output-component.tsx
```diff
@@ -159,5 +171,6 @@ export class TableOutputComponent extends AbstractOutputComponent<TableOutputPro
         this.setState({
             tableColumns: this.columnArray
         });
+
```
MatthewKhouzam, line 174: remove

tahini (changes requested):
Do I understand correctly that the type of the column header's type field should be an object? In the TSP, it is now a string. While this change probably works fine, iiuc, it won't match the TSP's specification anymore? We have to be careful to fix the TSP very soon, to avoid the trace server answering to this specific client and bypassing the TSP. Or am I wrong in my understanding of the patch?

viewer-prototype/src/browser/trace-viewer/components/table-output-component.tsx
```diff
@@ -140,12 +143,21 @@ export class TableOutputComponent extends AbstractOutputComponent<TableOutputPro
...
+            }
+
             const id = this.showIndexColumn ? ++columnHeader.id : columnHeader.id;
             colIds.push(id);
             columnsArray.push({
                 headerName: columnHeader.name,
                 field: columnHeader.id.toString(),
-                width: this.props.columnWidth
+                width: this.props.columnWidth,
+                hide: hiddenColumn
```
muddana-satish, line 159: According to the ag-grid, the hide property accepts boolean.
https://www.ag-grid.com/javascript-grid-column-properties/

In our case, the hidden column is set to string after JSON.parse. It's good to set to boolean though it works in this case.

viewer-prototype/src/browser/trace-viewer/components/table-output-component.tsx
```diff
@@ -140,12 +143,21 @@ export class TableOutputComponent extends AbstractOutputComponent<TableOutputPro
         }
 
         columnEntries.forEach(columnHeader => {
+            const typeString: string | null = columnHeader.type;
+            let hiddenColumn = false;
+            if (typeString) {
+                const type: ColumnHeaderType = JSON.parse(typeString);
+                hiddenColumn = type.hidden;
```
muddana-satish, line 150: check for undefined when the 'hidden' property doesn't exist.

bhufmann:
@MatthewKhouzam if the column is hidden, is there a way to show them in the front-end with the current implementation?

hoangphamEclipse:
@MatthewKhouzam Is this PR still valid?

bhufmann:
@MatthewKhouzam, the column model in the TSP doesn't have the hidden flag at the moment. If we'd like to support to allow the backend to hide columns by default this would have to be added to the TSP, trace server and front-end. 

Shall this one be issue be closed for now. We can work on it when it becomes priority?

Issue #149, opened 2020-10-09
Trace Explorer and Trace Viewer tabs out of sync


PatrickTasse:
From what I can gather:

- The Trace Explorer shows experiments that are opened in the trace server
- The Open Trace command in the file Explorer will open a Trace Viewer tab
- The initialisation of the Trace Viewer tab will create and/or open the experiment in the trace server, which should add it to Trace Explorer
- Closing a Trace Viewer tab will close and delete the experiment in the trace server, which should remove it from Trace Explorer
- When the browser is closed and reopened, its previously opened Trace Viewer tabs will be restored, and their initialisation will create and/or open the experiment in the trace server

Some observations:
- When opening a new browser tab, sometimes the Trace Explorer experiments are restored, sometimes they are not. Sometimes the Trace Viewer tabs are restored, sometimes they are not.
- It can happen that a Trace Explorer experiment is visible, but it has no corresponding Trace Viewer tab.
- It can happen that selecting a Trace Explorer experiment will populate the Available Analysis, sometimes not
- It can happen that after closing a Trace Viewer tab in one browser tab (experiment is removed from the Trace Explorer), then opening a new browser tab the removed experiment comes back in the new browser tab's Trace Explorer.
- If the trace server is killed then a Trace Viewer tab is closed, then the experiment does not come back after relaunching and reconnecting to the trace server. However the experiment's internal files are lingering in the trace server.

I see two approaches going forward with the Trace Explorer:

1) show Opened experiments

- At initialisation of the Trace Explorer then all opened experiments should have a corresponding Trace Viewer tab created, if one has not already been restored.
- When fetching the experiments from the trace server, all created experiments should be opened if not already opened, and returned.

2) show Created experiments

- Trace Explorer shows all created experiments whether they are opened or not.
- Need a new TSP request that returns created experiments.
- "Open" context menu and/or double-click on an experiment will open it.
- "Delete" context menu on an experiment to delete it from the trace server and remove it from the Trace Explorer.
- Closing a Trace Viewer tab only closes the experiment, does not delete it.
- Need a new TSP request to close an experiment.

bhufmann:
Releated to #140

bhufmann:
Thanks for the elaborate description and suggestions. I think this items needs to be split in multiple sub-issues to track it easilier.

mirsky-work:
For the MVP we'll implement the FE parts of proposal 2 above, without handlining closing an experiment on the server.

[...]

PR #180, opened 2020-11-05
Sync Experiments in Explorer and Viewer

### LINKED ISSUE #140: Better handle multiple open traces
It is currently possible to open multiple traces separately, but the trace explorer is not in sync. with the trace container in focus.

To Reproduce:

- Use "Open Trace(s)" menu to open trace1 -> trace1 is opened in trace container
- Open Trace Explorer
- Add a view -> view added and populated
- Go back to File Explorer and use "Open Trace(s)" menu to open trace2 -> trace is opened in trace container
- Select Trace Explorer
- Select trace container for trace2
- Open view -> view is added to trace container of **trace 1** and not for **trace 2**

Looking closely at the Trace Explorer, then the trace 1 is selected and the list of available analysis of trace 1 shown.

When selecting trace container for trace 2 (set focus) doesn't update the selection in the Trace Explorer to trace 2 as well as the available analysis are not updated. 

The trace container in focus and the content of the Trace Explorer need to be in sync.

muddana-satish:
fixes #140

### DIFF
--- a/viewer-prototype/src/browser/trace-explorer/trace-explorer-widget.tsx
+++ b/viewer-prototype/src/browser/trace-explorer/trace-explorer-widget.tsx
@@ -12,6 +12,8 @@ import { EditorManager, EditorOpenerOptions } from '@theia/editor/lib/browser';
 import URI from '@theia/core/lib/common/uri';
 import { Experiment } from 'tsp-typescript-client/lib/models/experiment';
 import { ExperimentManager } from '../../common/experiment-manager';
+/* FIXME: This may cause Circular dependency between trace-viewer and trace-explorer-widget */
+import { TraceViewerWidget } from '../trace-viewer/trace-viewer';
 
 export const TRACE_EXPLORER_ID = 'trace-explorer';
 export const TRACE_EXPLORER_LABEL = 'Trace Explorer';
@@ -49,10 +51,13 @@ export class TraceExplorerWidget extends ReactWidget {
     private sharingLink = '';
 
     private tooltip: { [key: string]: string } = {};
+    private selectedExperiment: Experiment | undefined;
 
-    // Open output
     private static outputAddedEmitter = new Emitter<OutputAddedSignalPayload>();
     public static outputAddedSignal = TraceExplorerWidget.outputAddedEmitter.event;
+
+    private static experimentSelectedEmitter = new Emitter<Experiment>();
+    public static experimentSelectedSignal = TraceExplorerWidget.experimentSelectedEmitter.event;
 
     constructor(
         @inject(ExperimentManager) private experimentManager: ExperimentManager,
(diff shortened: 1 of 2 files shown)


muddana-satish:
Here are the scenarios I have tested:

Viewer -> Explorer:

Kapture 2020-11-05 at 13 17 08

Explorer -> Viewer:

Kapture 2020-11-05 at 13 22 03


Refresh:

It always selects the first experiment in the explorer. But will retain the correct experiment after everything is loaded. Not a good UX. Any thoughts on this?

Kapture 2020-11-05 at 13 27 09

SplitWindow:

Only one experiment is selected in the explorer though there can be multiple tabs that can be visible because of split window. By default we always select the last window's active experiment.

Kapture 2020-11-05 at 13 32 48

Experiment Added:
Kapture 2020-11-05 at 13 46 37

Experiment Closed:
Kapture 2020-11-05 at 13 49 21


Please let me know if there are any other scenarios that I missed.

muddana-satish:
Note: I have not covered any of the issues mentioned in #149. I will create a new PR for that.

viewer-prototype/src/browser/trace-explorer/trace-explorer-widget.tsx
```diff
@@ -12,6 +12,7 @@ import { EditorManager, EditorOpenerOptions } from '@theia/editor/lib/browser';
 import URI from '@theia/core/lib/common/uri';
 import { Experiment } from 'tsp-typescript-client/lib/models/experiment';
 import { ExperimentManager } from '../../common/experiment-manager';
+import { TraceViewerWidget } from '../trace-viewer/trace-viewer';
```
tahini, line 15: Just highlighting this is adding a circular dependency between trace-viewer and trace-explorer-widget. It is definitely not a good practice in general, even though it works here.

With the split in packages and vscode extension, this is the kind of interaction that may well be re-architectured (how signals are sent between components, etc), so you don't have to try to fix it just yet. But could you add a FIXME above this line, just so we can come back to this eventually?

tahini (approved):
Works nicely, thanks!

mirsky-work:
Works great, thanks!

Can we have a visual indication which tab is active in a split window?
active_tab

viewer-prototype/src/browser/trace-explorer/trace-explorer-widget.tsx
muddana-satish: Thank you for sharing this.

muddana-satish:
> Can we have a visual indication which tab is active in a split window?

The focus on tabs in other theia-apps is working as expected (highlights the active tab name). But for some reason it isn't working in trace-extension. I would suggest creating a new issue for that (since it was never a working feature in trace-extension). 

Created #182. Thanks @mirsky-work for verifying and pointing out this issue.

viewer-prototype/src/browser/trace-explorer/trace-explorer-widget.tsx
muddana-satish: Added a `FIXME` comment. Thanks.

bhufmann (approved):
Nice improvement! Thanks for the contribution.

PR #185, opened 2020-11-19
Simplify openExperiment conflict handling

PatrickTasse:
Use the same code for the first and subsequent attempts.


bhufmann (changes requested):
Works fine but needs rebase

bhufmann (approved):
Looks good to me

PR #190, opened 2020-11-26
Fix and simplify openTrace conflict handling

PatrickTasse:
Use the same code for the first and subsequent attempts.

Set the name from traceURI if traceName is not specified.

Fix undefined base name in retry attempts.


tahini (approved):
Code looks good and it works. But I haven't been able to create a conflict. How can you create a conflict?

PatrickTasse:
> Code looks good and it works. But I haven't been able to create a conflict. How can you create a conflict?

Post two traces with different paths but that are in the same parent folder, and provide the same name for both. The theia-trace-extension will never do this because it sets the name to the path's last segment. So I tested by manually overriding the received name in the trace-server.

PR #192, opened 2020-12-01
Close traces after experiment and handle deletion rejection

PatrickTasse:
Wait for experiment deletion response to be received before attempting
to close its traces.

Handle trace and experiment deletion response and do not remove it from
its manager if the deletion was rejected by the trace server.

### DIFF
--- a/packages/base/src/experiment-manager.ts
+++ b/packages/base/src/experiment-manager.ts
@@ -3,6 +3,7 @@ import { TspClient } from 'tsp-typescript-client/lib/protocol/tsp-client';
 import { Query } from 'tsp-typescript-client/lib/models/query/query';
 import { OutputDescriptor } from 'tsp-typescript-client/lib/models/output-descriptor';
 import { Experiment } from 'tsp-typescript-client/lib/models/experiment';
+import { TraceManager } from './trace-manager';
 import { TspClientResponse } from 'tsp-typescript-client/lib/protocol/tsp-client-response';
 import { signalManager, Signals } from './signal-manager';
 
@@ -10,11 +11,15 @@ export class ExperimentManager {
 
     private fOpenExperiments: Map<string, Experiment> = new Map();
     private fTspClient: TspClient;
+    private fTraceManager: TraceManager;
 
     constructor(
-        tspClient: TspClient
+        tspClient: TspClient,
+        traceManager: TraceManager
     ) {
         this.fTspClient = tspClient;
+        this.fTraceManager = traceManager;
+        signalManager().on(Signals.EXPERIMENT_CLOSED, ({experiment}) => this.onExperimentClosed(experiment));
     }
 
     /**
@@ -130,6 +135,17 @@ export class ExperimentManager {
             if (deletedExperiment) {
                 signalManager().emit(Signals.EXPERIMENT_CLOSED, {experiment: deletedExperiment});
             }
+        }
+    }
+
+    private onExperimentClosed(experiment: Experiment) {
+        /*
+         * TODO: Do not close traces used by another experiment
+         */
+        // Close each trace
+        const traces = experiment.traces;
+        for (let i = 0; i < traces.length; i++) {
+            this.fTraceManager.closeTrace(traces[i].UUID);
         }
     }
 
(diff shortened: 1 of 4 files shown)


packages/base/src/experiment-manager.ts
```diff
@@ -125,10 +125,12 @@ export class ExperimentManager {
     async closeExperiment(experimentUUID: string): Promise<void> {
         const experimentToClose = this.fOpenExperiments.get(experimentUUID);
         if (experimentToClose) {
-            await this.fTspClient.deleteExperiment(experimentUUID);
-            const deletedExperiment = this.removeExperiment(experimentUUID);
-            if (deletedExperiment) {
-                signalManager().emit(Signals.EXPERIMENT_CLOSED, {experiment: deletedExperiment});
+            const deleteResponse = await this.fTspClient.deleteExperiment(experimentUUID);
+            if (deleteResponse.isOk()) {
```
bhufmann, line 129: What is the expected behavior when it fails? There could be different behaviors depending on the return code.  Maybe this needs to be handled in a different issue about error handling.

packages/base/src/trace-manager.ts
```diff
@@ -1,5 +1,6 @@
 
 import { Trace } from 'tsp-typescript-client/lib/models/trace';
+import { Experiment } from 'tsp-typescript-client/lib/models/experiment';
```
bhufmann, line 3: this adds a circular dependency between experiment manager and trace manager. Could you listen to the signal in the Experiment manager and then call the delete trace on the trace manager instead?

packages/base/src/trace-manager.ts
```diff
@@ -120,10 +122,12 @@ export class TraceManager {
     async closeTrace(traceUUID: string): Promise<void> {
         const traceToClose = this.fOpenTraces.get(traceUUID);
         if (traceToClose) {
-            await this.fTspClient.deleteTrace(traceUUID);
-            const deletedTrace = this.removeTrace(traceUUID);
-            if (deletedTrace) {
-                signalManager().emit(Signals.TRACE_CLOSED, {trace: deletedTrace});
+            const deleteResponse = await this.fTspClient.deleteTrace(traceUUID);
+            if (deleteResponse.isOk()) {
```
bhufmann, line 126: Same question here, what is the expected behavior when it fails? Maybe this needs to be handled in a different issue about error handling.

bhufmann (changes requested):

packages/base/src/trace-manager.ts
PatrickTasse: The important case here is to keep the trace in the manager if CONFLICT 409 is returned, so that it can still be deleted later when the last experiment that contains this trace is closed.

packages/base/src/trace-manager.ts
PatrickTasse: Actually it's only a dependency on the experiment model, not the experiment manager. Is it OK?

I couldn't do it in the experiment manager because it does not have any way to get a reference to the trace manager (that I know of)

packages/base/src/experiment-manager.ts
bhufmann: Let's remove it and take it on when working on error handling

packages/base/src/trace-manager.ts
bhufmann: I had the impression that experiment manager has already access to Trace Manager. I wonder what makes more sense, trace manager references experiment manager or vice versa. Same question is about experiment model reference to trace model or vice versa.

To get access to the trace manager you could do:
@inject(TspClientProvider) private tspClientProvider: TspClientProvider
and then:  tspClientProvider.getTraceManager().

packages/base/src/trace-manager.ts
bhufmann: Maybe for now we keep the trace for 409, otherwise we remove. Other error handling can be handled as a separate task.

packages/base/src/trace-manager.ts
bhufmann: I realized that using dependency injection would create a dependency to viewer-prototype and hence Theia which would make it difficult to migrate to VsCode.

bhufmann (approved):
Looks good to me

PR #196, opened 2020-12-03
Provide progress monitor for opening an experiment with cancel option

### LINKED ISSUE #195: Provide a progress monitor for opening traces
The current implementation of the trace opening, that from a root directory all relevant CTF traces are detected and they are opened as an experiment together. It would be good to provide a progress monitor with cancel option for the user in case a large directory structure is selected.

bhufmann:
In case a user selects a directory with large directory structure,
allow user to cancel the open procedure.

Fixes #195

### DIFF
--- a/viewer-prototype/src/browser/trace-viewer/trace-viewer.tsx
+++ b/viewer-prototype/src/browser/trace-viewer/trace-viewer.tsx
@@ -1,4 +1,4 @@
-import { Path } from '@theia/core';
+import { MessageService, Path } from '@theia/core';
 import { FileSystem, FileStat } from '@theia/filesystem/lib/common/filesystem';
 import { ApplicationShell, Message, StatusBar } from '@theia/core/lib/browser';
 import { ReactWidget } from '@theia/core/lib/browser/widgets/react-widget';
@@ -48,7 +48,8 @@ export class TraceViewerWidget extends ReactWidget {
         @inject(StatusBar) private statusBar: StatusBar,
         @inject(FileSystem) private readonly fileSystem: FileSystem,
         @inject(ApplicationShell) protected readonly shell: ApplicationShell,
-        @inject(TheiaMessageManager) private readonly _signalHandler: TheiaMessageManager
+        @inject(TheiaMessageManager) private readonly _signalHandler: TheiaMessageManager,
+        @inject(MessageService) protected readonly messageService: MessageService
     ) {
         super();
         this.uri = new Path(this.options.traceURI);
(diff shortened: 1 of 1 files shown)


bhufmann:
Picture below shows the dialog opened.

The text in the dialog is updated for the 3 steps during the opening process:

1. Open traces in an experiment: Finding traces
1. Open traces in an experiment: Opening traces
1. Open traces in an experiment: Creating experiment

image

viewer-prototype/src/browser/trace-viewer/trace-viewer.tsx
```diff
@@ -71,43 +72,79 @@ export class TraceViewerWidget extends ReactWidget {
...
-                tracesArray.push(this.uri);
-            }
-        }
 
-        const traces = new Array<Trace>();
+         const isCancelled = { value: false };
+
+        // This will show a progress dialog with "Cancel" option
+        this.messageService.showProgress({
+            text: 'Open traces in an experiment:'
```
bhufmann, line 83: remove experiment from user-facing string

tahini (changes requested):
That's great improvement!

But I have an issue when cancelling an operation, it does not close the corresponding tab, so it is not possible to open it again after that.

Also, there are 2 parts here: 1- Monitor the cancellable operation and determine the message to show and 2- Display the message to the user.  While 2- belongs to the theia trace viewer, 1- should be in the base package. Maybe it's a bit tight for the MVP. If so could you add a TODO for later and open a corresponding issue to track this?

bhufmann:
> That's great improvement!
> 
> But I have an issue when cancelling an operation, it does not close the corresponding tab, so it is not possible to open it again after that.
> 
True. You need to close the tab first and the it will work again. I try to find a way to close the widget as well.

> Also, there are 2 parts here: 1- Monitor the cancellable operation and determine the message to show and 2- Display the message to the user. While 2- belongs to the theia trace viewer, 1- should be in the base package. Maybe it's a bit tight for the MVP. If so could you add a TODO for later and open a corresponding issue to track this?
I don't think I understand your comment. This PR is meant to monitor the progress on higher level. If you mean to add ways to cancel deeper in the execution stack (e.g. cancelling a REST command) then it's out of scope of this PR. Cancellation of REST commands as well as error handling needs to be tracked and handled outside PR.

MatthewKhouzam (approved):
Works here, you have my approval when Genevieve's comments are addressed.

bhufmann:
> Works here, you have my approval when Genevieve's comments are addressed.

I addressed the first comment in the source code. I replied also the second comment in https://github.com/theia-ide/theia-trace-extension/pull/196#issuecomment-742737213.

MatthewKhouzam:
That's what I understood, @tahini are your concerns addressed?

tahini (approved):
Yes, my comments were adressed, cancellation works well now! As for the other comment, when we also have a vscode extension, it will be clearer if we can/should move some code to base package.

PR #197, opened 2020-12-07
Change widget colors based on theia themes

### LINKED ISSUE #76: Support themes in the theia-trace-extension and it's graphs
Theia supports changing the theme, however, the theia-trace-extension doesn't apply it. The current implementation only supports the dark them. When switching to a light theme, the background of all the different graph types is still black.

muddana-satish:
fixes #76

### DIFF
--- a/packages/react-components/src/components/trace-context-component.tsx
+++ b/packages/react-components/src/components/trace-context-component.tsx
@@ -20,6 +20,7 @@ import { XYOutputComponent } from './xy-output-component';
 import { NullOutputComponent } from './null-output-component';
 import { AbstractOutputProps } from './abstract-output-component';
 import * as Messages from '@trace-viewer/base/lib/message-manager';
+import { signalManager, Signals } from '@trace-viewer/base/lib/signal-manager';
 
 const ResponsiveGridLayout = WidthProvider(Responsive);
 
@@ -31,6 +32,7 @@ interface TraceContextProps {
     // Introduce dependency on Theia maybe it should be just a callback
     messageManager: Messages.MessageManager;
     addResizeHandler: (handler: () => void) => void;
+    backgroundTheme: string;
 }
 
 interface TraceContextState {
@@ -41,6 +43,7 @@ interface TraceContextState {
     experiment: Experiment
     traceIndexing: boolean;
     style: OutputComponentStyle;
+    backgroundTheme: string;
 }
 
 export class TraceContextComponent extends React.Component<TraceContextProps, TraceContextState> {
@@ -84,11 +87,12 @@ export class TraceContextComponent extends React.Component<TraceContextProps, Tr
                 chartWidth: this.DEFAULT_CHART_WIDTH,
                 height: this.DEFAULT_COMPONENT_HEIGHT,
                 rowHeight: this.DEFAULT_COMPONENT_ROWHEIGHT,
-                naviBackgroundColor: 0x3f3f3f,
-                chartBackgroundColor: 0x3f3f3f,
+                naviBackgroundColor: this.props.backgroundTheme === 'light' ? 0xf4f7fb : 0x3f3f3f,
+                chartBackgroundColor: this.props.backgroundTheme === 'light' ? 0xf4f7fb : 0x3f3f3f,
                 cursorColor: 0x259fd8,
-                lineColor: 0xbbbbbb
-            }
+                lineColor: this.props.backgroundTheme === 'light' ? 0x757575 : 0xbbbbbb
+            },
+            backgroundTheme: this.props.backgroundTheme
         };
         const absoluteRange = traceRange.getDuration();
         this.unitController = new TimeGraphUnitController(absoluteRange, { start: 0, end: absoluteRange });
@@ -106,6 +110,23 @@ export class TraceContextComponent extends React.Component<TraceContextProps, Tr
         this.unitController.onViewRangeChanged(viewRangeParam => { this.handleViewRangeChange(viewRangeParam); });
         this.traceContextContainer = React.createRef();
         this.initialize();
+        signalManager().on(Signals.THEME_CHANGED, (theme: string) => this.updateBackgroundTheme(theme));
+    }
+
+    public updateBackgroundTheme(theme: string) {
+        this.setState({
+            style: {
+                width: this.DEFAULT_COMPONENT_WIDTH,
+                chartWidth: this.DEFAULT_CHART_WIDTH,
+                height: this.DEFAULT_COMPONENT_HEIGHT,
+                rowHeight: this.DEFAULT_COMPONENT_ROWHEIGHT,
+                naviBackgroundColor: theme === 'light' ? 0xf4f7fb : 0x3f3f3f,
+                chartBackgroundColor: theme === 'light' ? 0xf4f7fb : 0x3f3f3f,
+                cursorColor: 0x259fd8,
+                lineColor: theme === 'light' ? 0x757575 : 0xbbbbbb
+            },
+            backgroundTheme: theme
+        });
     }
 
     private async initialize() {
@@ -150,6 +171,7 @@ export class TraceContextComponent extends React.Component<TraceContextProps, Tr
     }
 
     componentWillUnmount(): void {
+        signalManager().off(Signals.THEME_CHANGED, (theme: string) => this.updateBackgroundTheme(theme));
         this.props.messageManager.removeStatusMessage(this.INDEXING_STATUS_BAR_KEY);
         this.props.messageManager.removeStatusMessage(this.TIME_SELECTION_STATUS_BAR_KEY);
     }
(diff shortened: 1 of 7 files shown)


muddana-satish:
This is how it looks after the changes for the Theia(light) theme.

image

There are very limited themes available for ag-grid. For a light theme, thi is what I could find. Please let me know if you think the colors are good.

ebugden:
> **muddana-satish:** There are very limited themes available for ag-grid.

Do you mean you are picking from a predefined set of themes and it's not possible to manually set the colours? 
Which element is ag-grid?

---

> **muddana-satish:** Please let me know if you think the colors are good.

My first reflex would be to use shades of neutral grey for the chart background, gridlines, and labels instead of shades of gold.
* The standard light theme looks composed entirely of neutral greys (no colour bias) with a couple pops of blue. (Excluding the random warm grey used for the editor's main horizontal scroll bar.) 
* I think the shades of gold used for the chart background, gridlines, and labels feel out of place in this cool toned background colour palette.

image

packages/react-components/src/components/trace-context-component.tsx
```diff
@@ -72,6 +74,8 @@ export class TraceContextComponent extends React.Component<TraceContextProps, Tr
             traceRange = new TimeRange(experiment.start - this.props.experiment.start, experiment.end - this.props.experiment.start, this.props.experiment.start);
             viewRange = new TimeRange(experiment.start - this.props.experiment.start, experiment.end - this.props.experiment.start, this.props.experiment.start);
         }
+        ThemeService.get().onThemeChange(() => this.updateBackgroundTheme());
+        const currentThemeId = ThemeService.get().getCurrentTheme().id;
```
bhufmann, line 78: If you use type instead of id then you will support all light and dark themes and not just the Theia light theme.

packages/react-components/src/components/trace-context-component.tsx
```diff
@@ -114,6 +119,23 @@ export class TraceContextComponent extends React.Component<TraceContextProps, Tr
         this.unitController.viewRange = { start: 0, end: this.state.experiment.end - this.state.timeOffset };
     }
 
+    private updateBackgroundTheme() {
+        const currentThemeId = ThemeService.get().getCurrentTheme().id;
```
bhufmann, line 123: use type here too

packages/react-components/src/components/trace-context-component.tsx
```diff
@@ -20,6 +20,7 @@ import { XYOutputComponent } from './xy-output-component';
 import { NullOutputComponent } from './null-output-component';
 import { AbstractOutputProps } from './abstract-output-component';
 import * as Messages from '@trace-viewer/base/lib/message-manager';
+import { ThemeService } from '@theia/core/lib/browser/theming';
```
bhufmann, line 23: "react-component" and "base" are supposed to be independent from Theia (and VsCode). So, the code to detect the theme and update of theme needs to be under viewer-prototype which is Theia specific. From there you need to send a signal that the trace-context-compenent is receiving and updating according to the change. 

See for the signals in base/signal-manager.ts and how they are used.

bhufmann (changes requested):
Thanks for the contributions. Nice improvement.

Some initial comments.

I also noticed that when changing the theme a trace with graphs are open, then the background colors are not updated

marcdumais-work:
I think this contribution will need to be done in a different way, in the trace viewer `vscode extension`, most likely using theme-related `vscode APIs`. Maybe @vince-fugnitto can confirm. I guess if the change is small/simple and approved by reviewers, it can still be worth it, for nicer-looking demos.

bhufmann:
> I think this contribution will need to be done in a different way, in the trace viewer `vscode extension`, most likely using theme-related `vscode APIs`. Maybe @vince-fugnitto can confirm. I guess if the change is small/simple and approved by reviewers, it can still be worth it, for nicer-looking demos.

Yes, the vscode-extension will have to take care of the monitoring the theme an propagating the information to the react-component using the vscode APIs. the vscode-extension is not ready and right now we'll use the theia extension.

packages/react-components/src/components/trace-context-component.tsx
muddana-satish: Thank you for the suggestion. I moved the logic to trace-viewer.

muddana-satish:
> Thanks for the contributions. Nice improvement.
> 
> Some initial comments.

@bhufmann Thank you for the suggestions. I have updated the changes based on your suggestion. Please review it.

> I also noticed that when changing the theme a trace with graphs are open, then the background colors are not updated

I want to fix this as part of a new PR. I will log an issue. Do you think this feature should be part of MVP? Please let me know.

packages/react-components/src/components/trace-context-component.tsx
```diff
@@ -84,11 +87,12 @@ export class TraceContextComponent extends React.Component<TraceContextProps, Tr
                 chartWidth: this.DEFAULT_CHART_WIDTH,
                 height: this.DEFAULT_COMPONENT_HEIGHT,
                 rowHeight: this.DEFAULT_COMPONENT_ROWHEIGHT,
-                naviBackgroundColor: 0x3f3f3f,
-                chartBackgroundColor: 0x3f3f3f,
+                naviBackgroundColor: this.props.backgroundTheme === 'light' ? 0xf9f6e8 : 0x3f3f3f,
+                chartBackgroundColor: this.props.backgroundTheme === 'light' ?0xf9f6e8 : 0x3f3f3f,
```
bhufmann, line 91: space after ?

packages/react-components/src/components/trace-context-component.tsx
```diff
@@ -106,6 +110,23 @@ export class TraceContextComponent extends React.Component<TraceContextProps, Tr
...
+
+    public updateBackgroundTheme(theme: string) {
+        this.setState({
+            style: {
+                width: this.DEFAULT_COMPONENT_WIDTH,
+                chartWidth: this.DEFAULT_CHART_WIDTH,
+                height: this.DEFAULT_COMPONENT_HEIGHT,
+                rowHeight: this.DEFAULT_COMPONENT_ROWHEIGHT,
+                naviBackgroundColor: theme === 'light' ? 0xf9f6e8 : 0x3f3f3f,
+                chartBackgroundColor: theme === 'light' ?0xf9f6e8 : 0x3f3f3f,
```
bhufmann, line 124: space after ?

bhufmann (changes requested):
> 
> I want to fix this as part of a new PR. I will log an issue. Do you think this feature should be part of MVP? Please let me know.

I think it is acceptable to fix it after the MVP. Users don't switch the theme that often. Closing and opening the trace will fix it as a workaround. 

About the colors chosen, @ebugden do you have any suggestion on which ones we should use for the MVP? Or are you ok with the proposal?

ebugden:
@muddana-satish Clarification for colours: Please use shades of neutral grey for the chart background, gridlines, and labels instead of shades of gold. Replace all gold or tan in the UI.

image

For the MVP I think the table theme colours can be used as replacements. Suggestions: 
* **Chart backgrounds:** Use the darker list item background colour:
image

* **Chart gridlines and axis labels:** Use the table outline colour (darken as needed to make axis labels comfortably legible):
image

muddana-satish:
@ebugden Thank you for the suggestions.

Attached the screenshot with new colors. Let me know what you think about this.

image

packages/react-components/src/components/trace-context-component.tsx
muddana-satish: Updated. Thanks

ebugden:
@muddana-satish Thanks for the update! It's cleaner and more consistent. 

I have a couple requests since the current chart background colour is darker than what is used in the Balham table theme.  A darker background makes the chart data less legible because it decreases the colour contrast with the data displayed.

Please make the following changes:
* **Chart background:** Start by making the chart backgrounds the exact odd row background colour specified in the Balham table theme file. The background should only be as dark as it needs to be for users to tell that it's a distinct area from the rest of the panel background.
* **Chart gridlines and labels:** Start by making the chart gridlines and and labels the exact outline colour in the Balham table theme. If this grey is too light and it's difficult to read the labels, increase the darkness of the colour until you find it's comfortable to read. 

---

**Useful bits**
Please double check, but I think this is the official [Balham theme file](https://github.com/ag-grid/ag-grid/blob/master/dist/styles/ag-theme-balham/vars/_ag-theme-balham-vars.scss) so you could refer there for the colours:
image

(Current background colour:) 
image

For the background, start by trying this colour:
image

If the chart area is not visibly distinct from the main panel background, test gradually darker shades. Stop when it's visible that the chart is a different area. Btw [colorizer](http://colorizer.org/) is great for making slight adjustments to colours.
image
image
image

packages/react-components/src/components/trace-context-component.tsx
```diff
@@ -106,6 +110,23 @@ export class TraceContextComponent extends React.Component<TraceContextProps, Tr
...
+        this.setState({
+            style: {
+                width: this.DEFAULT_COMPONENT_WIDTH,
+                chartWidth: this.DEFAULT_CHART_WIDTH,
+                height: this.DEFAULT_COMPONENT_HEIGHT,
+                rowHeight: this.DEFAULT_COMPONENT_ROWHEIGHT,
+                naviBackgroundColor: theme === 'light' ? 0xf5f7f7 : 0x3f3f3f,
+                chartBackgroundColor: theme === 'light' ? 0xf5f7f7 : 0x3f3f3f,
+                cursorColor: 0x259fd8,
+                lineColor: theme === 'light' ? 0x8b8c8c : 0xbbbbbb
```
tahini, line 126: It could be nice to extract these style methods to somewhere central, so there would be a single goto place to implement new themes or change values for an existing theme.

tahini (approved):
Very nice improvement, my comment is a suggestion for later.

muddana-satish:
@ebugden I have modified the colors again based on your suggestions. Looks good for me. When I use the same background color as events-table, I couldn't distinguish between the widget and the panel background. 

This suggested color looked better than the rest.
> image

For the grid lines and labels, I used the same color as the column headers in events table. Please let me know what you think.

image

packages/react-components/src/components/trace-context-component.tsx
muddana-satish: Agree. I would like to fix it as part of another PR. Thank you for the suggestion :)

ebugden:
@muddana-satish I'm happy with this!

bhufmann (approved):

bhufmann:
Looks very nice in light theme as well. Thanks for the contribution!

Issue #53, opened 2020-05-25
Use the Theia backend to automatically start the trace server


tahini:
On basic installations, there may not be a trace server running separately from theia. Admins need to make sure such a server exists and is started. The Theia default installation should include fetching and starting the web server.

The trace server should also be configurable, so if there's a remote server running, it should all point to it.

ebugden:
@bhufmann @ankusht-work @mirsky-work Could we group our thoughts on here? I'm thinking about how to automatically start/manage the trace server and looked at [trace-server-status.ts](https://github.com/theia-ide/theia-trace-extension/blob/master/viewer-prototype/src/browser/trace-server-status.ts) for inspiration.

My first reflex was to reimplement the [handleOffline( )](https://github.com/theia-ide/theia-trace-extension/blob/a57e90032868818e31091996584385e5e0e49d09/viewer-prototype/src/browser/trace-server-status.ts#L130) function in the TraceServerConnectionStatusService to start/restart the server.

The new handleOffline( ) could be loosely inspired by the [openTrace( )](https://github.com/theia-ide/theia-trace-extension/blob/a57e90032868818e31091996584385e5e0e49d09/packages/base/src/trace-manager.ts#L73) function in [trace-manager.ts](https://github.com/theia-ide/theia-trace-extension/blob/a57e90032868818e31091996584385e5e0e49d09/packages/base/src/trace-manager.ts) and try restarting the server a limited number of times with a counter. 

Eventually [schedulePing( )](https://github.com/theia-ide/theia-trace-extension/blob/a57e90032868818e31091996584385e5e0e49d09/viewer-prototype/src/browser/trace-server-status.ts#L38) could be also changed to use @tahini's new server ping instead of using fetchExperiments( ) (at line 44) to check if the server is alive.

ankusht-work:
For starting the server automatically for the user I have listed down all the scenarios that I could think of. @tahini @bhufmann @ebugden @mirsky-work Please let me know your thoughts. 

- Remove the start and stop button from the status bar. Add a text field in the status bar indicating if the server is offline or online. 

- Upon clicking 'OpenTrace' from the trace viewer tab or 'Open with -> Trace Viewer' context menu", check if the server is running. If the server is not started then read the default values of path/port from the preferences and spawn the server.

- If the user tries to modify the value in the path or port preferences, we first kill the server that is running and spawn the server with the new values. Add a dialog message to notify the user about the same.

[...]

PR #198, opened 2020-12-07
Add preference settings for trace configuration

### LINKED ISSUE #184: Add Trace-extension configuration settings
ToDo:
1) Add new preferences settings that will take two parameters a) path to the executable b) port on which user would like to spawn the server.
2) Implement an event-listener that will monitor the changes in trace-server preferences.
3) Spawn the server at the start of the application and/or whenever the preferences settings are modified?---Need more investigation on this. 

The draft of the preferences UI can be seen in the image attached. Any feedback is much appreciated.
Config

### LINKED ISSUE #191: Add buttons to explicitly start and stop the trace server
We can consider two options: 

1) Placing the set of buttons under the Trace explorer widget as seen in the image.
Option 1
     
Pros: 
- We are keeping all the logic within the trace plugin so if the plugin is disabled/removed, we don't have to explicitly 
remove buttons specific to this plugin.

Cons : 
- Might clutter the space since we have an open Trace button as well. 
- User can accidentally click the stop button while trying to import (maybe?)
 
     

2) Add buttons to the existing status bar. Once the server is running, the status bar changes to blue color. 
Status_bar

Pros:
- We already provide notification that the Trace server is offline. 
- Providing the ability to start the server next to the warning is a good user experience I believe.

Cons:
- Might need to explicitly remove buttons from the status bar if the extension is disabled (maybe). Not too sure about this honestly.

ankusht-work:
Set the default values for the trace server preferences
Read the trace server preference values and spawn the server from the path specified in the
preferences

fixes #191
fixes #184

Signed-off-by: Ankush Tyagi ankush.tyagi@ericsson.com

### DIFF
--- a/viewer-prototype/src/browser/trace-server-preference.ts
+++ b/viewer-prototype/src/browser/trace-server-preference.ts
@@ -1,0 +1,30 @@
+import { PreferenceSchema, PreferenceProxy, PreferenceScope } from '@theia/core/lib/browser';
+
+export const TRACE_PATH = 'trace-viewer.path';
+export const TRACE_PORT = 'trace-viewer.port';
+
+export const ServerSchema: PreferenceSchema = {
+    type: 'object',
+    properties: {
+        [TRACE_PATH]: {
+            'type': 'string',
+            'default': '',
+            'description': 'The path to trace-server executable, e.g.: /usr/bin/tracecompass-server',
+        },
+        [TRACE_PORT]: {
+            'type': 'number',
+            'default': '',
+            'description': 'Specify the port on which you want to execute the server.',
+        }
+
+    },
+    scope: PreferenceScope.Folder,
+};
+
+interface TracePreferenceContribution {
+    [TRACE_PATH]: string;
+    [TRACE_PORT]: number;
+}
+
+export const TracePreferences = Symbol('TracePreferences');
+export type TracePreferences = PreferenceProxy<TracePreferenceContribution>;
--- a/viewer-prototype/src/browser/trace-server-status.ts
+++ b/viewer-prototype/src/browser/trace-server-status.ts
@@ -12,6 +12,9 @@ import { Disposable, DisposableCollection } from '@theia/core/lib//common';
 import { ConnectionStatusService, ConnectionStatus, AbstractConnectionStatusService } from '@theia/core/lib/browser/connection-status-service';
 import { TspClient } from 'tsp-typescript-client/lib/protocol/tsp-client';
 import { TspClientProvider } from './tsp-client-provider';
+import { TraceServerConfigService } from '../common/trace-server-config';
+import { PreferenceService } from '@theia/core/lib/browser';
+import { TRACE_PATH, TRACE_PORT } from './trace-server-preference';
 
 @injectable()
 export class TraceServerConnectionStatusService extends AbstractConnectionStatusService {
@@ -52,6 +55,29 @@ export class TraceServerConnectionStatusService extends AbstractConnectionStatus
 @injectable()
 export class TraceServerConnectionStatusContribution extends DefaultFrontendApplicationContribution {
 
+    @inject(PreferenceService) protected readonly preferenceService: PreferenceService;
+
+    private path: string | undefined;
+    private port: number | undefined;
+
+    @postConstruct()
+    async init(): Promise<void> {
+
+        this.path = this.preferenceService.get(TRACE_PATH);
+        this.port = this.preferenceService.get(TRACE_PORT);
+
+        this.preferenceService.onPreferenceChanged(event => {
+            if (event.preferenceName === TRACE_PORT) {
+                this.port = event.newValue;
+            }
+            if (event.preferenceName === TRACE_PATH) {
+                this.path = event.newValue;
+            }
+        });
+    }
+
+    @inject(TraceServerConfigService) protected readonly traceServerConfigService: TraceServerConfigService;
+
     protected readonly toDisposeOnOnline = new DisposableCollection();
 
     constructor(
@@ -61,6 +87,9 @@ export class TraceServerConnectionStatusContribution extends DefaultFrontendAppl
     ) {
         super();
         this.connectionStatusService.onStatusChange(state => this.onStateChange(state));
+        if (this.connectionStatusService.currentStatus === ConnectionStatus.ONLINE) {
+            this.handleOnline();
+        }
     }
 
     protected onStateChange(state: ConnectionStatus): void {
@@ -80,15 +109,34 @@ export class TraceServerConnectionStatusContribution extends DefaultFrontendAppl
 
     protected handleOnline(): void {
         this.toDisposeOnOnline.dispose();
+        this.statusBar.setElement(this.statusbarId, {
+            alignment: StatusBarAlignment.LEFT,
+            text: '$(fas fa-stop) Stop trace server',
+            tooltip: 'Click here to stop the trace server',
+            priority: 5003,
+            onclick: this.stopServer.bind(this)
+        });
+
+    }
+
+    private async startServer() {
+        await this.traceServerConfigService.startTraceServer(this.path, this.port);
+    }
+
+    private async stopServer() {
+        await this.traceServerConfigService.stopTraceServer(this.port);
     }
 
     protected handleOffline(): void {
+        this.toDisposeOnOnline.dispose();
         this.statusBar.setElement(this.statusbarId, {
             alignment: StatusBarAlignment.LEFT,
-            text: 'Trace Server Offline',
-            tooltip: 'Cannot connect to trace server.',
-            priority: 5000
+            text: '$(fas fa-play) Start trace server',
+            tooltip: 'Click here to start the trace server',
+            priority: 5001,
+            onclick: this.startServer.bind(this)
         });
+
         this.toDisposeOnOnline.push(Disposable.create(() => this.statusBar.removeElement(this.statusbarId)));
         document.body.classList.add('traceserver-mod-offline');
         this.toDisposeOnOnline.push(Disposable.create(() => document.body.classList.remove('traceserver-mod-offline')));
--- a/viewer-prototype/src/node/trace-server-service.ts
+++ b/viewer-prototype/src/node/trace-server-service.ts
@@ -1,0 +1,14 @@
+import { spawn, exec } from 'child_process';
+import { injectable } from 'inversify';
+import { TraceServerConfigService } from '../common/trace-server-config';
+@injectable()
+export class TraceServerServiceImpl implements TraceServerConfigService {
+
+    async startTraceServer(path: string | undefined, port: number | undefined): Promise<void> {
+        spawn(`${path}`, ['-vmargs', `-Dtraceserver.port=${port}`]);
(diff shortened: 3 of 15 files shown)


ankusht-work:
Start-Stop-Server

Above Gif illustrates the behavior of the start/stop server. This might not be the optimal solution and should be considered as an incremental step (especially for MVP). This will be revisited once we finalize the design and the UX of the Trace configuration for both the remote and the local use case.

viewer-prototype/src/browser/trace-server-preference.ts
```diff
@@ -0,0 +1,30 @@
...
+export const TRACE_PATH = 'trace-viewer.path';
+export const TRACE_PORT = 'trace-viewer.port';
+
+export const ServerSchema: PreferenceSchema = {
+    type: 'object',
+    properties: {
+        [TRACE_PATH]: {
+            'type': 'string',
+            'default': '',
+            'description': 'The path to trace-server executable, e.g.: /usr/bin/',
```
tahini, line 12: e.g. should be /usr/bin/tracecompass-server since it must include the executable name

viewer-prototype/src/browser/trace-server-status.ts
```diff
@@ -80,15 +100,34 @@ export class TraceServerConnectionStatusContribution extends DefaultFrontendAppl
 
     protected handleOnline(): void {
         this.toDisposeOnOnline.dispose();
+        this.statusBar.setElement(this.statusbarId, {
+            alignment: StatusBarAlignment.LEFT,
+            text: '$(fas fa-stop) Stop trace server',
```
tahini, line 105: This is really great, except that it stopped my entire browser!! (or it crashed I'm not sure)

viewer-prototype/src/browser/trace-server-preference.ts
tahini, line 12: e.g. should be /usr/bin/tracecompass-server as it must include the executable

viewer-prototype/src/browser/trace-server-preference.ts
```diff
@@ -0,0 +1,30 @@
...
+        },
+        [TRACE_PORT]: {
+            'type': 'number',
+            'default': '',
+            'description': 'Specify the port on which you want to execute the server.',
```
tahini, line 17: "you want", is it really up to me? Changing the port will have the server run on that port? It seems that if I put port 8082, everything is still on port 8080.

viewer-prototype/src/node/trace-server-service.ts
```diff
@@ -0,0 +1,14 @@
...
+import { TraceServerConfigService } from '../common/trace-server-config';
+@injectable()
+export class TraceServerServiceImpl implements TraceServerConfigService {
+
+    async startTraceServer(path: string | undefined, port: number | undefined): Promise<void> {
+        spawn(`..${path}`, ['-vmargs', `-Dtraceserver.port=${port}`]);
+    }
+
+    async stopTraceServer(port: number | undefined): Promise<void> {
+        exec(`kill -9 $(lsof -t -i:${port})`);  // FIXME: Better approach to kill the server at a given port
```
tahini, line 12: I guess you'll have to find a better approach now... ;-)
```
$ lsof -t -i:8080
32190
32627
$ ps aux | grep 32190
gbastien 32190 26.1  3.0 9837920 483804 tty2   Sl+  14:13   1:13 /usr/lib/firefox/firefox -new-window
$ ps aux | grep 32627
<The real trace server process>
```

tahini (changes requested):
This is a really great feature! ... except that stopping the server kills the browser...

What I'd like to test once it works well is the behavior when you specify a trace server URL, just to see if it does not end up in unintuitive behaviors. That's a 100000x improvement on the local use case, I just want to make sure it does not break too much the remote one.

ebugden:
**Why don't we just say "The trace server is running" after the server is started and not allow the user to restart manually?**

image

This is safe because according to @mirsky-work's [comment](https://github.com/theia-ide/theia-trace-extension/issues/191#issuecomment-738302796) on #191 the server stops automatically then the tool is closed.

Motivation: Manually restarting the server is an uncommon use case so this alternative streamlines the UX. More importantly, this alternative avoids spending time fixing bugs caused by restarting the server on the fly:
* Stopping the server crashed @tahini's browser
* On gitpod, open traces don't reconnect properly to the server when it is restarted with the defaults, but the traces are still listed as open. This is confusing/messy for the user and gives the tool an unfinished appearance.

---

More detail about this PR being an intermediate solution: Based on @bhufmann's proposal under "Add buttons to explicitly start and stop the trace server" #191 ([comment](https://github.com/theia-ide/theia-trace-extension/issues/191#issuecomment-738829709)) the end goal for the local case is to start the server automatically.

viewer-prototype/src/browser/trace-server-preference.ts
ankusht-work: Updated.

viewer-prototype/src/browser/trace-server-status.ts
ankusht-work: Fixed

viewer-prototype/src/node/trace-server-service.ts
ankusht-work: Interesting it didn't kill the browser for me (I was using chrome) and I was able to start/stop multiple times on the same tab. However, I think what you suggested (-sTCP: LISTEN) makes more sense and I have updated it.

viewer-prototype/src/browser/trace-server-preference.ts
ankusht-work: Fixed. You can specify the port on which you want to execute the server and the listener will update the TraceServerUrl's port accordingly.

viewer-prototype/src/node/trace-server-service.ts
```diff
@@ -0,0 +1,14 @@
+import { spawn, exec } from 'child_process';
+import { injectable } from 'inversify';
+import { TraceServerConfigService } from '../common/trace-server-config';
+@injectable()
+export class TraceServerServiceImpl implements TraceServerConfigService {
+
+    async startTraceServer(path: string | undefined, port: number | undefined): Promise<void> {
+        spawn(`..${path}`, ['-vmargs', `-Dtraceserver.port=${port}`]);
```
bhufmann, line 8: Are the ".." before the path intentional? When I put the path in the the settings /usr/bin/tracecompass-server then it tries to execute "../usr/bin/tracecompass-server" which fails and cause weird behaviour when opening a trace (i.e. it hangs the tsp calls).

viewer-prototype/src/node/trace-server-service.ts
```diff
@@ -0,0 +1,14 @@
...
+import { TraceServerConfigService } from '../common/trace-server-config';
+@injectable()
+export class TraceServerServiceImpl implements TraceServerConfigService {
+
+    async startTraceServer(path: string | undefined, port: number | undefined): Promise<void> {
+        spawn(`${path}`, ['-vmargs', `-Dtraceserver.port=${port}`]);
+    }
+
+    async stopTraceServer(port: number | undefined): Promise<void> {
+        exec(`kill -9 $(lsof -t -i:${port}) -sTCP:LISTEN`);  // FIXME: Better approach to kill the server at a given port
```
tahini, line 12: The -sTCP:LISTEN goes in the parenthesis ;-) My browser still dies.

tahini (changes requested):

viewer-prototype/src/node/trace-server-service.ts
ankusht-work: Yes :) Will make that change. It doesn't kill the browser for me. I am using Firefox version 83 and chrome version 75. Could you please share what browser/version you are using?

tahini: firefox 83 also, with ubuntu beaver. Maybe it's because I usually happen to have a tab opened to some app at port localhost:8080?

Also, does your patch work on Windows? Not that I care ;-) But just for info. What will be Windows users' experience with the trace server?

ankusht-work: > firefox 83 also, with ubuntu beaver. Maybe it's because I usually happen to have a tab opened to some app at port localhost:8080?
> Yeah could be. Maybe change the URL to something else instead of localhost and test it again? This will also verify if my changes are not breaking the remote case. :)

@Bernd: Are you also facing the same issue?


> Also, does your patch work on Windows? Not that I care ;-) But just for info. What will be Windows users' experience with the trace server?
The last time I played with the window's trace server executable (many months ago) I was running into some path issues. I have not yet played with the latest windows server. I will try and let you know.

viewer-prototype/src/node/trace-server-service.ts
ankusht-work: Fixed.

viewer-prototype/src/node/trace-server-service.ts
ankusht-work: Fixed the parenthesis.

bhufmann:
It works fine in our production environment. On my development machine (Ubuntu 18.04 and Chrome Version 87.0.4280.141) there seems to be a delay in the connection timeout when sending a fetch request to determine if the server is up at startup. Once the connection timeout is received it works fine.

bhufmann (approved):
Thanks for the contribution. It's great UX improvement for the user and an important step towards automatically start/stop of the trace server for the user in the local use case.

bhufmann:
@tahini any more comments? Otherwise we'd like to merge it.

tahini (approved):
It works well and is a great UX improvement, if the trace server is available on the system.

Here are a few issues I noticed, that can be fixed separately. I can open bugs for it this patch is merged:

1- Starting/Stopping is an operation that takes time, so the notification bar does not change instantly. The user is not aware that it did anything and may try clicking again. Some progress of the operation would be nice

2- If anything happens, we have no feedback (path does not exist, any other error), so the user doesn't know how to react if the server is not started/stopped after a while

3- If is possible to change the port of the server while it is running. It is then impossible to start it, we don't know why, and need to go back to previous port to stop the server and start it again. It shouldn't be possible to modify the port while the server is running, or else the user should be asked if they want to restart the server on this new port.

bhufmann:
> It works well and is a great UX improvement, if the trace server is available on the system.
> 
> Here are a few issues I noticed, that can be fixed separately. I can open bugs for it this patch is merged:
> 
> 1- Starting/Stopping is an operation that takes time, so the notification bar does not change instantly. The user is not aware that it did anything and may try clicking again. Some progress of the operation would be nice
> 
> 2- If anything happens, we have no feedback (path does not exist, any other error), so the user doesn't know how to react if the server is not started/stopped after a while
> 
> 3- If is possible to change the port of the server while it is running. It is then impossible to start it, we don't know why, and need to go back to previous port to stop the server and start it again. It shouldn't be possible to modify the port while the server is running, or else the user should be asked if they want to restart the server on this new port.

Thanks. I agree to handle these items separately.

ebugden:
I agree with @tahini's points. My intuition would be to prioritize managing the server automatically #53 (starting, restarting). I think this would improve/resolve a lot of these issues.

PR #199, opened 2020-12-08
Support fetching annotations and populating them in TimelineChart model

PatrickTasse:

### DIFF
--- a/packages/react-components/src/components/data-providers/tsp-data-provider.ts
+++ b/packages/react-components/src/components/data-providers/tsp-data-provider.ts
@@ -45,11 +45,14 @@ export class TspDataProvider {
             const end = viewRange.end + this.timeGraphEntries[0].start;
             statesParameters = QueryHelper.selectionTimeQuery(QueryHelper.splitRangeIntoEqualParts(Math.trunc(start), Math.trunc(end), resolution), ids);
         }
-        const stateResponse = (await this.client.fetchTimeGraphStates(this.traceUUID,
-            this.outputId, statesParameters)).getModel();
-
-        this.timeGraphRows = stateResponse.model.rows;
-        this.timeGraphRowsOrdering(ids);
+        const tspClientResponse = await this.client.fetchTimeGraphStates(this.traceUUID, this.outputId, statesParameters);
+        const stateResponse = tspClientResponse.getModel();
+        if (tspClientResponse.isOk() && stateResponse) {
+            this.timeGraphRows = stateResponse.model.rows;
+            this.timeGraphRowsOrdering(ids);
+        } else {
+            this.timeGraphRows = [];
+        }
 
         // the start time which is normalized to logical 0 in timeline chart.
         const chartStart = this.timeGraphEntries[0].start;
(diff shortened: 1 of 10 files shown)


packages/react-components/src/components/timegraph-output-component.tsx
```diff
@@ -353,9 +353,38 @@ export class TimegraphOutputComponent extends AbstractTreeOutputComponent<Timegr
...
+                if (modelStyle) {
+                    const currentStyle = Object.assign({}, modelStyle.values, elementStyle.values);
+                    if (currentStyle) {
+                        let color = 0;
+                        if (currentStyle['color']) {
+                            color = this.hexStringToNumber(currentStyle['color']);
+                        }
+                        let symbolSize = this.props.style.rowHeight * 0.8 / 2;
+                        if (currentStyle['height']) {
+                            symbolSize = currentStyle['height'] * this.props.style.rowHeight;
```
PatrickTasse, line 377: This should be ' * symbolSize'

tahini:
If this patch requires changes to other libraries, could you point to them here please?

bhufmann (changes requested):
you need to update the yarn.lock file to pick up the latest tsp-typescript-client and timeline-chart.

packages/react-components/src/components/data-providers/tsp-data-provider.ts
```diff
@@ -61,12 +62,32 @@ export class TspDataProvider {
                 rows.push(this.getRowModel(row, chartStart, rowId, entry));
             }
         });
-
+        const annotations: Map<number, TimelineChart.TimeGraphAnnotation[]> = new Map();
+        const annotationsResponse = (await this.client.fetchAnnotations(this.traceUUID,
```
tahini, line 66: I tried it with Matthew's patch https://git.eclipse.org/r/c/tracecompass/org.eclipse.tracecompass/+/173517 for ust annotations in kernel trace and for some reason, the server returned a 500 and i'm not sure where the bug is exactly, but the view returns nothing. You should probably the response status, or at least surrount the annotation block with try-catch to avoid an empty view when annotations return errors

tahini (changes requested):
Tried it but got a server-side exception. The code looks good, but I'll wait to be able to test it to fully appreciate it ;-)

packages/react-components/src/components/data-providers/tsp-data-provider.ts
PatrickTasse: Have you figured out where the 500 Internal Server Error comes from? It doesn't look like one of the possible failures in the trace-server.

We plan to address robustness and error handling in a later patch, if this is OK.

tahini: > Have you figured out where the 500 Internal Server Error comes from? It doesn't look like one of the possible failures in the trace-server.

Not, but I did not investigate!

> We plan to address robustness and error handling in a later patch, if this is OK.

Sure, but you could just put a try catch around those lines to avoid empty views for now. Real error handling can wait.

MatthewKhouzam: I got the patch to work... after some finagling... lots of finagling... my take:

We need to call this.client.fetchCategories() first. Even if we don't use the result, at least to get the dp into the right state.

MatthewKhouzam (changes requested):
As mentionned in the comments, please call fetchAnnotationCategories upon openning a view, even if it's not used.

packages/react-components/src/components/data-providers/tsp-data-provider.ts
MatthewKhouzam: Also: 
annotations

MatthewKhouzam (approved):

MatthewKhouzam (approved):
I am OK with these changes.

tahini (approved):
I couldn't test it with actual annotations since the trace server is broken here. But code-wise, it looks good and everything still works. Thanks!

Issue #111, opened 2020-08-25
Selecting a row in the events table should select the corresponding state in the time graph


mirsky-work:
- The time graph's X-axis position should be scrolled to the timestamp of the selected row.
- The time graph's Y-axis should be scrolled to the correct row that contains the corresponding state (would be nice to highlight it).
- The zoom-level of the graph shouldn't be affected.

mirsky-work:
Fixed by #207

muddana-satish:
@mirsky-work It's not a fully implemented feature with #207 . I think this should still be open.

PR #207, opened 2020-12-23
Sync event timestamp with timegraph

muddana-satish:

depends on https://github.com/theia-ide/timeline-chart/pull/85

### DIFF
--- a/packages/base/src/signal-manager.ts
+++ b/packages/base/src/signal-manager.ts
@@ -4,27 +4,33 @@ export declare interface SignalManager {
 
     fireTooltipSignal(tooltip: { [key: string]: string }): void;
     fireThemeChangedSignal(theme: string): void;
+    fireSelectionChangedSignal(payload: { [key: string]: string }): void;
 
 }
 
 export const Signals = {
-    TRACE_OPENED : 'trace opened',
-    TRACE_CLOSED : 'trace closed',
+    TRACE_OPENED: 'trace opened',
+    TRACE_CLOSED: 'trace closed',
     EXPERIMENT_OPENED: 'experiment opened',
     EXPERIMENT_CLOSED: 'experiment closed',
     EXPERIMENT_SELECTED: 'experiment selected',
     TOOLTIP_UPDATED: 'tooltip updated',
-    THEME_CHANGED: 'theme changed'
+    THEME_CHANGED: 'theme changed',
+    SELECTION_CHANGED: 'selection changed'
 };
 
 export class SignalManager extends EventEmitter implements SignalManager {
 
     fireTooltipSignal(tooltip: { [key: string]: string; }): void {
-        this.emit(Signals.TOOLTIP_UPDATED, {tooltip});
+        this.emit(Signals.TOOLTIP_UPDATED, { tooltip });
     }
 
     fireThemeChangedSignal(theme: string) {
         this.emit(Signals.THEME_CHANGED, theme);
+    }
+
+    fireSelectionChangedSignal(payload: { [key: string]: string; }): void {
+        this.emit(Signals.SELECTION_CHANGED, { payload });
     }
 
 }
--- a/packages/react-components/src/components/table-output-component.tsx
+++ b/packages/react-components/src/components/table-output-component.tsx
@@ -2,9 +2,10 @@
 import { AbstractOutputComponent, AbstractOutputProps, AbstractOutputState } from './abstract-output-component';
 import * as React from 'react';
 import { AgGridReact } from 'ag-grid-react';
-import { ColDef, IDatasource, GridReadyEvent } from 'ag-grid-community';
+import { ColDef, IDatasource, GridReadyEvent, CellClickedEvent } from 'ag-grid-community';
 import { QueryHelper } from 'tsp-typescript-client/lib/models/query/query-helper';
 import { cloneDeep } from 'lodash';
+import { signalManager } from '@trace-viewer/base/lib/signal-manager';
 
 type TableOuputState = AbstractOutputState & {
     tableColumns: ColDef[];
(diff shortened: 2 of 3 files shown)


bhufmann:
@muddana-satish I currently reviewing this patch. I'll get back to you soon.

packages/base/src/signal-manager.ts
```diff
@@ -3,6 +3,7 @@ import { EventEmitter } from 'events';
 export declare interface SignalManager {
 
     fireTooltipSignal(tooltip: { [key: string]: string }): void;
+    fireEventClickSignal(payload: {[timeStamp: string]: number }): void;
```
bhufmann, line 6: "timestamp" should be a different name. Timestamp is one of the keys. Later on we might have multiple once.
I wonder if we should pass strings instead of numbers and the signal receivers have to convert them if necessary.
Name could be fireSelectionChangedSignal(), selection changed (see comment below on signal name)

packages/base/src/signal-manager.ts
```diff
@@ -12,7 +13,8 @@ export const Signals = {
     EXPERIMENT_OPENED: 'experiment opened',
     EXPERIMENT_CLOSED: 'experiment closed',
     EXPERIMENT_SELECTED: 'experiment selected',
-    TOOLTIP_UPDATED: 'tooltip updated'
+    TOOLTIP_UPDATED: 'tooltip updated',
+    EVENT_CLICKED: 'Event clicked'
```
bhufmann, line 17: Event -> event: should lower case. I wonder if we will re-use this signal for other selections changes (e.g. chart to table). So, maybe we call it "selection changed".

packages/react-components/src/components/table-output-component.tsx
```diff
@@ -62,11 +63,25 @@ export class TableOutputComponent extends AbstractOutputComponent<TableOutputPro
...
+                onCellClicked={this.onEventClick}
+                rowSelection='single'
             >
             </AgGridReact>
         </div>;
     }
 
+    private onEventClick(event: CellClickedEvent) {
+        const columns = event.columnApi.getAllColumns();
+        const timeStampHeader = columns.find(column => column.getColDef().headerName === "Timestamp (ns)");
```
bhufmann, line 75: The column header string is acutally "Timestamp ns" in the trace compass server.

packages/react-components/src/components/timegraph-output-component.tsx
```diff
@@ -86,6 +87,7 @@ export class TimegraphOutputComponent extends AbstractTreeOutputComponent<Timegr
             }
             this.onElementSelected(this.selectedElement);
         });
+        signalManager().on(Signals.EVENT_CLICKED, ({ timeStamp }) => this.eventClicked(timeStamp));
```
bhufmann, line 90: use payload instead of timestamp

packages/react-components/src/components/timegraph-output-component.tsx
```diff
@@ -141,6 +143,19 @@ export class TimegraphOutputComponent extends AbstractTreeOutputComponent<Timegr
         this.setState({ collapsedNodes: newList });
     }
 
+    private eventClicked(timeStampObj: { [key: string]: number }) {
```
bhufmann, line 146: use payload instead of timestamp.

packages/react-components/src/components/timegraph-output-component.tsx
```diff
@@ -141,6 +143,19 @@ export class TimegraphOutputComponent extends AbstractTreeOutputComponent<Timegr
...
+        const offset = this.props.viewRange.getOffset();
+        const rawTimeStamp = timeStampObj['timeStamp'];
+        const newViewRangeStart = rawTimeStamp - (offset ? offset : 0);
+        this.props.unitController.selectionRange = {
+            start: newViewRangeStart,
+            end: newViewRangeStart
+        };
+        this.chartLayer.setEventClickedFlag(true);
```
bhufmann, line 154: I would prefer to separate the timestamp sync and row sync in 2 different pull requests (in theia-trace-extesion and timeline-chart).

bhufmann (changes requested):

packages/react-components/src/components/table-output-component.tsx
tahini: If we need this, then it's a sign there's a problem with the TSP, this is the kind of string that should never be hard-coded. Time to work on https://github.com/theia-ide/trace-server-protocol/issues/35?

mirsky-work:
Should be linked to #111

muddana-satish:
@bhufmann Modified as per your suggestions. Now the selection cursor moves based on the timestamp from the event row.

packages/base/src/signal-manager.ts
```diff
@@ -3,6 +3,7 @@ import { EventEmitter } from 'events';
 export declare interface SignalManager {
 
     fireTooltipSignal(tooltip: { [key: string]: string }): void;
+    fireEventClickSignal(payload: {[key: string]: number }): void;
```
bhufmann, line 6: we should send the keys and value as strings and the receiving side should convert to the number. Later on we'll add other key/value pairs to sync on.

packages/react-components/src/components/table-output-component.tsx
```diff
@@ -62,11 +63,25 @@ export class TableOutputComponent extends AbstractOutputComponent<TableOutputPro
...
         </div>;
     }
 
+    private onEventClick(event: CellClickedEvent) {
+        const columns = event.columnApi.getAllColumns();
+        const timeStampHeader = columns.find(column => column.getColDef().headerName === "Timestamp ns");
+        if (timeStampHeader) {
+            const timeStamp = timeStampHeader.getColDef().field;
+            const payload = {
+                'timestamp': Number(event.data[timeStamp!])
```
bhufmann, line 79: Send the timestamp as a string

packages/react-components/src/components/timegraph-output-component.tsx
```diff
@@ -141,6 +143,17 @@ export class TimegraphOutputComponent extends AbstractTreeOutputComponent<Timegr
         this.setState({ collapsedNodes: newList });
     }
 
+    private selectionChanged(payload: { [key: string]: number }) {
+        const offset = this.props.viewRange.getOffset();
+        const rawTimeStamp = payload['timestamp'];
```
bhufmann, line 148: Here you need to convert it the string to a number. If conversion fails in the code then ignore signal.

bhufmann (changes requested):
Please update and rebase this patch to latest in master.

muddana-satish:
> Please update and rebase this patch to latest in master.

Addressed all the review comments.

bhufmann (approved):
Looks good to me. Thanks for the improvement.

tahini (approved):
Works well! Thanks!

PR #208, opened 2021-01-07
Fixed placeholder text positioning and updated placeholder text

danieltomasku:
This PR fixes the positioning of the placeholder text when opening a trace. Previously the text was off-centered and non-responsive such that the text would be cut-off in certain viewport widths.

This PR also updates the text based on @ssmagula's recommendations. I understand there is a desire to obviate this step in future iterations, but this should improve things in the meantime!

| Before | After |
| ------ | ------ |
|image|image|

### DIFF
--- a/packages/react-components/src/components/trace-context-component.tsx
+++ b/packages/react-components/src/components/trace-context-component.tsx
@@ -265,7 +265,9 @@ export class TraceContextComponent extends React.Component<TraceContextProps, Tr
 
     private renderPlaceHolder() {
         return <div className='no-output-placeholder'>
-            {'Add outputs by clicking on an analysis in the trace explorer view'}
+            {'Trace loaded successfully.'}
+            <br />
+            {'To see available analyses, open the Trace Explorer view.'}
         </div>;
     }
 
--- a/packages/react-components/src/style/trace-context-style.css
+++ b/packages/react-components/src/style/trace-context-style.css
@@ -1,7 +1,12 @@
 .no-output-placeholder {
     color: var(--theia-ui-font-color3);
     font-size: 24px;
-    transform: translate(30%, 500%);
+    text-align: center;
+    position: relative;
+    transform: translate(0, -50%);
+    top: 50%;
+    width: 90%;
+    align-self: center;
 }
 
 .react-grid-item.react-grid-placeholder {


packages/react-components/src/components/trace-context-component.tsx
```diff
@@ -265,7 +265,7 @@ export class TraceContextComponent extends React.Component<TraceContextProps, Tr
 
     private renderPlaceHolder() {
         return <div className='no-output-placeholder'>
-            {'Add outputs by clicking on an analysis in the trace explorer view'}
+            {'Trace file loaded successfully.\nTo see available analyses, view the Trace Explorer tab.'}
```
tahini, line 268: The \n in the text does not work.

placeholder_text

I think you could just put 2 texts between { } and add a `<br>` in between

tahini (changes requested):

packages/react-components/src/components/trace-context-component.tsx
danieltomasku: Thanks for testing and catching that @tahini! Should be fixed now.

packages/react-components/src/components/trace-context-component.tsx
```diff
@@ -265,7 +265,9 @@ export class TraceContextComponent extends React.Component<TraceContextProps, Tr
 
     private renderPlaceHolder() {
         return <div className='no-output-placeholder'>
-            {'Add outputs by clicking on an analysis in the trace explorer view'}
+            {'Trace file loaded successfully.'}
+            <br />
+            {'To see available analyses, view the Trace Explorer tab.'}
```
tahini, line 270: say you're a first time user and read "view the Trace Explorer tab", where would one look for that "tab"? According to [this](https://code.visualstudio.com/docs/getstarted/userinterface), the Trace Explorer would be in the "Activity bar".

But to be really picky... this file is in react-components, we don't know where it will be used, theia, vscode, other, we don't even know that there will be a Trace Explorer tab! That placeholder text should probably come as a prop to the trace-context-component.

tahini (changes requested):
Text style works as advertised and is well responsive. 

I have some comments on the text itself, that could be tackled in a separate patch if you prefer, I don't want to stop progress!

packages/react-components/src/components/trace-context-component.tsx
```diff
@@ -265,7 +265,9 @@ export class TraceContextComponent extends React.Component<TraceContextProps, Tr
 
     private renderPlaceHolder() {
         return <div className='no-output-placeholder'>
-            {'Add outputs by clicking on an analysis in the trace explorer view'}
+            {'Trace file loaded successfully.'}
```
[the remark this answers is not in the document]
danieltomasku: Good feedback 👍  I've removed 'file' from the sentence.

packages/react-components/src/components/trace-context-component.tsx
danieltomasku: I definitely agree with your points here. I've made the language a little bit more generic to cover use cases that might exist outside of Theia/VS Code, however it doesn't really address making it easier for first time users to know where to go next. I know there have been talks about potentially getting rid of this intermediate placeholder step altogether and re-working the user flow. Perhaps adding a button underneath the text that activates the Trace Explorer view might help so there is no ambiguity of where the tab/activity bar/trace explorer view is? That being said, it's probably out of the scope of this PR to address those changes so I'll keep the placeholder text as fixed rather than as a prop if that's okay. I can also revert the text entirely and just leave the styling changes if you'd like.

ebugden (approved):
Thank you for this improvement! It's simple, effective, and matches the tool's visual style. The instructions catch the eye immediately when they appear.

tahini (approved):
Excellent! Thanks for the update. Theia/vscode is the only consumer of this package yet, so the text for trace explorer is ok. We can make it a prop eventually. For now, it's great improvement!

bhufmann (approved):

PR #211, opened 2021-01-11
Support inband OutputElementStyle for annotations

bhufmann:
To provide annotation markers the output styles needed to defined
in the list of style provided through the fetch style method which
are referenced by the parent style key in the annotation itself.

This patch makes it possible to supply the OutputElementStyle inside
the annotation without using the parent style key.


MatthewKhouzam (approved):
Works great! Now we also proved objectively it does what we want

PR #213, opened 2021-01-12
transforming cursor into cross-arrow when moving is possible #163

### LINKED ISSUE #163: Cursor is an arrow cross when moving is possible
The cursor becoming an arrow cross when the object underneath can be moved supports intuitive understanding by following UI standards set by other tools (ex. Grafana, Google Slides):

cursor-arrow-cross

grafana-arrow-cross-cursor

Ab-rahe:
Transforming the cursor into a cross-arrow when moving is possible
fixes #163

### DIFF
--- a/packages/react-components/src/style/output-components-style.css
+++ b/packages/react-components/src/style/output-components-style.css
@@ -19,6 +19,7 @@
     user-select: none;
     align-self: center;
     justify-self: center;
+    cursor: move;
 }
 .remove-component-button {
     background: none;


packages/react-components/src/style/output-components-style.css
```diff
@@ -9,6 +9,7 @@
     background-color: var(--theia-layout-color4);
     display: grid;
     grid-template-rows: 25px 1fr;
+    cursor: url('cross-arrows.png'), auto ;
```
tahini, line 12: In trace-context-component, the  class used for dragging widget is the `title-bar-label`, so this should be the class whose style to update. The widget-handle is bigger than the title-bar-label, so sometimes, the cursor is the cross-arrows but it is not draggable.

tahini (changes requested):
Nice improvement, where did you get the icon? I'll let @ebugden judge if the icon fits the general style or if it could use the same icon as the one for moving file tabs. 

The current icon becomes invisible when using the light theme, as it does not seem to have a black contour. (to switch to light theme: File -> Settings -> Color Theme

packages/react-components/src/style/output-components-style.css
Ab-rahe: Thanks for your feedback, I didn't know that there was a light theme, so I choosed the icon based on the dark one. I will change it.

I took the icon from: 
 https://favpng.com/png_view/arrow-arrow-cross-cursor-symbol-png/zycH9794

I changed the color since it was dark and resized it to 20 pixels because it was too large.

Concerning your second comment, I will make the changes.

Ab-rahe: @ebugden , here is the video

Do you have any suggestion for the icon I should use ?


https://user-images.githubusercontent.com/46821405/104362505-e158cc80-54e1-11eb-986d-de9ee66c4c5e.mp4

MatthewKhouzam: What license is the icon originally? I can't find anything on their site. :(

Ab-rahe: I can't find it either :/
Since there is no cross-arrow icon (from what I saw) in the CSS libraries, I took an image from the internet just to start.

MatthewKhouzam: image
hand drawn pixel by pixel licensed under epl right here ^ if you guys need it.

MatthewKhouzam: Also if you want a wider arrow, here you go. 
image

Ab-rahe:
@tahini this is how the cursor looks actually. I increased the contrast of the icon that proposed @MatthewKhouzam to make it visible with the light theme and made the changes also to the css file.

https://user-images.githubusercontent.com/46821405/104387874-019a8280-5506-11eb-966c-d49299385c72.mp4

packages/react-components/src/style/output-components-style.css
danieltomasku: Looks like you already have an icon, but one suggestion is to use the native browser `move` cursor:
```css
cursor: move;
```
That way you don't have to worry about licensing and you won't need to add another image asset. You also get accessibility contrast for free. Downside is that you don't get ultimate control of what the cross-arrow icon looks like. On Windows/Chrome, this is roughly what it looks like:

move-cursor-resized

Just a thought! Feel free to ignore if you are happy with your current icon/approach 😃

tahini: I agree with @danieltomasku  that using `cursor: move` solves many issues with hand drawn icons (whose edges are not as clear in light mode). 

moveLightMode


But then, the icon is not the same everywhere. In Linux/Firefox, it is a little fist dragging the window: moveIcon Just how acceptable is that?

ebugden (changes requested):
Some thoughts:
* Let's try using the default browser cursor.
* The point of the arrow needs to line up with the center of the crosshairs when the cursor switches. Right now this is not the case and it looks like the cursor jumps when it switches. I can't show a video of the correct behaviour (my screen cap doesn't show the cursor position I actually see), but you move the panels in the [Grafana live demo](https://play.grafana.org/) to see what I'm expecting.

Thank you for working on this issue!

----

Other related comments, but I will put these in a separate issue:
* The click-drag possibility should be associated with the whole panel title bar. Right now it seems to only be associated with the panel title text.
* Title bar should slightly change colour on hover. This tutorializes that the click-drag is associated with the whole bar 

Again see Grafana demo for the correct behaviour for these two points. 

image

packages/react-components/src/style/output-components-style.css
ebugden: I still think we should take this option. I think it's the obvious good first choice and then we can adjust if users don't like it.

bhufmann:
I'm coming late to the discussion, but I think we should go with default browser cursor `cursor: move` as suggested above. It's already a great UX improvement. Running Theia or VsCode I see that the default cursor is used for other move operations and it would be consistent. Could we get the change to the default cursor merged first?

Ab-rahe:
Thanks for your feedbacks @ebugden , @bhufmann , in this case I will make the changes with the default browser cursor: move

mirsky-work:
LGTM, just please squash your 2 commits (rebasing wouldn't hurt either).

tahini (approved):
Great, thanks! We can squash and rebase from the PR when merging, it's trivial anyway

tahini:
oh sorry, just saw it can't be rebased safely, it's not trivial after all! So please rebase!

tahini:
Looks good now, @mirsky-work you can merge if you wish

PR #225, opened 2021-01-18
Keep the color of status bar blue even if the server is offline

### LINKED ISSUE #214: Should the status bar change color according to the connection status with the server?
The default color of the status bar in Theia is blue. The trace extension changes that default to yellow, to indicate that the trace server is offline (or connection is interrupted):
statusbar-yellow

And changes back to the default blue when the server is started (or connection restored):
statusbar-blue

When the trace extension is integrated into an IDE, the yellow status bar, unless the user pro-actively started the trace server, might be a visual distraction when the user is performing other activities unrelated to tracing, such as writing code. Moreover, it might conflict with other plugins that also change the status bar's color, e.g., [Debugger in VS Code](https://www.youtube.com/watch?v=BYRoxIYP04M).

The suggestion is to avoid changing the color of the status bar, and instead rely solely on the _start/stop server_ button that changes according to the connection status. Thoughts?

@ebugden, @danieltomasku, @bhufmann

ankusht-work:
fixes #214
Signed-off-by: Ankush-Tyagi ankush.tyagi@ericsson.com


mirsky-work:
68747470733a2f2f692e726564642e69742f336d643770356d7839736f792e676966

bhufmann (approved):
Looks good to me and works. Thanks!

MatthewKhouzam (approved):
I'm blue badabeedo badbaddabbeedo.

PR #226, opened 2021-01-18
Support error handling for start/stop trace server

ankusht-work:

This PR adds error handling for the start/stop trace server. It also prevents the user from accidentally killing a process through the preferences settings. 

Steps to test:
- Provide an incorrect path for the executable and click the start trace server button. 
- Provide a busy port in the preferences settings and click the start trace server button.
You should see a Toast message with a detailed error.

### DIFF
--- a/viewer-prototype/src/browser/trace-server-status.ts
+++ b/viewer-prototype/src/browser/trace-server-status.ts
@@ -11,7 +11,7 @@ import { StatusBar, StatusBarAlignment, StatusBarEntry } from '@theia/core/lib/b
 import { ConnectionStatus, AbstractConnectionStatusService } from '@theia/core/lib/browser/connection-status-service';
 import { TspClient } from 'tsp-typescript-client/lib/protocol/tsp-client';
 import { TspClientProvider } from './tsp-client-provider';
-import { TraceServerConfigService } from '../common/trace-server-config';
+import { PortBusy, TraceServerConfigService } from '../common/trace-server-config';
 import { PreferenceService } from '@theia/core/lib/browser';
 import { TRACE_PATH, TRACE_PORT } from './trace-server-preference';
 import { Deferred } from '@theia/core/lib/common/promise-util';
@@ -20,7 +20,7 @@ import { MessageService } from '@theia/core';
 @injectable()
 export class TraceServerConnectionStatusService extends AbstractConnectionStatusService {
 
-    private scheduledPing: number | undefined;
+    private lastPing = new Deferred();
     private tspClient: TspClient;
 
     private constructor(
@@ -37,20 +37,20 @@ export class TraceServerConnectionStatusService extends AbstractConnectionStatus
     }
 
     protected schedulePing(): void {
-        if (this.scheduledPing) {
-            this.clearTimeout(this.scheduledPing);
-        }
-        this.scheduledPing = this.setTimeout(async () => {
+        const ping = async () => {
+            this.lastPing.reject();
+            this.lastPing = new Deferred();
             try {
-                const pingTimeout = new Promise((_, reject) => { setTimeout(reject, this.options.offlineTimeout); });
-                await Promise.race([this.tspClient.fetchExperiments(), pingTimeout]);
+                await Promise.race([this.tspClient.fetchExperiments(), this.lastPing.promise]);
                 this.updateStatus(true);
             } catch (e) {
                 this.updateStatus(false);
                 this.logger.trace(e);
             }
-            this.schedulePing();
-        }, this.options.offlineTimeout * 0.8);
+        };
+
+        ping();
+        setInterval(ping, this.options.offlineTimeout);
     }
 }
 
@@ -110,8 +110,14 @@ export class TraceServerConnectionStatusContribution extends DefaultFrontendAppl
 
         try {
             await this.withTimeout(this.traceServerConfigService.startTraceServer(this.path, this.port));
-        } catch {
-            this.messageService.error('Failed to start trace server.');
+        } catch (err) {
+            if (PortBusy.is(err)) {
+                this.messageService.error(
+                    `Error opening serial port ${this.port}. (Port busy)`);
+            } else {
+                this.messageService.error(
+                    'Failed to start the trace server: no such file or directory. Please make sure that the path is correct in Trace Viewer settings and retry');
+            }
             this.handleOffline();
         }
     }
@@ -120,9 +126,9 @@ export class TraceServerConnectionStatusContribution extends DefaultFrontendAppl
         this.updateStatusBar({ text: '$(sync~spin) Trace server stopping' });
 
         try {
-            await this.withTimeout(this.traceServerConfigService.stopTraceServer(this.port));
-        } catch {
-            this.messageService.error('Failed to terminate trace server.');
+            await this.withTimeout(this.traceServerConfigService.stopTraceServer());
+        } catch (err) {
+            this.messageService.error(`Failed to stop the trace server on port: ${this.port}.`);
             this.handleOnline();
         }
     }
@@ -158,7 +164,7 @@ export class TraceServerConnectionStatusContribution extends DefaultFrontendAppl
 
     protected withTimeout<T = void>(serverAction: Promise<T>): Promise<[T, void]> {
         this.serverPending = new Deferred();
-        setTimeout(this.serverPending.reject, 10000);
+        setTimeout(this.serverPending.reject, 12000);
         return Promise.all([serverAction, this.serverPending.promise]);
     }
 }
--- a/viewer-prototype/src/node/trace-server-service.ts
+++ b/viewer-prototype/src/node/trace-server-service.ts
@@ -1,33 +1,56 @@
-import { spawn, exec } from 'child_process';
+import { spawn } from 'child_process';
 import { injectable } from 'inversify';
-import { TraceServerConfigService } from '../common/trace-server-config';
+import { PortBusy, TraceServerConfigService } from '../common/trace-server-config';
+import treeKill = require('tree-kill');
 @injectable()
 export class TraceServerServiceImpl implements TraceServerConfigService {
+    private processId: number;
 
-    startTraceServer(path: string | undefined, port: number | undefined): Promise<void> {
+    startTraceServer(path: string | undefined, port: number | undefined): Promise<string> {
         const server = spawn(`${path}`, ['-vmargs', `-Dtraceserver.port=${port}`]);
-        return new Promise<void>((resolve, reject) => {
-            // If the server provides output on any channel before it exits, consider that a success.
+        this.processId = server.pid;
+        const timeouts: NodeJS.Timeout[] = [];
+
+        return new Promise<string>((resolve, reject) => {
+            // TODO: Since the resolve never actually gets returned, we need a better way to determine if the child process
+            // initiated by the spawn successfully starts the trace server
+
+            // If the server doesn't error or exit within 2 seconds, we consider that a success.
             // That doesn't mean that the server has really started. On the frontend, the `TraceServerConnectionStatusService`
             // will ping the port until it receives a response, which is the official measure of success.
-            server.stdout.on('data', () => resolve());
-            server.stderr.on('data', () => resolve());
+
+            timeouts.push(setTimeout(() => resolve('success'), 2000));
+
             // If the server exits or errors before it outputs, consider it a failure.
-            server.on('exit', code => reject(code));
-            server.on('error', error => reject(error));
-        })
-            .finally(() => { server.removeAllListeners(); });
+            server.on('error', err => {
+                reject(err);
+            });
+            server.on('exit', code => {
+                reject(PortBusy(code));
+            });
+            // If no response recieved from the trace server in 10 seconds, reject with an error (for internal use)
+            timeouts.push(setTimeout(() => reject('Waited 10 seconds but nothing happened and hence exiting'), 10000));
+
+        }).finally(() => {
+            server.removeAllListeners();
+            timeouts.forEach(timeout => clearTimeout(timeout));
+        });
+
     }
 
-    stopTraceServer(port: number | undefined): Promise<void> {
-        const terminator = exec(`kill -9 $(lsof -t -i:${port} -sTCP:LISTEN)`);  // FIXME: Better approach to kill the server at a given port
-        return new Promise((resolve, reject) => {
-            terminator.on('exit', code => {
-                if (code === 0 || code === 1) {
-                    resolve();
-                } else {
-                    reject(code);
+    stopTraceServer(): Promise<string> {
+        return new Promise<string>((resolve, reject) => {
+            if (this.processId === -1 || !this.processId) {
+                reject('process already killed or process hasn\'t been started');
+                return;
+            }
+            treeKill(this.processId, error => {
+                this.processId = -1;
+                if (error) {
+                    reject(error);
+                    return;
                 }
+                resolve('success');
             });
         });
     }
--- a/viewer-prototype/package.json
+++ b/viewer-prototype/package.json
@@ -30,7 +30,8 @@
     "eslint-plugin-no-null": "^1.0.2",
     "eslint-plugin-react": "^7.20.0",
     "rimraf": "latest",
-    "typescript": "latest"
+    "typescript": "latest",
+    "tree-kill": "latest"
   },
   "scripts": {
     "build": "tsc",
--- a/viewer-prototype/src/common/trace-server-config.ts
+++ b/viewer-prototype/src/common/trace-server-config.ts
@@ -1,15 +1,20 @@
+import { ApplicationError } from '@theia/core';
 
 export const traceServerPath = '/services/trace-server-config';
 export const TraceServerConfigService = Symbol('TraceServerConfigService');
+export const PortBusy = ApplicationError.declare(-32650, code => ({
+    message: 'Port busy',
+    data: { code }
+}));
 export interface TraceServerConfigService {
     /**
      * Spawn the trace server from a given path
      */
-    startTraceServer(path: string | undefined, port: number | undefined): Promise<void>;
+    startTraceServer(path: string | undefined, port: number | undefined): Promise<string>;
 
     /**
      * Stop the trace server
      */
-    stopTraceServer(port: number | undefined): Promise<void>;
+    stopTraceServer(): Promise<string>;
 }
 
--- a/yarn.lock
+++ b/yarn.lock
@@ -15572,6 +15572,11 @@ trash@^6.1.1:
   resolved "https://registry.yarnpkg.com/traverse/-/traverse-0.3.9.tgz#717b8f220cc0bb7b44e40514c22b2e8bbc70d8b9"
   integrity sha1-cXuPIgzAu3tE5AUUwisui7xw2Lk=
 
+tree-kill@latest:
+  version "1.2.2"
+  resolved "https://registry.yarnpkg.com/tree-kill/-/tree-kill-1.2.2.tgz#4ca09a9092c88b73a7cdc5e8a01b507b0790a0cc"
+  integrity sha512-L0Orpi8qGpRG//Nd+H90vFB+3iHnue1zSSGmNOOCh1GLJ7rUKVwV2HvijphGQS2UmhUZewS9VgvxYIdgr+fG1A==
+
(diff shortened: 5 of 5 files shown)


viewer-prototype/src/node/trace-server-service.ts
```diff
@@ -4,8 +4,13 @@ import { TraceServerConfigService } from '../common/trace-server-config';
 @injectable()
 export class TraceServerServiceImpl implements TraceServerConfigService {
 
-    async startTraceServer(path: string | undefined, port: number | undefined): Promise<void> {
-        spawn(`${path}`, ['-vmargs', `-Dtraceserver.port=${port}`]);
+    startTraceServer(path: string | undefined, port: number | undefined): Promise<true | string> {
```
mirsky-work, line 7: So the Promise returns `true` on success and an error message on failure? Can you please document that?
Alternatively, wouldn't it be simpler to always just return an error string (empty on success)?

viewer-prototype/src/node/trace-server-service.ts
```diff
@@ -4,8 +4,13 @@ import { TraceServerConfigService } from '../common/trace-server-config';
...
+        return new Promise((resolve) => {
+            const server = spawn(`${path}`, ['-vmargs', `-Dtraceserver.port=${port}`]);
+            server.on('error', err => resolve(err.message));
+            server.on('message', () => resolve(true));
+        });
+
     }
 
     async stopTraceServer(port: number | undefined): Promise<void> {
```
mirsky-work, line 16: Can we also capture errors stopping the server?

mirsky-work (commented):
Thanks for the contribution! Verified on Gitpod, some comments inline.

viewer-prototype/src/browser/trace-server-status.ts
```diff
@@ -120,11 +122,21 @@ export class TraceServerConnectionStatusContribution extends DefaultFrontendAppl
     }
 
     private async startServer() {
-        await this.traceServerConfigService.startTraceServer(this.path, this.port);
+        // TODO: Add proper error message for the port as well. Right now the server doesn't return any error code or message whenever a port is incorrect and/or the port is busy.
+        try {
+            await this.traceServerConfigService.startTraceServer(this.path, this.port);
+        } catch (err) {
+            this.messageService.error(
+                'Failed to start the trace server: no such file or directory exists. Please make sure that the path is correct in Trace viewer settings and retry');
```
mirsky-work, line 130: I think we can remove the word "exists".
If it's possible to insert a line break in the message, I think it will read better with a line break between the two sentences.

viewer-prototype/src/node/trace-server-service.ts
```diff
@@ -4,11 +4,31 @@ import { TraceServerConfigService } from '../common/trace-server-config';
...
+            });
+        });
+
     }
 
-    async stopTraceServer(port: number | undefined): Promise<void> {
-        exec(`kill -9 $(lsof -t -i:${port} -sTCP:LISTEN)`);  // FIXME: Better approach to kill the server at a given port
+    stopTraceServer(port: number | undefined): Promise<string> {
+        return new Promise<string>((resolve, reject) => {
+            const killProcess = exec(`kill -9 $(lsof -t -i:${port} -sTCP:LISTEN)`);  // FIXME: Better approach to kill the server at a given port
```
marechal-p, line 22: IIUC this Theia extension is able to spawn multiple trace servers, but no more than 1 per port?

If so this class should internally keep a map of "port to child_process" that you can then use to issue a kill on the process pid (like `child_process.kill()`).

viewer-prototype/src/browser/trace-server-status.ts
```diff
@@ -120,11 +122,21 @@ export class TraceServerConnectionStatusContribution extends DefaultFrontendAppl
     }
 
     private async startServer() {
-        await this.traceServerConfigService.startTraceServer(this.path, this.port);
+        // TODO: Add proper error message for the port as well. Right now the server doesn't return any error code or message whenever a port is incorrect and/or the port is busy.
```
marechal-p, line 125: What happens if the port is incorrect or busy at the moment? Is there really nothing that can be leveraged?

viewer-prototype/src/browser/trace-server-status.ts
```diff
@@ -120,11 +122,21 @@ export class TraceServerConnectionStatusContribution extends DefaultFrontendAppl
...
+                'Failed to start the trace server: no such file or directory exists. Please make sure that the path is correct in Trace viewer settings and retry');
+        }
     }
 
     private async stopServer() {
-        await this.traceServerConfigService.stopTraceServer(this.port);
+        try {
+            await this.traceServerConfigService.stopTraceServer(this.port);
+        } catch (err) {
+            this.messageService.error('Failed to stop the trace server');
```
mirsky-work, line 138: Might be useful to report the port, for debugging purposes. Something like:
"Failed to stop the trace server on port <port>"

mirsky-work (commented):
Thanks for addressing the previous issues. Verified the new patch on Gitpod.

viewer-prototype/src/node/trace-server-service.ts
```diff
@@ -4,11 +4,31 @@ import { TraceServerConfigService } from '../common/trace-server-config';
...
-        spawn(`${path}`, ['-vmargs', `-Dtraceserver.port=${port}`]);
+    startTraceServer(path: string | undefined, port: number | undefined): Promise<string> {
+        return new Promise<string>((resolve, reject) => {
+            const server = spawn(`${path}`, ['-vmargs', `-Dtraceserver.port=${port}`]);
+            server.on('error', err => {
+                reject(err.message);
+            });
+            server.on('message', () => {
+                resolve('success');
+            });
```
marechal-p, line 15: I'm surprised this even works. From what I understand, the `child_process` event `message` is fired when the spawned process communicates back using Node's special IPC (which requires an extra pipe to be setup). Are you sure this code works, and if so why?

marechal-p: See:
- https://nodejs.org/docs/latest-v12.x/api/child_process.html#child_process_event_message
- https://nodejs.org/docs/latest-v12.x/api/process.html#process_process_send_message_sendhandle_options_callback

viewer-prototype/src/browser/trace-server-status.ts
bhufmann: @marechal-p the latest trace server from today will exit with a non-zero return value which should be checked in trace-server-service.ts

viewer-prototype/src/node/trace-server-service.ts
bhufmann: @marechal-p  Yes, I commented on #233 the same. The resolve method will never be called. @ankusht-work the reason why the status changes is the periodic scheduled ping which notices the whether the server is up or down.

The thing is that when the server starts successfully, then the child process doesn't return. So, the spawn won't return and the promise won't be resolve successfully. Only error cases can be caught if we add server.on('exit') and check the exit code returned. Not sure how we should handle the successful case. How does is it done with the language server, I wonder?

viewer-prototype/src/node/trace-server-service.ts
bhufmann: I think child_process.kill() will be more deterministic

viewer-prototype/src/node/trace-server-service.ts
marechal-p: So we agree that this snippet is bogus, cool.

I usually check for the PID after the spawn call to check if the process started. This doesn't mean that there won't be an error later, but it's a best effort. From then on you should listen for exit anyway and you'll know if a crash occurred thanks to the return code.

```ts
const child = spawn(...);
child.on('error', reject);
// If the executable is not found, PID will be undefined:
if (typeof child.pid === 'number') {
    resolve();
}
```

marechal-p: Else you could wait for a specific output line to show up. I'm always on the fence for this kind of stdout reading but it should be rather reliable depending on what the server does. If you try doing that you should look at [`readline`](https://nodejs.org/docs/latest-v12.x/api/readline.html#readline_readline_createinterface_options)

viewer-prototype/src/node/trace-server-service.ts
ankusht-work: > IIUC this Theia extension is able to spawn multiple trace servers, but no more than 1 per port?
> Yes
> If so this class should internally keep a map of "port to child_process" that you can then use to issue a kill on the process pid (like `child_process.kill()`).

I have actually tried using process.kill(pid) but ran into an interesting behavior.

**spawn(`${path}`, ['-vmargs', `-Dtraceserver.port=${port}`]);** 
The above spawn command will return the UNIX pid. The parameters provided within this spawn command is the path to the java executable which, when executed successfully will have a different process id. Hence when you execute process.kill(pid) [returned from the spawn], it won't kill the trace-server process running on the back-end. Instead, we will have to recursively kill the entire PID tree, in this case, to be able to successfully kill the trace-server as well (working on it as we speak).

There is also an issue where a user can kill any process by modifying the port parameter and running the stop trace server command. Ideally, user should only be allowed to kill the process started by them. This is also a work in progress and hence the reason why this PR is in Draft state :)

viewer-prototype/src/browser/trace-server-status.ts
ankusht-work: > What happens if the port is incorrect or busy at the moment? Is there really nothing that can be leveraged?

Earlier it couldn't be done however with bernd's latest server patch we should be able to leverage the exit code that is returned from the trace server.

viewer-prototype/src/node/trace-server-service.ts
marechal-p: > [...] when executed successfully will have a different process id.

Ok so if the process started by `${path}` spins another process, are stdio still plugged to the spawned process?

Also, it seems like your are implying that multiple users could start different Trace Servers on the same machine. Here each user has to explicitly set the port and make sure it's not one already in use. A solution to this UX problem is to use port 0 that usually means "let the OS pick a free one". The issue then is to know which one was picked but it's often easy to fetch, then you might want to send it up to the parent, through stdout for instance.

> Instead, we will have to recursively kill the entire PID tree [...]

Yup, see [pidtree](https://www.npmjs.com/package/pidtree) for that. You'll still need to keep track of the spawned PID as they should be the tree's root.

marechal-p: > There is also an issue where a user can kill any process by modifying the port parameter and running the stop trace server command.

If the port is fetched from the preferences, you should listen for changes and act accordingly. It could be killing the server, maybe restarting it with the new port. Depends on the UX you want.

viewer-prototype/src/browser/trace-server-status.ts
```diff
@@ -120,7 +122,16 @@ export class TraceServerConnectionStatusContribution extends DefaultFrontendAppl
     }
 
     private async startServer() {
-        await this.traceServerConfigService.startTraceServer(this.path, this.port);
+        const fileNotFound = "ENOENT";
```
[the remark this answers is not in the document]
ankusht-work: Not used anymore.

viewer-prototype/src/browser/trace-server-status.ts
```diff
@@ -120,7 +122,16 @@ export class TraceServerConnectionStatusContribution extends DefaultFrontendAppl
...
+        //TODO: Add proper error message for the port as well. Right now the server doesn't return any error code or message whenever a port is incorrect and/or the port is busy.
+        try {
+            const startedServer = await this.traceServerConfigService.startTraceServer(this.path, this.port);
+            if (startedServer !== true && startedServer.includes(fileNotFound)) {
+                this.messageService.error("Error: no such file or directory exists. Please make sure that the path is correct in Trace viewer settings and retry");
```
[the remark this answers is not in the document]
ankusht-work: Done.

viewer-prototype/src/node/trace-server-service.ts
ankusht-work: Modified the logic.

viewer-prototype/src/node/trace-server-service.ts
ankusht-work: Done

viewer-prototype/src/browser/trace-server-status.ts
ankusht-work: Done.

viewer-prototype/src/browser/trace-server-status.ts
ankusht-work: Done

viewer-prototype/src/browser/trace-server-status.ts
ankusht-work: updated

viewer-prototype/src/node/trace-server-service.ts
ankusht-work: Thanks for the feedback Paul. 
I am using [tree-kill](https://www.npmjs.com/package/tree-kill) to kill the process tree. This one has more downloads and hence this npm package looked more attractive :).

viewer-prototype/src/node/trace-server-service.ts
ankusht-work: More clarification as to how this is working is provided in [226](https://github.com/theia-ide/theia-trace-extension/pull/233), I have not been able to figure out yet a proper solution to handle the resolve case. 
Below are the assumptions we are running with for now:
1) If the spawn itself fails (for ex: the path to the executable is incorrect)   
       -----error listener will return the error message back to the front-end.
2) If the port is busy:
       -----exit listener will return the exit code back 
3)  If the process generates any output or message, consider that as a success.

viewer-prototype/src/node/trace-server-service.ts
```diff
@@ -1,33 +1,51 @@
...
-            .finally(() => { server.removeAllListeners(); });
+
+            server.stdout.on('data', () => setTimeout(()=> resolve('success'), 2000));
+            server.stderr.on('data', () => setTimeout(()=> resolve('success'), 2000));
+             // If the server exits or errors before it outputs, consider it a failure.
+            server.on('error', err => {
+                reject(err);
+            });
+            server.on('exit', code => {
+                reject(PortBusy(code));
```
bhufmann, line 27: maybe code 0 can be seen as success as well. Non-zero code values are for the reject then.

viewer-prototype/src/browser/trace-server-status.ts
```diff
@@ -110,8 +110,14 @@ export class TraceServerConnectionStatusContribution extends DefaultFrontendAppl
...
         try {
             await this.withTimeout(this.traceServerConfigService.startTraceServer(this.path, this.port));
-        } catch {
-            this.messageService.error('Failed to start trace server.');
+        } catch (err) {
+            if (PortBusy.is(err)) {
+                this.messageService.error(
+                    `Error opening serial port ${this.port}. (Port busy)`);
+            } else {
+                this.messageService.error(
```
bhufmann, line 118: Sometimes it seems the front-end seems to timeout even after successful start in the backend and I get this toaster message. Mabye front-end ping code can be updated to avoid that.

bhufmann (changes requested):

colin-grant-work:
I think we're still running into some problems with the variable time of the pinging in the `TraceServerConnectionStatusService` when the connection to the server is slow. I'd propose the following changes to the `TraceServerConnectionStatusService`:

- replace `private scheduledPing: number | undefined;` with `private lastPing = new Deferred();`
- change the `schedulePing` function to the following:

```typescript
    protected schedulePing(): void {
        const ping = async () => {
            this.lastPing.reject();
            this.lastPing = new Deferred();
            try {
                await Promise.race([this.tspClient.fetchExperiments(), this.lastPing.promise]);
                this.updateStatus(true);
            } catch (e) {
                this.updateStatus(false);
                this.logger.trace(e);
            }
        };

        ping();

        setInterval(ping, this.options.offlineTimeout);
    }
```

The current code still allows up to a 9 second gap between state updates, but this should guarantee at least one state update in every five-second period after startup, though the actual gap between updates may vary.

viewer-prototype/src/node/trace-server-service.ts
```diff
@@ -1,33 +1,51 @@
...
-            server.stdout.on('data', () => resolve());
-            server.stderr.on('data', () => resolve());
-            // If the server exits or errors before it outputs, consider it a failure.
-            server.on('exit', code => reject(code));
-            server.on('error', error => reject(error));
-        })
-            .finally(() => { server.removeAllListeners(); });
+
+            server.stdout.on('data', () => setTimeout(()=> resolve('success'), 2000));
+            server.stderr.on('data', () => setTimeout(()=> resolve('success'), 2000));
```
bhufmann, line 21: if we do only "setTimeout(()=> resolve('success'), 2000)", then it's seen as successful start. This timer will give the server startup with an error which will be rejected below. The front-end ping will notice if the server is up.

ankusht-work:
> I think we're still running into some problems with the variable time of the pinging in the `TraceServerConnectionStatusService` when the connection to the server is slow. The current code still allows up to a 9 second gap between state updates, but it's desirable to report more often. I'd propose the following changes to the `TraceServerConnectionStatusService`:
> 
> * replace `private scheduledPing: number | undefined;` with `private lastPing = new Deferred();`
> * change the `schedulePing` function to the following:
> 
> ```ts
>     protected schedulePing(): void {
>         const ping = async () => {
>             this.lastPing.reject();
>             this.lastPing = new Deferred();
>             try {
>                 await Promise.race([this.tspClient.fetchExperiments(), this.lastPing.promise]);
>                 this.updateStatus(true);
>             } catch (e) {
>                 this.updateStatus(false);
>                 this.logger.trace(e);
>             }
>         };
> 
>         ping();
> 
>         setInterval(ping, this.options.offlineTimeout);
>     }
> ```
> 
> This version will report in every five-second period after startup, although the actual time between updates may vary.
> 
> Alternatively, this code will report status (more or less) exactly every five seconds:
> 
> ```ts
>     protected schedulePing(): void {
>         let timeout = new Deferred();
>         let lastPing: Promise<unknown> = Promise.reject();
> 
>         const ping = async () => {
>             timeout.reject();
>             timeout = new Deferred();
>             try {
>                 await lastPing;
>                 this.updateStatus(true);
>             } catch (e) {
>                 this.updateStatus(false);
>                 this.logger.trace(e);
>             } finally {
>                 lastPing = Promise.race([this.tspClient.fetchExperiments(), timeout.promise]);
>             }
>         };
> 
>         ping();
> 
>         setInterval(ping, this.options.offlineTimeout);
>     }
> ```
> 
> That has the benefit of regularity, but the downside that it's no faster for success than for failure.

Bernd pointed out some issues with gitpod. It looks like we were running into timeout issues likely due to the slowness of the start and stop process on Gitpod. We decided to modify the changes proposed by @colin-grant-work in this patch itself. 
Note: One change to the assumption is that if we do not get an error or exit within 2 seconds, we are considering it as a success.

PatrickTasse:
I tried pressing "Start trace server" right after stopping it. After 10 seconds I got a toaster message "Failed to start the trace server: no such file or directory. Please make sure that the path is correct in Trace Viewer settings and retry".

Not only is the path setting correct and the file exists, but 60 seconds later I can see that the server is started and connection is established.

This reproduces only intermittently.

viewer-prototype/src/browser/trace-server-status.ts
bhufmann: Increasing the timeout in withTimeout() below gives the scheduled ping time to clear the serverPending timer in time. Could you please increase it to 12000?

bhufmann (changes requested):
One more comment below

bhufmann:
> I tried pressing "Start trace server" right after stopping it. After 10 seconds I got a toaster message "Failed to start the trace server: no such file or directory. Please make sure that the path is correct in Trace Viewer settings and retry".
> 
> Not only is the path setting correct and the file exists, but 60 seconds later I can see that the server is started and connection is established.
> 
> This reproduces only intermittently.

@PatrickTasse  Yes, I've seen it. We'll increase the timeout. In a separate patch we'll revise the startup procedure and error handling. I'd like to merge this one because it fixes the stop procedure by using a npm library and the PID.

MatthewKhouzam (approved):
Thanks!

bhufmann (approved):
Looks good to me. Works in GitPod and my development environment. Thanks for the contribution!

PR #233, opened 2021-01-20
Pending Message on Server Start/Stop

colin-grant-work:

This PR adds a temporary status message when the user gestures to start or stop the trace server.

To test:
1. Point your trace viewer executable and port preferences to good locations
2. Click `Start trace server` in the bottom left status bar.
3. See a message that indicates that server is starting.
4. ... wait < 10 seconds
5. See a change indicating that the server has started
6. Kill the server, see the same message

## Unhappy paths:
1. Point your preferences to a bad location (e.g. non-existent file, occupied port)
2. Try to start server
3. In <= 10 seconds you should see a failure message, and the display should revert to the previous state.

There is substantial overlap with #226, which supplies more exact error handling.

### DIFF
--- a/viewer-prototype/src/browser/trace-server-status.ts
+++ b/viewer-prototype/src/browser/trace-server-status.ts
@@ -7,14 +7,15 @@
 import { inject, injectable, postConstruct } from 'inversify';
 import { ILogger } from '@theia/core/lib/common/logger';
 import { DefaultFrontendApplicationContribution } from '@theia/core/lib/browser/frontend-application';
-import { StatusBar, StatusBarAlignment } from '@theia/core/lib/browser/status-bar/status-bar';
-import { Disposable, DisposableCollection } from '@theia/core/lib//common';
-import { ConnectionStatusService, ConnectionStatus, AbstractConnectionStatusService } from '@theia/core/lib/browser/connection-status-service';
+import { StatusBar, StatusBarAlignment, StatusBarEntry } from '@theia/core/lib/browser/status-bar/status-bar';
+import { ConnectionStatus, AbstractConnectionStatusService } from '@theia/core/lib/browser/connection-status-service';
 import { TspClient } from 'tsp-typescript-client/lib/protocol/tsp-client';
 import { TspClientProvider } from './tsp-client-provider';
 import { TraceServerConfigService } from '../common/trace-server-config';
 import { PreferenceService } from '@theia/core/lib/browser';
 import { TRACE_PATH, TRACE_PORT } from './trace-server-preference';
+import { Deferred } from '@theia/core/lib/common/promise-util';
+import { MessageService } from '@theia/core';
 
 @injectable()
 export class TraceServerConnectionStatusService extends AbstractConnectionStatusService {
@@ -41,7 +42,8 @@ export class TraceServerConnectionStatusService extends AbstractConnectionStatus
         }
         this.scheduledPing = this.setTimeout(async () => {
             try {
-                await this.tspClient.fetchExperiments();
+                const pingTimeout = new Promise((_, reject) => { setTimeout(reject, this.options.offlineTimeout); });
+                await Promise.race([this.tspClient.fetchExperiments(), pingTimeout]);
                 this.updateStatus(true);
             } catch (e) {
                 this.updateStatus(false);
@@ -55,17 +57,24 @@ export class TraceServerConnectionStatusService extends AbstractConnectionStatus
 @injectable()
 export class TraceServerConnectionStatusContribution extends DefaultFrontendApplicationContribution {
 
-    @inject(PreferenceService) protected readonly preferenceService: PreferenceService;
+    static readonly STATUS_BAR_ID = 'trace-connection-status';
+    static readonly SERVER_OFFLINE_CLASSNAME = 'traceserver-mod-offline';
 
-    private path: string | undefined;
-    private port: number | undefined;
+    @inject(StatusBar) protected readonly statusBar: StatusBar;
+    @inject(ILogger) protected readonly logger: ILogger;
+    @inject(PreferenceService) protected readonly preferenceService: PreferenceService;
+    @inject(TraceServerConfigService) protected readonly traceServerConfigService: TraceServerConfigService;
+    @inject(TraceServerConnectionStatusService) protected readonly connectionStatusService: TraceServerConnectionStatusService;
+    @inject(MessageService) protected readonly messageService: MessageService;
+
+    protected path: string | undefined;
+    protected port: number | undefined;
+
+    protected serverPending = new Deferred<void>();
 
     @postConstruct()
-    async init(): Promise<void> {
-
-        this.path = this.preferenceService.get(TRACE_PATH);
-        this.port = this.preferenceService.get(TRACE_PORT);
-
+    protected async init(): Promise<void> {
+        this.connectionStatusService.onStatusChange(state => this.onStateChange(state));
         this.preferenceService.onPreferenceChanged(event => {
             if (event.preferenceName === TRACE_PORT) {
                 this.port = event.newValue;
@@ -74,19 +83,10 @@ export class TraceServerConnectionStatusContribution extends DefaultFrontendAppl
                 this.path = event.newValue;
             }
         });
-    }
 
-    @inject(TraceServerConfigService) protected readonly traceServerConfigService: TraceServerConfigService;
+        this.path = this.preferenceService.get(TRACE_PATH);
+        this.port = this.preferenceService.get(TRACE_PORT);
 
-    protected readonly toDisposeOnOnline = new DisposableCollection();
-
-    constructor(
-        @inject(TraceServerConnectionStatusService) protected readonly connectionStatusService: ConnectionStatusService,
-        @inject(StatusBar) protected readonly statusBar: StatusBar,
-        @inject(ILogger) protected readonly logger: ILogger
-    ) {
-        super();
-        this.connectionStatusService.onStatusChange(state => this.onStateChange(state));
         if (this.connectionStatusService.currentStatus === ConnectionStatus.ONLINE) {
             this.handleOnline();
         }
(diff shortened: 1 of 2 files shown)


MatthewKhouzam:
This is fixing part of #234

mirsky-work:
Thanks for this very useful addition!
Verified on Gitpod, works well.

viewer-prototype/src/node/trace-server-service.ts
```diff
@@ -4,11 +4,24 @@ import { TraceServerConfigService } from '../common/trace-server-config';
 @injectable()
 export class TraceServerServiceImpl implements TraceServerConfigService {
 
-    async startTraceServer(path: string | undefined, port: number | undefined): Promise<void> {
-        spawn(`${path}`, ['-vmargs', `-Dtraceserver.port=${port}`]);
+    startTraceServer(path: string | undefined, port: number | undefined): Promise<void> {
+        const server = spawn(`${path}`, ['-vmargs', `-Dtraceserver.port=${port}`]);
+        return new Promise((resolve, reject) => {
+            server.on('message', () => resolve());
```
bhufmann, line 10: As far as I understand, message is only called when there are messages send between parent and child process. I have seen code when doing a google search, that message is used with a fork and not spawn. In my environment the promise never resolves here.

viewer-prototype/src/node/trace-server-service.ts
```diff
@@ -4,11 +4,24 @@ import { TraceServerConfigService } from '../common/trace-server-config';
 @injectable()
 export class TraceServerServiceImpl implements TraceServerConfigService {
 
-    async startTraceServer(path: string | undefined, port: number | undefined): Promise<void> {
-        spawn(`${path}`, ['-vmargs', `-Dtraceserver.port=${port}`]);
+    startTraceServer(path: string | undefined, port: number | undefined): Promise<void> {
+        const server = spawn(`${path}`, ['-vmargs', `-Dtraceserver.port=${port}`]);
```
bhufmann, line 8: Spawning the child process will start the server, if it's success full, the server process will never exit. So, I don't think we will be able to get a resolve beside the timeout. I actually notice that when running it on my laptop. I get a timeout in the calling method and the error dialog is shown, while there server is actually running. We need to find another way to determine that the server is actually started successfully.

viewer-prototype/src/node/trace-server-service.ts
```diff
@@ -4,11 +4,24 @@ import { TraceServerConfigService } from '../common/trace-server-config';
...
+        return new Promise((resolve, reject) => {
+            server.on('message', () => resolve());
+            server.on('error', () => reject());
+        });
```
bhufmann, line 12: With the latest update in the trace server which can be downloaded here (https://download.eclipse.org/tracecompass.incubator/trace-server/rcp/), the server will return a non-zero code if it cannot be started successfully, e.g. when server is already running, port is taken etc. So, you need to add server.on('exit') and reject it. I tried it and worked.

viewer-prototype/src/browser/trace-server-status.ts
```diff
@@ -105,40 +104,62 @@ export class TraceServerConnectionStatusContribution extends DefaultFrontendAppl
...
 
-    private statusbarId = 'trace-connection-status';
+    protected async startServer(): Promise<void> {
+        this.updateStatusBar({ text: '$(sync~spin) Trace server starting' });
+
+        try {
+            await this.withTimeout(this.traceServerConfigService.startTraceServer(this.path, this.port));
+        } catch {
+            this.messageService.error('Failed to start trace server.');
+            this.handleOffline();
```
bhufmann, line 114: I think for start of the server, it will always timeout when the server is started successfully because the spawned process doesn't exit in that case. Calling handleOffline() will change indicate that the sever needs to be started, however the server is actually running.

viewer-prototype/src/node/trace-server-service.ts
bhufmann, line 8: I'm not suggesting to merge the following lines, but during debugging it can be useful because you'll see what the server outputs to the console:
         server.stdout.on('data', data => {
            console.log(\`stdout:\n${data}\`);
          });
          server.stderr.on('data', data => {
            console.error(\`stderr: ${data}\`);
          });

bhufmann (changes requested):
I really like the idea of giving visual feedback during start and stop. On GitPod, an older version of the trace server is used that doesn't return an non-zero error code if the start up fails. The GitPod workspace cache is not updated for me. So, initially it looked good. However, when testing it on my local laptop, the solution doesn't behave behind the scenes as expected.

bhufmann:
I think why it works on GitPod is because the scheduledPing() is setting the status bar correctly. If you'd like to reproduce the startup timeout uncomment out the scheduledPing() and change the initial state of the server as offline so that the start button is active. Pressing the start button will timeout and the error message will appear while the server is actually running. 

BTW, on my laptop for some reason it takes longer for the scheduledPing() to kick in and that's why it happens even without removing the ping.

colin-grant-work:
Hey @bhufmann, thanks for the feedback. It sounds like things are behaving quite a bit differently in your environment than they are in mine. Could you tell me a little more about your set up (OS and server executable in particular) so that I can try to reproduce what you're seeing?

colin-grant-work:
@bhufmann, one problem here is that defining 'success' in starting the process is not entirely straightforward. One server executable I'm using directs all output to `stderr`, so receiving `stderr` data may count as a success. On the other hand, if the trace compass server finds a bad Java environment, it emits an empty string to `stderr`, then shows a dialogue, and only exits once the user has interacted with the dialog. Since the problem you mentioned was that the backend was never resolving correctly, I'm inclined to let the the backend resolve prematurely if it detects _any_ output, and then rely on the timeout to reject if the server never in fact registers as online. This should prevent the program from getting into a state where the frontend knows about a change (offline -> online) before the backend has signaled its status.

viewer-prototype/src/node/trace-server-service.ts
colin-grant-work: I've switched this to listen for output on `stdout` and `stderr` and resolve the promise if output is detected on either channel.

viewer-prototype/src/node/trace-server-service.ts
colin-grant-work: Yes, determining 'success' is difficult here. I think the best we can do is detect whether the process emits output of any kind, and resolve if it does. This means that it will reject if the spawn itself produces an error, but it would struggle with a server that started and ran completely silently.

bhufmann:
> Hey @bhufmann, thanks for the feedback. It sounds like things are behaving quite a bit differently in your environment than they are in mine. Could you tell me a little more about your set up (OS and server executable in particular) so that I can try to reproduce what you're seeing?
@colin-grant-work yes, let's have chat. Ping me at work. Maybe include @ankusht-work since he is also looking at this.

viewer-prototype/src/node/trace-server-service.ts
colin-grant-work: I've added an exit listener as well, though it will not be fired if the server has first emitted output. It would be possible to do something like 
```
onOutput
   timeout(resolve); // leave time to reject if it exits near-immediately.
onExit
  reject; // immediately
```
such that exiting within a certain interval, even if output were produced before it, would count as a rejection, but that would be somewhat arbitrary. Still, willing to consider that if it seems like the best approach.

marechal-p:
@colin-grant-work @bhufmann would it be worthwile to design the Trace Server to be easier to spawn and handle, as in standardizing its output and not showing dialogs that could block the startup? This way you would safely listen for stdout and/or stderr for the OK message, or listen for some error message or some exit code?

bhufmann:
> @colin-grant-work @bhufmann would it be worthwile to design the Trace Server to be easier to spawn and handle, as in standardizing its output and not showing dialogs that could block the startup? It could be set with a cli flag perhaps? This way you would safely listen for stdout and/or stderr for the OK message, or listen for some error message or some exit code?

@marechal-p Unfortunately, the Trace Compass server startup is not straight forward to change. The print-outs are actually not created by the Trace Compass code, it's coming from libraries that we use as well as in the error case where the server returns non-zero the process launcher does some error print-out as well before exiting. We can't disable them. Also, the case of a successful start of the server, the process doesn't exit doesn't help.

Having said that, I think we should investigate how the startup procedure of the server so that it's more predictable and the Theia extension hopefully can make use of a deterministic behavior. 

I think this patch is already an improvement, because it gives visual feedback that something is starting (or stopping). I tested it in GitPod and my company internal environment and it works as described. 

So, I suggest to merge it as is and improve on it in a subsequent activity.

Any thoughts?

bhufmann:
> @bhufmann, one problem here is that defining 'success' in starting the process is not entirely straightforward. One server executable I'm using directs all output to `stderr`, so receiving `stderr` data may count as a success. On the other hand, if the trace compass server finds a bad Java environment, it emits an empty string to `stderr`, then shows a dialogue, and only exits once the user has interacted with the dialog. Since the problem you mentioned was that the backend was never resolving correctly, I'm inclined to let the the backend resolve prematurely if it detects _any_ output, and then rely on the timeout to reject if the server never in fact registers as online. This should prevent the program from getting into a state where the frontend knows about a change (offline -> online) before the backend has signaled its status.

Sounds like a good compromise.

viewer-prototype/src/node/trace-server-service.ts
```diff
@@ -4,11 +4,29 @@ import { TraceServerConfigService } from '../common/trace-server-config';
...
-    async startTraceServer(path: string | undefined, port: number | undefined): Promise<void> {
-        spawn(`${path}`, ['-vmargs', `-Dtraceserver.port=${port}`]);
+    startTraceServer(path: string | undefined, port: number | undefined): Promise<void> {
+        const server = spawn(`${path}`, ['-vmargs', `-Dtraceserver.port=${port}`]);
+        return new Promise<void>((resolve, reject) => {
+            // If the server provides output on any channel before it exits, consider that a success.
+            server.stdout.on('data', () => resolve());
+            server.stderr.on('data', () => resolve());
+            // If the server exits or errors before it outputs, consider it a failure.
+            server.on('exit', code => reject(code));
```
marechal-p, line 14: FYI: The `exit` event passes 2 parameters: `code` and `signal`. Only one is set while the other is `null`.

viewer-prototype/src/node/trace-server-service.ts
```diff
@@ -4,11 +4,29 @@ import { TraceServerConfigService } from '../common/trace-server-config';
...
+            server.on('error', error => reject(error));
+        })
+            .finally(() => { server.removeAllListeners(); });
```
tahini, line 17: Here's a use case: another service that is not the trace server is running on port 8080. Data gets shown on stderr, so the promise resolve is called, but the server exits right after, but since you removed all listeners, it does not catch the exit. But since the server is never online, the timeout is hit and the message appears correctly after 8 seconds (instead of earlier as it could have been otherwise, fwiw here).

You can probably keep the server.on('exit') listener in general, so to detect the server process exiting at any moment (like a computer can decide to kill it if it takes too much RAM or CPU). If this happens without the user having asked for it, we can show her a message.

tahini (commented):
This is great improvement for starting the server! And it works generally well, though not necessarily because of the intended code path...

iiuc, it is the scheduledPing that decides whether the server is running or not, as the case of the other service running on port 8080 shows (see comment). If so, _for this patch_ (as #226 has a different, more error management purpose), you might as well make it clear in the code that you try to spawn the process and hope for the ping to come after X seconds, otherwise, you assume the startup failure and show a message.

In short, I'd say, don't bother with the server listeners here, just spawn and wait for ping since that's how it's really making this work! But that's just my opinion

viewer-prototype/src/node/trace-server-service.ts
colin-grant-work: I agree that that case deserves to be handled, but it can't be handled by the promise as currently structured. Since promises can only be rejected or resolve once - any call to `resolve` or `reject` thereafter is a no-op - once we have resolved or rejected the promise for any reason, no listeners can affect the state of the promise after that, and it's better to dispose of them. I suggest that the use case you mention be part of @ankusht-work's work on #226, which will include more detailed error handling for a wider range of cases. An additional argument for separating the two is that the server going down for any reason at any time is a problem that should be brought to the user's attention, and it it's worth having dedicated machinery outside of startup checks.

colin-grant-work: However, if you'd like to see it handled here, one option would be to structure the promise logic so that we wait to resolve it on (seemingly) successful startup for a second or two so that the exit handler has time to catch startups with immediate termination before we resolve the promise.

tahini: > > In short, I'd say, don't bother with the server listeners here, just spawn and wait for ping since that's how it's really making this work! But that's just my opinion
> 
> I'm also happy to take this route - I think you're right that ignoring the backend is fine for now, since it's the `ConnectionStateService` that plays the decisive role. On the other hand, the `error` listener does catch bad configurations earlier than waiting for a timeout.

I guess the real message in my comment is "don't keep code that doesn't really do what it's intended to do, like the resolves above". But if the rejects do have their utility, then yes keep them! And the port-in-use use case can indeed be handled by @ankusht-work 's patch. Sorry for the confusion.

Also, whatever you keep or don't keep, please add the proper comments in the code so we know how the status is actually handled, for instance, if  you need to keep the resolution, please say that resolution here does not mean the server is really started, the ping will do.

colin-grant-work: @tahini, thanks for the feedback. I have added some detail to my comments, but left things otherwise unchanged. If we're interested in the possibility of rejection, we have to keep the resolution as well so the promise doesn't just hang.

PatrickTasse: When I tried it this morning with gitpod, after clicking "Start trace server", 8 seconds later I got a toaster error "Failed to start server". Then after 22 seconds total, the status changed to "Stop trace server".

colin-grant-work: So that problem can occur if the response to a bad ping is very slow. The `TraceServerConnectionStateService` only schedules a new ping when the last one is finished, whether it's successful or not. If a bad ping takes a long time to signal its failure, you can get into a no-man's land where you try to start the server (and it starts) during the time in which the `TraceServerConnectionStateService` is waiting on bad ping, and so doesn't make a good ping (which is usually much faster) until the time out has passed. One way to avoid this would be to timeout the pings in the `TraceServerConnectionStateService` so that it never waits too long before trying again. I'm happy to incorporate that fix into this PR, if you would like.

colin-grant-work: I've modified the `TraceServerConnectionStatusService` with these lines:

```
const pingTimeout = new Promise((_, reject) => { setTimeout(reject, this.options.offlineTimeout); });
await Promise.race([this.tspClient.fetchExperiments(), pingTimeout]);
```

That should mean that any ping that takes more than the `offlineTimeout` (default of 5 seconds, it looks like) to settle (successfully or otherwise) will be considered a failure. That seems consistent with the intent of having an `offlineTimeout`, and should prevent the `Failed to start server.` message from ever appearing during a cycle during which the `TraceServerStatusService` would have signaled success - the timeout on the message waits two full cycles. It's still conceivable that the server might have started and simply take a long time to become ready to handle requests, but that contingency seems better left to the backend, or the entire strategy of depending on regular pinging to determine whether the server is active needs to be rethought.

bhufmann (approved):

tahini (approved):
Excellent, thanks for the contribution!

PR #241, opened 2021-01-25
Update total height on toggle of node collapse/expand state

PatrickTasse:


MatthewKhouzam (approved):
Makes sense! I am nervous about the setState with promises, but I tried this, and nothing broke. :)

bhufmann (changes requested):
Using a kernel trace I notice that the scrolling still was possible when the trace node was collapsed.

bhufmann (approved):
Looks good to me. Thanks!

PR #205, opened 2020-12-19
reorganize the project

marechal-p:
See commit messages.

PR #244, opened 2021-01-27
Make the quick start work on windows

arfio:
Use the shx library to make cp and mkdir command work under windows
terminal.


bhufmann (approved):
Looks good to me and still works in Linux. Don't have a way to check Windows though.

bhufmann:
@arfio Sorry, #205 was just merged. Your patch has some merge conflicts. Could you please resolve them? I think there are new "mkdir" commands added in one or more package.json file. If true, could you please update them too?

arfio:
@bhufmann This merge request is no longer useful, the mkdir and copy commands have been removed from the scripts with the #205. Now the build works without the need of the shx package.

Also, building on Windows for this package is not needed for now, because of the following issue with Theia: https://github.com/eclipse-theia/theia/issues/8098.

bhufmann:
> @bhufmann This merge request is no longer useful, the mkdir and copy commands have been removed from the scripts with the #205. Now the build works without the need of the shx package. I'm not sure where the "mkdir" commands have been added but the options used must be compatible with the windows alias.
> 
> Also, building on Windows for the theia-trace-extension is not needed for now, because of the following issue with Theia: [eclipse-theia/theia#8098](https://github.com/eclipse-theia/theia/issues/8098).

Thanks for letting me know.

PR #245, opened 2021-01-27
Allow timeline gridlines to extend to the end of the view, not the tr…

MatthewKhouzam:
Not sure if it works, to be confirmed.


bhufmann:
Upgrade to latest timeline-chart (#247) included a fix for this issue. See https://github.com/theia-ide/timeline-chart/pull/106 for details of the fix. I'll close this one. Thanks for looking into this.

PR #248, opened 2021-01-28
Decouple delete experiments from trace-viewer

### LINKED ISSUE #149: Trace Explorer and Trace Viewer tabs out of sync

muddana-satish:
On Click of 'x' in trace-viewer tab doesn't delete the experiment from trace explorer.
Added a context menu to open, close and delete traces.

fixes #149

### DIFF
--- a/theia-extensions/viewer-prototype/src/browser/trace-explorer/trace-explorer-sub-widgets/trace-explorer-opened-traces-widget.tsx
+++ b/theia-extensions/viewer-prototype/src/browser/trace-explorer/trace-explorer-sub-widgets/trace-explorer-opened-traces-widget.tsx
@@ -5,13 +5,16 @@ import { List, ListRowProps, Index, AutoSizer } from 'react-virtualized';
 import { Experiment } from 'tsp-typescript-client/lib/models/experiment';
 import { ExperimentManager } from '@trace-viewer/base/lib/experiment-manager';
 import { OutputDescriptor } from 'tsp-typescript-client/lib/models/output-descriptor';
-import { Emitter } from '@theia/core';
+import { Emitter, CommandService } from '@theia/core';
 import { TspClientProvider } from '../../tsp-client-provider';
 import { signalManager, Signals } from '@trace-viewer/base/lib/signal-manager';
 import { TraceExplorerTooltipWidget } from './trace-explorer-tooltip-widget';
 import ReactModal from 'react-modal';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import { faCopy } from '@fortawesome/free-solid-svg-icons';
+import { ContextMenuRenderer } from '@theia/core/lib/browser';
+import { TraceExplorerMenus } from '../trace-explorer-commands';
+import { TraceViewerCommand } from '../../trace-viewer/trace-viewer-commands';
 
 @injectable()
 export class TraceExplorerOpenedTracesWidget extends ReactWidget {
@@ -30,6 +33,8 @@ export class TraceExplorerOpenedTracesWidget extends ReactWidget {
 
     @inject(TspClientProvider) protected readonly tspClientProvider!: TspClientProvider;
     @inject(TraceExplorerTooltipWidget) protected readonly tooltipWidget!: TraceExplorerTooltipWidget;
+    @inject(ContextMenuRenderer) protected readonly contextMenuRenderer!: ContextMenuRenderer;
+    @inject(CommandService) protected readonly commandService!: CommandService;
 
     protected readonly updateRequestEmitter = new Emitter<void>();
     widgetWasUpdated = this.updateRequestEmitter.event;
@@ -97,6 +102,34 @@ export class TraceExplorerOpenedTracesWidget extends ReactWidget {
         await this.updateAvailableAnalysis(undefined);
     }
 
+    protected doHandleContextMenuEvent(event: React.MouseEvent<HTMLDivElement>, traceUUID: string): void {
+        this.doHandleOnExperimentSelected(event);
+        this.contextMenuRenderer.render({
+            menuPath: TraceExplorerMenus.PREFERENCE_EDITOR_CONTEXT_MENU,
+            anchor: { x: event.clientX, y: event.clientY },
+            args: [traceUUID]
+        });
+        event.preventDefault();
+        event.stopPropagation();
+    }
+
+    public openExperiment(traceUUID: string): void {
+        this.commandService.executeCommand(TraceViewerCommand.id, { traceUUID });
+    }
+
+    public closeExperiment(traceUUID: string): void {
+        signalManager().fireCloseTraceViewerTabSignal(traceUUID);
+    }
+
+    public deleteExperiment(traceUUID: string): void {
+        this.experimentManager.closeExperiment(traceUUID);
+        this.closeExperiment(traceUUID);
+    }
+
+    public getExperiment(traceUUID: string): Experiment | undefined {
+        return this._openedExperiments.find(experiment => experiment.UUID === traceUUID);
+    }
+
     render(): React.ReactNode {
         const totalHeight = this.getTotalHeight();
         this.forceUpdateKey = !this.forceUpdateKey;
@@ -136,6 +169,8 @@ export class TraceExplorerOpenedTracesWidget extends ReactWidget {
     protected doRenderExperimentRow(props: ListRowProps): React.ReactNode {
         const traceName = this._openedExperiments.length && props.index < this._openedExperiments.length
             ? this._openedExperiments[props.index].name : '';
+        const traceUUID = this._openedExperiments.length && props.index < this._openedExperiments.length
+            ? this._openedExperiments[props.index].UUID : '';
         let traceContainerClassName = 'trace-list-container';
         if (props.index === this._selectedExperimentIndex) {
             traceContainerClassName = traceContainerClassName + ' theia-mod-selected';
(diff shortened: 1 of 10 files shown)


muddana-satish:
Added a context menu to open, close and delete traces from trace explorer.

Kapture 2021-01-28 at 23 35 21

PatrickTasse (changes requested):
1) I am not sure if it's necessary to have "Close Trace" in the context menu. The user can just close the trace tab.

2) If you close the trace, close the browser, restart the trace-server, then reopen the browser, you get a state where the trace resources exist on the trace-server, but the trace is not opened in the trace-server. So the trace is not shown in the Trace Viewer and the user cannot delete the trace resources from the trace-server (unless the user chooses to open the same trace again).

3) If you restart the trace-server while a trace is opened in the browser, then the trace is closed in the trace-server. Opening an analysis for the trace in the browser opens a blank graph or table. Closing the trace tab in the browser and selecting "Open Trace" from the Trace Viewer opens the trace tab, but the Trace Viewer switches to "You have not yet opened a trace".

4) If you restart the trace-server while a trace is opened in the browser, then the trace is closed in the trace-server. If you select "Delete Trace" the delete operation will fail in the trace-server ("NOT FOUND") and the Trace Viewer switches to "You have not yet opened a trace". The trace resources are still present in the trace-server.

PatrickTasse:
I believe, if we agree to this implementation, that the problems 2, 3 and 4 above can be resolved if the trace-server:

- Includes all created experiments in the GET /experiments response, whether they are opened or not
- Accepts other APIs with expUUID for created experiments whether they are opened or not, opening them on-the-fly if necessary

mirsky-work:
> * Includes all created experiments in the GET /experiments response, whether they are opened or not
> * Accepts other APIs with expUUID for created experiments whether they are opened or not, opening them on-the-fly if necessary

I agree. In fact, it's not clear from the TSP documentation that GET /experiments returns only opened experiments today, and the whole notion of opened/closed experiment isn't part of the protocol, AFAICT.

PatrickTasse:
> > * Includes all created experiments in the GET /experiments response, whether they are opened or not
> > * Accepts other APIs with expUUID for created experiments whether they are opened or not, opening them on-the-fly if necessary
> 
> I agree. In fact, it's not clear from the TSP documentation that GET /experiments returns only opened experiments today, and the whole notion of opened/closed experiment isn't part of the protocol, AFAICT.

Then my only remaining concern is if we want to keep "Close Trace" in the context menu or not. If we do, then I am fine with this patch.

viewer-prototype/src/browser/trace-explorer/trace-explorer-commands.ts
```diff
@@ -0,0 +1,21 @@
+import { Command, MenuPath } from '@theia/core';
+
+export namespace TraceExplorerMenus {
+    export const PREFERENCE_EDITOR_CONTEXT_MENU: MenuPath = ['trace-explorer-opened-traces-context-menu'];
+}
+export namespace TraceExplorerCommands {
+    export const OPEN_TRACE: Command = {
+        id: 'trace-explorer:preferences.open-trace',
```
PatrickTasse, line 8: Is 'preferences' appropriate in these id's?

muddana-satish: Updated. Thanks.

theia-extensions/viewer-prototype/src/browser/trace-explorer/trace-explorer-sub-widgets/trace-explorer-opened-traces-widget.tsx
```diff
@@ -223,6 +259,9 @@ export class TraceExplorerOpenedTracesWidget extends ReactWidget {
 
     protected async doUpdateOpenedExperiments(): Promise<void> {
         this._openedExperiments = await this.experimentManager.getOpenedExperiments();
+        this._openedExperiments.forEach(experiment => {
+            this.experimentManager.addExperiment(experiment);
```
PatrickTasse, line 263: Open a trace, close the trace, close the browser. Restart the trace-server, then reopen the browser.

This code will add the created-but-closed experiment to the experiment-manager. But it does not add its traces to the trace-manager. So if you try to delete this experiment, its traces won't be be deleted on the trace-server because the trace-manager doesn't have them in its open traces map.

PatrickTasse (changes requested):

mirsky-work:
I'd like to propose some rephrasing of the context menu items. Since the 'Open Trace' action adds the experiment entry to the Opened Traces list, 'Close Trace' should be the name of the action that removes it from the list. 'Open Trace' is already used in other context to describe a slightly different action, so perhaps something that would emphasize that it's purely client operations:
- Open Trace -> Show Trace
- Close Trace -> Hide Trace
- Delete Trace -> Close Trace

Also, can we show the trace tab by double clicking on the Opened Traces list entry? Can be implemented in a separate patch.

PatrickTasse:
> I'd like to propose some rephrasing of the context menu items. Since the 'Open Trace' action adds the experiment entry to the Opened Traces list, 'Close Trace' should be the name of the action that removes it from the list. 'Open Trace' is already used in other context to describe a slightly different action, so perhaps something that would emphasize that it's purely client operations:
> 
> * Open Trace -> Show Trace
> * Close Trace -> Hide Trace
> * Delete Trace -> Close Trace

I would tend to disagree but I'm open to the discussion. For me, the terms Open and Close intuitively refer to the opening and closing of the trace tab. The fact that opening a trace adds it to the list is a side effect.

I have reservations about Delete though. It might be interpreted incorrectly that it means the original trace is deleted from the file system. It does not, this action only deletes the link to that original trace in the trace-server workspace, as well as deleting the supplementary files associated with it. It also closes the tab as a side effect. I would perhaps suggest the term Remove (ie. from the list and from the trace-server).

mirsky-work:
> I have reservations about Delete though. It might be interpreted incorrectly that it means the original trace is deleted from the file system. It does not, this action only deletes the link to that original trace in the trace-server workspace, as well as deleting the supplementary files associated with it. It also closes the tab as a side effect.

+1. I would think twice before performing an operation that contains the word 'Delete'.

mirsky-work:
> I would tend to disagree but I'm open to the discussion. For me, the terms Open and Close intuitively refer to the opening and closing of the trace tab. The fact that opening a trace adds it to the list is a side effect.

My point is that 'Open Trace' is already used elsewhere to describe a different operation. If 'Show Trace' isn't intuitive in your opinion, how about something like 'Open Trace Tab'?

bhufmann:
> > I have reservations about Delete though. It might be interpreted incorrectly that it means the original trace is deleted from the file system. It does not, this action only deletes the link to that original trace in the trace-server workspace, as well as deleting the supplementary files associated with it. It also closes the tab as a side effect.
> 
> +1. I would think twice before performing an operation that contains the word 'Delete'.

Maybe we should just call it 'Remove' instead 'Delete'. Delete could be misunderstood to delete the trace file resource.

bhufmann:
> > I would tend to disagree but I'm open to the discussion. For me, the terms Open and Close intuitively refer to the opening and closing of the trace tab. The fact that opening a trace adds it to the list is a side effect.
> 
> My point is that 'Open Trace' is already used elsewhere to describe a different operation. If 'Show Trace' isn't intuitive in your opinion, how about something like 'Open Trace Tab'?

Would be 'Reopen Trace' a better option?

PatrickTasse (approved):
Functionality looks good to me, unless we want to rename some of the menu items in this or another patch.

bhufmann (approved):
Looks good to me. I propose to change command names in a separate patch (if decided so).
Thanks!

PR #258, opened 2021-02-05
Fetch the style model only once

PatrickTasse:
Fetch the style model from the style provider when the component is
mounted, not every time the component is updated if the outputStatus is
not COMPLETED.


bhufmann (approved):
Looks good to me! Thanks

PR #259, opened 2021-02-05
Set cancelable flag in progress monitor for opening of traces

bhufmann:
This is necessary because the latest Theia version changed the default
behavior to align with VsCode.

Fixes #256


PatrickTasse (approved):
The Cancel button appears and can be used to dismiss the dialog.

The trace tab doesn't appear to close itself as a result, though.

PR #262, opened 2021-02-08
Fetch the tree only while running analysis and use throttling

PatrickTasse:
The tree should not be re-fetched when the component is updated, as this
creates an endless loop of fetching the tree while the analysis is
running. Use a single fetch tree to both update the model and monitor
the analysis state. Introduce a delay between fetch tree iterations.

Update the chart while the analysis is running and once more on analysis
completion.

Always check that the tree model is defined before trying to access it.


bhufmann (approved):
Looks good to me. Works well. Tested in GitPod

PR #266, opened 2021-02-13
XY Chart: Support time and time range selection

### LINKED ISSUE #121: XY Chart: support time and time range selection
timegraph chart supports selecting a time or time range, but not the XY chart. This support should be added

arfio:
When clicking on the XY chart, the user can select a time.
When clicking and dragging, the user can select a time range.

fixes #121 

### DIFF
--- a/packages/react-components/src/components/xy-output-component.tsx
+++ b/packages/react-components/src/components/xy-output-component.tsx
@@ -29,6 +29,25 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
     private colorMap: Map<string, number> = new Map();
 
     private lineChartRef: any;
+    private mouseIsDown = false;
+    private posPixelSelect = 0;
+
+    private updateSelection = (event: MouseEvent) => {
+        if (this.mouseIsDown && this.props.unitController.selectionRange) {
+            const xStartPos = this.props.unitController.selectionRange.start;
+            const scale = this.props.viewRange.getEnd() - this.props.viewRange.getstart();
+            this.props.unitController.selectionRange = {
+                start: xStartPos,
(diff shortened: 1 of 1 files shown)


tahini (changes requested):
Thanks for this feature! It's a great improvement and it works well! ... as long as the selection is in range... See the screencast below, when I move the time range so that the selection is out of sight, the xy chart is completely selected.

xySelection

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -154,28 +181,41 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
+                return;
+            }
+            const minValue = this.findNearestValue(min, ticks);
+            const minPixel = xScale.getPixelForValue(min, minValue);
+            const maxValue = this.findNearestValue(max, ticks);
+            let maxPixel = xScale.getPixelForValue(max, maxValue);
+            // In the case the selection is going out of bounds, the pixelValue needs to be in the displayed range.
+            if (maxPixel === 0) {
+                maxPixel = chart.chartArea.right;
+            }
```
tahini, line 197: Just asking, why do you have to change this method? The behavior of drawing the the selection in XY chart is correct in master. You added the possibility to also select from the XY chart. I'm just wondering why this needs to be modified too.

tahini (approved):
It works great! Thanks!

packages/react-components/src/components/xy-output-component.tsx
arfio: if your selection is partially out of range, the selection is not correctly displayed. Also the borders shown were not the same as for the timeline chart. When I corrected these issues, I had to account for other things as well, start > end and the window out of range completely.

bhufmann (approved):
Looks good to me! Thanks for the improvement

bhufmann (approved):

PR #273, opened 2021-02-22
Use plugin per chart instance instead of global for time selection

### LINKED ISSUE #269: Time selection not cleared when closing xy chart
The time selection in an xy-chart is not cleared when closing the chart. This can be observed when re-opening the xy-chart after the time selecting was change while the xy-chart was closed.

bug-xy-chart-selection

bhufmann:
Fixes #269

### DIFF
--- a/packages/react-components/src/components/xy-output-component.tsx
+++ b/packages/react-components/src/components/xy-output-component.tsx
@@ -31,7 +31,9 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
     private lineChartRef: any;
     private mouseIsDown = false;
     private posPixelSelect = 0;
-
+    private plugin = {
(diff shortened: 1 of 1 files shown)


PatrickTasse (changes requested):
When there is a selection range but no checked element in the XY tree, it appears that the XY chart is completely filled by a blue selection rectangle.

PatrickTasse:
> When there is a selection range but no checked element in the XY tree, it appears that the XY chart is completely filled by a blue selection rectangle.

Sorry, this behaviour exists before this patch.

bhufmann:
> > When there is a selection range but no checked element in the XY tree, it appears that the XY chart is completely filled by a blue selection rectangle.
> 
> Sorry, this behaviour exists before this patch.

No problem. I think this should be fixed in a separate patch.

PatrickTasse (approved):

MatthewKhouzam (approved):
I confirm, it fixes the bug.

Issue #161, opened 2020-10-13
Cursor location on graph is colour contrasting vertical bar


ebugden:
When mousing over a graph, an eye-catching bar can clearly communicate the user’s location:

![cursor-red-bar](https://user-images.githubusercontent.com/28311615/95927942-2740b000-0d8e-11eb-9206-f7bd0e5163f3.PNG)

PR #280, opened 2021-02-26
Make panel titles Horizontal

### LINKED ISSUE #154: Make panel titles horizontal
Horizontal text is the fastest and most comfortable to read (compared to vertical, diagonal, etc.) 

Position the ‘X’ for deleting the chart at the far right (keeps deletion intuitive by following the standard for horizontal windows in VS Code).

horizontal-panel-title

Other benefits:
* Horizontal titles can structure the layout by doubling as visual dividers between each panel
* Horizontal orientation leaves room for longer titles

Quick test you can try on yourself to show that horizontal text is significantly faster to read ([Article](https://jov.arvojournals.org/article.aspx?articleid=2121153)). 

horizontal-vertical-text-reading2

muddana-satish:
fixes #154

### DIFF
--- a/packages/react-components/src/components/abstract-output-component.tsx
+++ b/packages/react-components/src/components/abstract-output-component.tsx
@@ -43,7 +43,7 @@ export interface AbstractOutputState {
 export abstract class AbstractOutputComponent<P extends AbstractOutputProps, S extends AbstractOutputState> extends React.Component<P, S> {
 
     private mainAreaContainer: React.RefObject<HTMLDivElement>;
-    private readonly HANDLE_WIDTH = 30;
+    private readonly HANDLE_HEIGHT = 20;
 
     constructor(props: P) {
         super(props);
@@ -62,11 +62,11 @@ export abstract class AbstractOutputComponent<P extends AbstractOutputProps, S e
             onTouchStart={this.props.onTouchStart}
             onTouchEnd={this.props.onTouchEnd}
            >
-            <div className='widget-handle' style={{ width: this.HANDLE_WIDTH, height:this.props.style.height }}>
+            <div className='widget-handle' style={{ width: '100%', height: this.getHandleHeight()}}>
                 {this.renderTitleBar()}
             </div>
             <div className='main-output-container' ref={this.mainAreaContainer}
-            style={{ width: this.props.widthWPBugWorkaround - this.HANDLE_WIDTH, height:this.props.style.height }}>
+            style={{ width: this.props.widthWPBugWorkaround, height: parseInt(this.props.style.height.toString()) - this.getHandleHeight() }}>
                 {this.renderMainArea()}
             </div>
             {this.props.children}
@@ -76,12 +76,15 @@ export abstract class AbstractOutputComponent<P extends AbstractOutputProps, S e
     private renderTitleBar(): React.ReactNode {
         const outputName = this.props.outputDescriptor.name;
         return <React.Fragment>
-            <button className='remove-component-button' onClick={this.closeComponent}>
-                <FontAwesomeIcon icon={faTimes} />
-            </button>
             <div className='title-bar-label'>
                 {outputName}
             </div>
+            <div className='remove-button-panel'>
+                <button className='remove-component-button' onClick={this.closeComponent}>
+                    <FontAwesomeIcon icon={faTimes} />
+                </button>
+            </div>
+
         </React.Fragment>;
     }
 
@@ -96,8 +99,8 @@ export abstract class AbstractOutputComponent<P extends AbstractOutputProps, S e
         return 1000;
     }
 
-    public getHandleWidth(): number {
-        return this.HANDLE_WIDTH;
+    public getHandleHeight(): number {
+        return this.HANDLE_HEIGHT;
     }
 
     abstract renderMainArea(): React.ReactNode;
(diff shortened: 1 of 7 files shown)


muddana-satish:
Screenshot of the horizontal panel titles. Any suggestions @ebugden.

image

ebugden:
Thanks for this work Satish! There needs to be more visual emphasis on the panel titles so that your eye is drawn to it and it's clear that this is what separates each panel. Before merging, we need to at minimum:  
* **Make the titles bold** (emphasize)
* **Remove the bold from the settings section titles** (de-emphasize)

And maybe also:
* Put a little breathing room between the title and the filter box / panel content? The title looks pretty glued to the rest right now.
* Temporary measure: Make the title bars slightly thicker so that it's more clear where the panel separation is? 

**Important sidenote:** At the moment the whole visual hierarchy of the layout is pretty messed up (ex. filter box and chart backgrounds are too "loud" compared to how important they are) and this is making it difficult to implement horizontal titles that stand out clearly, but that's outside the scope of this issue.  

Here's what it could look like after the above changes:
image

Below is a good example where the title stands out immediately:
image

ebugden (changes requested):
See comment above for review.

muddana-satish:
> * **Make the titles bold** (emphasize)
Titles are now bold and also changed the text color to white with a bit of padding which leaves some space below the title.

> * **Remove the bold from the settings section titles** (de-emphasize)
Do you mean remove the bold from the table headers? Looks like it is added to differentiate between header and values. And the table-tree is present in only few charts. Do you still think I should remove the bold for the table headers? Please let me know.

Here is the updated screenshot.
image

MatthewKhouzam:
I think this is great already, I would like to discuss some pros and cons on Tiling in a separate thread. @ebugden do we have an issue open for that?

The selling point of TC is everything is time aligned, tiling seems like it breaks that promise.

MatthewKhouzam:
Also, and I am just asking here... we now have gaps between the views, does that make correlation between charts worse?

ebugden:
> **MatthewKhouzam:** we now have gaps between the views, does that make correlation between charts worse?

@bhufmann also raised this concern when #154 was proposed. I pasted my response [comment](https://github.com/theia-ide/theia-trace-extension/issues/154#issuecomment-709551522) below with minor edits.

## Spending vertical space on titles: More difficult to compare many graphs?

> **bhufmann:** Users often want to correlate different graphs by putting them under each other on the same time axis and have them close to each other.

My expectation is that we can still deliver on that desire for easily comparable and information rich layouts even with some of that space spent on horizontal titles. For example, I think being able to easily add, reorder, and collapse many data-focused panels with a minimal style will go a long way to helping users feel like they can "lay everything they want to compare out on the table".

I would also argue that a title (or some sort of data label) is one thing you always want to have clearly visible/legible (to avoid forgetting or mixing up the data you're looking at).

----

> **MatthewKhouzam:** I would like to discuss some pros and cons on Tiling in a separate thread.

By tiling do you mean what Grafana does with being able to arrange visualisations in different shapes and sizes? If, yes there's no issue for discussing that right now since a single vertical stack feels sufficient at the moment (and allowing users that much flexibility adds significant complexity on the UX side of things if you want to keep layouts feeling clean/elegant). 

image

ebugden:
@muddana-satish The new screenshot meets my requirements for closing #154! Thanks for the update.
* I especially enjoy that the white panel titles now feel associated with the trace name (tab label). Good information hierarchy! 

----

Outside the scope of this issue, the visual hierarchy issues in the layout make the improved legibility of horizontal titles less obvious. **I'm worried that if this is merged now that users won't enjoy the horizontal titles even if it's the better decision in the long term.** Should we merge after other visual hierarchy issues are resolved? @bhufmann @ssmagula @mirsky-work @MatthewKhouzam 

Visual hierarchy issues:
* **Panel titles and the data should pop the most** (#282 should help with this)
* **Stronger title/chart association:** Is the black title bar associated with the graph above or below? You can logically deduce the answer, but the title/chart clusters should be visually obvious without need for deduction (see comparison below with Grafana). 
   * In the trace viewer, if you squint your eyes, it's not clear where the panels start and end (it looks like they start/end with the chart area). This means there's going to be a delay in the brain's "clustering algorithm" and this can make interactions (ex. moving panels) feel clunky.

| Weak title association | Strong title association |
|-------------------|--------------------------|
| image             | image |

ebugden:
After some discussion, **horizontal titles shouldn't be merged right now, since it breaks the tool's current visualisation logic.**

@MatthewKhouzam highlighted that most charts currently share the single time axis at the top so creating spaces between the charts breaks the association between the time axis and the charts below. Tool breaking consequence: With larger spaces it looks like the time-based charts don't have a horizontal scale. 

Two thoughts came out of this discussion that I'll dump below for now, but will quickly move to separate issues since they're off topic for this PR.

----

## **1. The tool's visualisation logic is already broken** 
Right now, panels can be placed in any order. So, association with the time scale is broken as soon as a graph that doesn't have time as the x-axis is placed between a time-based graphs and the scale.

 image

This broken association could be greatly improved with a **unified cursor for the time-based charts** (extension of #161). 
 
| Mouse over time-based chart | Mouse over not time-based |
|--------------------------|--------------------------|
|  image |  image |
| Strong association with scale across all time-based charts. | No association with time scale. |

## 2. Could we discuss the logic behind a single time axis?
Until today, I didn't know that having a single axis for all time-based charts was a core principle for the new tool. So I'm only now considering this in my design suggestions. Since this is a foundational decision that has widespread implications for UX, I'd like to know the design criteria that lead to this decision.
* **Motivation for single time axis?** To maximize vertical space by avoiding repeating redundant UI elements? Any other reasons? 
* **How to integrate views that aren't time-based?** There is precedent for a single time axis when all data is time-based (see Google's Perfetto below), but because not all our views are time-based (and can be in mixed order with time-based views) this is a UX pain that needs careful consideration.

Example: Perfetto uses a single time axis
image

Issue #289, opened 2021-03-04
Duplicate "Open Trace" commands in command palette (CTRL+SHIFT+P)


bhufmann:
Typing CTRL+SHIFT+P and search for Open Trace, will find 2 Open Trace commands. It turns out, that one is to open the open trace dialog and the second one is the command used in the context sensitive menu for open traces in the trace viewer. 

Executing the latter from the command from palette fails. It actually cause an unwanted behaviour because the required trace UUID is not passed on to the open procedure and it will try to find traces from an unset URI.

![DuplicateOpenTraceCommand](https://user-images.githubusercontent.com/413424/110017486-41a30800-7cf4-11eb-9ac8-0c4ccd495141.png)

bhufmann:
@muddana-satish I think one solution would be to not list the commands that are context sensetive in the trace viewer into the global commands. @marechal-p do you have a suggestion to not show certain commands in the command palette in Theia? Or any other suggestions for that bug?

PR #291, opened 2021-03-05
menu: fix duplicate 'open trace' command

vince-fugnitto:
#### Description

Fixes: #289

The following commit fixes the issue where a duplicate `open trace` command was present in the command palette. Instead of registering the command with a label (which makes it show up in the command palette), the label is added to the menu only.

_Note_: only commands with `labels` will show up in the command palette.

Signed-off-by: vince-fugnitto <vincent.fugnitto@ericsson.com>


bhufmann (approved):
Looks good to me. Tested with GitHub. Thanks. Bernd

MatthewKhouzam (approved):

PR #298, opened 2021-03-10
Automate trace server start

ankusht-work:
This MR does the following:
- Remove trace server start/stop buttons from the status bar
- Add start/stop actions in the command palette
- If not started, launch the trace server whenever the user tries to open a trace
- If the port param under the preferences is modified then kill the
  server on the existing port first and then propagate the new value.


theia-extensions/viewer-prototype/src/common/trace-viewer-environment.ts
```diff
@@ -3,18 +3,28 @@ import { EnvVariablesServer } from '@theia/core/lib/common/env-variables';
...
 
         this.port = this.preferenceService.get(TRACE_PORT);
-        this.preferenceService.onPreferenceChanged(event => {
+        this.preferenceService.onPreferenceChanged(async event => {
             if (event.preferenceName === TRACE_PORT) {
+                try {
+                    await this.traceServerConfigService.stopTraceServer();
+                    this.messageService.info(`Trace server disconnected on port: ${this.port}.`);
+                } catch (e){
+                    // Do nothing
```
MatthewKhouzam, line 26: maybe log this?

theia-extensions/viewer-prototype/src/browser/trace-server-url-provider-frontend-impl.ts
```diff
@@ -13,15 +15,24 @@ export class TraceServerUrlProviderImpl implements TraceServerUrlProvider, Front
...
     ) {
         this.port = this.preferenceService.get(TRACE_PORT);
-        this.preferenceService.onPreferenceChanged(event => {
+        this.preferenceService.onPreferenceChanged(async event => {
             if (event.preferenceName === TRACE_PORT) {
+                try {
+                    await this.traceServerConfigService.stopTraceServer();
+                    this.messageService.info(`Server disconnected on port: ${this.port}.`);
+                } catch (e){
+                    // Do nothing
```
MatthewKhouzam, line 30: log this

MatthewKhouzam (commented):
I came here to log errors and approve patches. I guess we can do both. :)

theia-extensions/viewer-prototype/src/common/trace-viewer-environment.ts
ankusht-work: In case if the user modifies the port before starting the server, then we don't want to show any errors :)

MatthewKhouzam: So comment that :)

ankusht-work: Done :)

theia-extensions/viewer-prototype/src/browser/trace-server-url-provider-frontend-impl.ts
ankusht-work: Done

theia-extensions/viewer-prototype/src/browser/trace-viewer/trace-viewer-contribution.ts
```diff
@@ -43,6 +82,32 @@ export class TraceViewerContribution extends WidgetOpenHandler<TraceViewerWidget
...
+                text: ''
+            }).then(async progress => {
+                    progress.report({ message: 'Launching trace server... ', work: { done: 10, total: 100 } });
+                    const resolve = await this.traceServerConfigService.startTraceServer(this.path, this.port);
+                    if (resolve === 'success') {
+                        progress.report({ message: `Trace server started on port: ${this.port}.`, work: { done: 100, total: 100 } });
+                        progress.cancel();
+                        const widget = super.open(traceURI, traceUUID);
+                        return widget;
+                    }
```
bhufmann, line 104: in the case the start of the server is not successful, the still need to call progress.cancel() or done() otherwise the progress dialog stays open. Moreover, since the progress dialog is by default not cancelable, the dialog will stay open forever and the user can't close the dialog. I had this problem.

This also raises the question, if the progress monitor should be cancelable. At least the dialog should be closable. 

image

There was also no indication that the server start failed (maybe a error dialog or status message?). I only saw it in the debug console:
 ERROR Request startTraceServer failed with error: spawn ./trace-compass-server/tracecompass-server ENOENT Error: spawn ./trace-compass-server/tracecompass-server ENOENT
    at Process.ChildProcess._handle.onexit (internal/child_process.js:267:19)
    at onErrorNT (internal/child_process.js:469:16)
    at processTicksAndRejections (internal/process/task_queues.js:84:21)

bhufmann (changes requested):
I also noticed, that when the healhCheck() + start server takes a long time after pressing the "Open Trace" button or "Open with -> Trace Viewer" then there is no progress indication for the user that the open process is ongoing. On suggestion is for, to change the order for the button case: 1) open dialog box 2) healthCheck() 3) start server 4) finish open the trace.
Maybe have the healthCheck as a step in the progress dialog when launch the server.

theia-extensions/viewer-prototype/src/browser/trace-viewer/trace-viewer-contribution.ts
ankusht-work: Added progress.cancel() and toast messages in case there is an error starting the server.

The idea was not to open the dialog box if the server isn't started successfully. Yes, I have also noticed that the gethealth() API call takes some time to respond (especially when the trace server is down). To mitigate that I have added a spinner that says "Connecting to trace server" as an interim solution for now. In the future, I think we may need a better/faster solution to determine the state of the trace server.

bhufmann (approved):
Thanks for the updates. Looks good to me.

MatthewKhouzam:
I reaffirm my approval!

ebugden:
Some related questions for @ankusht-work or anyone who knows: 
* Do users first have to download and configure the server settings manually for this to work?
* What happens if the server crashes? Does it restart automatically? Do users have to restart it manually using the command?
* How do users know via the UI if the server has crashed?

PR #300, opened 2021-03-12
Fixed event_table component's font-family/font-size

### LINKED ISSUE #216: Avoid using many different fonts and styles in different widgets
In the following screenshot, there are at least 3 different fonts, in areas circled using different color:

different_fonts

There are also discrepancies in the colors used in the different widgets, for example the shade of blue used for selection in the Trace Explorer vs. the events table (not shown in the screenshot above).

ankusht-work:
fixes #216

-The ag-grid has default color for selected row which isn't consistent
with theia's font family and font size. This change will override those
default values.

### DIFF
--- a/packages/react-components/style/output-components-style.css
+++ b/packages/react-components/style/output-components-style.css
@@ -95,3 +95,17 @@ canvas {
     width: 180px;
     color: var(--theia-input-placeholder-foreground)
 }
+
+.ag-theme-balham{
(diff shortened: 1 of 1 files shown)


MatthewKhouzam (approved):
Yes!

bhufmann (approved):
It looks good for the events table. Thanks!

PR #301, opened 2021-03-15
Enable text selection in events table to allow copying

muddana-satish:


MatthewKhouzam (approved):
Works!

bhufmann (approved):
Looks good to me. Thanks!

PR #302, opened 2021-03-17
Fix console warnings

muddana-satish:


MatthewKhouzam:
How do I test this?

bhufmann:
> How do I test this?

@MatthewKhouzam you can see the warning in the Debug console view, without this patch.

root WARN ag-Grid: since v20, grid options enableColResize is gone. Instead set resizable=true on the column definition for the columns resizing are allowed on. To migrate from gridOption.enableColResize=true, set gridOptions.defaultColDef.resizable=true

bhufmann (approved):
Looks good to me. Thanks!

PR #305, opened 2021-03-18
Use trace range for time graph view boundary validation before a fetch

### LINKED ISSUE #304: TspDataProvider@getData() uses the first time graph entry for boundary check
TspDataProvider@getData() uses the first time graph entry for boundary check, which can lead that certain time range are not drawn when the first entry end time is smaller than other entries.

bhufmann:
fixes #304

### DIFF
--- a/packages/react-components/src/components/data-providers/tsp-data-provider.ts
+++ b/packages/react-components/src/components/data-providers/tsp-data-provider.ts
@@ -4,6 +4,7 @@ import { TimelineChart } from 'timeline-chart/lib/time-graph-model';
 import { QueryHelper } from 'tsp-typescript-client/lib/models/query/query-helper';
 import { OutputElementStyle } from 'tsp-typescript-client/lib/models/styles';
 import { Annotation, Type } from 'tsp-typescript-client/lib/models/annotation';
+import { TimeRange } from '@trace-viewer/base/lib/utils/time-range';
 
 export class TspDataProvider {
 
@@ -27,7 +28,8 @@ export class TspDataProvider {
         this.totalRange = 0;
     }
 
-    async getData(ids: number[], entries: TimeGraphEntry[], viewRange?: TimelineChart.TimeGraphRange, resolution?: number): Promise<TimelineChart.TimeGraphModel> {
+    async getData(ids: number[], entries: TimeGraphEntry[], totalTimeRange: TimeRange,
+            viewRange?: TimelineChart.TimeGraphRange, resolution?: number): Promise<TimelineChart.TimeGraphModel> {
         this.timeGraphEntries = [...entries];
         if (!this.timeGraphEntries.length) {
             return {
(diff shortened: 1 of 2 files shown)


packages/react-components/src/components/data-providers/tsp-data-provider.ts
```diff
@@ -39,11 +41,11 @@ export class TspDataProvider {
...
         }
 
-        this.totalRange = this.timeGraphEntries[0].end - this.timeGraphEntries[0].start; // 1332170682540133097 - starttime
+        this.totalRange = totalTimeRange.getEnd() - totalTimeRange.getstart();
         let fetchParameters = QueryHelper.selectionTimeQuery(QueryHelper.splitRangeIntoEqualParts(1332170682440133097, 1332170682540133097, 1120), ids);
         if (viewRange && resolution) {
-            const start = viewRange.start + this.timeGraphEntries[0].start;
-            const end = viewRange.end + this.timeGraphEntries[0].start;
+            const start = viewRange.start + totalTimeRange.getstart();
+            const end = viewRange.end + totalTimeRange.getEnd();
```
PatrickTasse, line 48: Should be totalTimeRange.getstart() also

packages/react-components/src/components/data-providers/tsp-data-provider.ts
```diff
@@ -39,11 +41,11 @@ export class TspDataProvider {
             };
         }
 
-        this.totalRange = this.timeGraphEntries[0].end - this.timeGraphEntries[0].start; // 1332170682540133097 - starttime
+        this.totalRange = totalTimeRange.getEnd() - totalTimeRange.getstart();
         let fetchParameters = QueryHelper.selectionTimeQuery(QueryHelper.splitRangeIntoEqualParts(1332170682440133097, 1332170682540133097, 1120), ids);
         if (viewRange && resolution) {
-            const start = viewRange.start + this.timeGraphEntries[0].start;
-            const end = viewRange.end + this.timeGraphEntries[0].start;
+            const start = viewRange.start + totalTimeRange.getstart();
```
PatrickTasse, line 47: I would suggest to swap the order in next two lines (totalTimeRange.getstart() + viewRange.start/end)

PatrickTasse (changes requested):

packages/react-components/src/components/data-providers/tsp-data-provider.ts
bhufmann: good catch

packages/react-components/src/components/data-providers/tsp-data-provider.ts
bhufmann: Will do

PatrickTasse (approved):

MatthewKhouzam (approved):
Makes sense, I tried with an experiment that had a year between both traces. :)

PR #306, opened 2021-03-19
Set Focus to the trace viewer tab on click of analysis

muddana-satish:

### DIFF
--- a/packages/base/src/signal-manager.ts
+++ b/packages/base/src/signal-manager.ts
@@ -1,4 +1,5 @@
 import { EventEmitter } from 'events';
+import { Experiment } from 'tsp-typescript-client/lib/models/experiment';
 
 export declare interface SignalManager {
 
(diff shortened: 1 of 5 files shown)


muddana-satish:
Focus is set to the correct widget once the output analysis is selected.

Kapture 2021-03-19 at 01 16 24

PatrickTasse:
It does what I believe was intended, which is bring to front the trace tab of the selected trace.

However I'm not sure that is what I would consider 'set focus', which would be, set the keyboard focus on the analysis graph (so that keyboard navigation works without clicking) and you would see a blue border around the graph.

In addition, I would suggest the following improvements that are in addition to this pull request:

- When clicking on a trace tab, the keyboard focus should be given to the first, or last selected, analysis graph/table in the tab.
- When clicking on an analysis title in the tab, the keyboard focus should be given to its graph/table
- When clicking on an analysis in the Available Analysis list, the trace tab should be scrolled to make that analysis visible, whether it is being opened for the first time or if it was already created previously. If the visible page is already full, opening a new analysis currently is only noticed by the scrollbar thumb position/size being updated, as the new graph is out of view.

PatrickTasse (approved):

bhufmann (approved):
Looks good to me. Tested with GitPod

PR #338, opened 2021-03-30
Show spinning wheel when analysis is running for xy and timeline charts

### LINKED ISSUE #284: Add analysis progress bar
Adding a progress bar in the analysis panel would help users know if analysis is proceeding normally. 
* Seeing how fast things are going would **help users notice unusual stalls, crashes, etc.** Could help reveal configuration issues?
* **Especially useful when processing large traces** that take some time to get through.

This is a specific case of #56

bhufmann:
Fixes #284 (partly)


bhufmann:
will re-upload

PR #343, opened 2021-04-15
Support Range selection synchronization between events table and timegraph chart

muddana-satish:

### DIFF
--- a/packages/react-components/src/components/table-output-component.tsx
+++ b/packages/react-components/src/components/table-output-component.tsx
@@ -2,11 +2,12 @@
 import { AbstractOutputComponent, AbstractOutputProps, AbstractOutputState } from './abstract-output-component';
 import * as React from 'react';
 import { AgGridReact } from 'ag-grid-react';
-import { ColDef, IDatasource, GridReadyEvent, CellClickedEvent, GridApi, ColumnApi } from 'ag-grid-community';
+import { ColDef, IDatasource, GridReadyEvent, CellClickedEvent, GridApi, ColumnApi, Column, RowNode } from 'ag-grid-community';
 import { QueryHelper } from 'tsp-typescript-client/lib/models/query/query-helper';
 import { cloneDeep } from 'lodash';
 import { signalManager } from '@trace-viewer/base/lib/signals/signal-manager';
 import { TimelineChart } from 'timeline-chart/lib/time-graph-model';
+import { CellKeyDownEvent } from 'ag-grid-community/dist/lib/events';
 
 type TableOuputState = AbstractOutputState & {
     tableColumns: ColDef[];
@@ -30,9 +31,16 @@ export class TableOutputComponent extends AbstractOutputComponent<TableOutputPro
     private components: any;
     private gridApi: GridApi | undefined = undefined;
     private columnApi: ColumnApi | undefined = undefined;
-    private pendingIndex: number | undefined = undefined;
-    private lastIndex = { timestamp: Number.MIN_VALUE, index: 0 };
+    private prevStartTimestamp: number = Number.MIN_VALUE;
+    private startTimestamp: number = Number.MAX_VALUE;
+    private endTimestamp: number = Number.MIN_VALUE;
+    private isIndexChanged = false;
     private columnsPacked = false;
+    private timestampCol: string | undefined = undefined;
+    private eventSignal = false;
+    private enableIndexSelection = true;
+    private selectStartIndex = -1;
+    private selectEndIndex = -1;
 
     static defaultProps: Partial<TableOutputProps> = {
         cacheBlockSize: 200,
@@ -58,6 +66,7 @@ export class TableOutputComponent extends AbstractOutputComponent<TableOutputPro
 
         this.onEventClick = this.onEventClick.bind(this);
         this.onModelUpdated = this.onModelUpdated.bind(this);
+        this.onKeyDown = this.onKeyDown.bind(this);
     }
 
     renderMainArea(): React.ReactNode {
@@ -74,9 +83,9 @@ export class TableOutputComponent extends AbstractOutputComponent<TableOutputPro
                 onGridReady={this.onGridReady}
                 components={this.components}
                 onCellClicked={this.onEventClick}
-                rowSelection='single'
+                rowSelection='multiple'
                 onModelUpdated={this.onModelUpdated}
-                enableCellTextSelection={true}
+                onCellKeyDown={this.onKeyDown}
             >
             </AgGridReact>
         </div>;
@@ -101,23 +110,140 @@ export class TableOutputComponent extends AbstractOutputComponent<TableOutputPro
 
     private onEventClick(event: CellClickedEvent) {
         const columns = event.columnApi.getAllColumns();
-        const tooltipObj: { [key: string]: string } = {};
+        const data = event.data;
+        const mouseEvent = event.event as MouseEvent;
+        const gridApi = event.api;
+        const rowIndex = event.rowIndex;
+        const itemPropsObj = this.fetchItemProperties(columns, data);
+        signalManager().fireTooltipSignal(itemPropsObj);
+
+        const currTimestamp = (this.timestampCol && data) ? data[this.timestampCol] : undefined;
+        this.enableIndexSelection = true;
+        if (mouseEvent.shiftKey) {
+            if (this.selectStartIndex === -1) {
+                this.selectStartIndex = rowIndex;
+                if (currTimestamp) {
+                    this.startTimestamp = Number(currTimestamp);
+                }
+            }
+            this.selectEndIndex = rowIndex;
+            if (currTimestamp) {
+                this.endTimestamp = Number(currTimestamp);
+            }
+            this.selectRows();
+        } else {
+            if (mouseEvent.ctrlKey || mouseEvent.metaKey) {
+                gridApi.deselectAll();
+                const rowNode = gridApi.getRowNode(String(rowIndex));
+                if (rowNode && rowNode.id) {
+                    rowNode.setSelected(true);
+                }
+            }
+            this.selectStartIndex = this.selectEndIndex = rowIndex;
+            if (currTimestamp) {
+                this.startTimestamp = this.endTimestamp = Number(currTimestamp);
+            }
+        }
+        this.handleRowSelectionChange();
+    }
+
+    private fetchItemProperties(columns: Column[], data: any) {
+        const itemPropsObj: { [key: string]: string } = {};
         columns.forEach(column => {
             const headerName = column.getColDef().headerName;
             const colField = column.getColDef().field;
-            if (headerName && colField && event.data[colField]) {
-                tooltipObj[headerName] = event.data[colField];
+            if (headerName && colField && data && data[colField]) {
+                itemPropsObj[headerName] = data[colField];
             }
         });
-        signalManager().fireTooltipSignal(tooltipObj);
-        const timestampHeader = columns.find(column => column.getColDef().headerName === 'Timestamp ns');
-        if (timestampHeader) {
-            const timestampCol = timestampHeader.getColDef().field;
-            if (timestampCol) {
-                const timestamp = event.data[timestampCol];
-                const payload = { 'timestamp': timestamp };
-                this.lastIndex = { timestamp: Math.trunc(Number(timestamp)), index: event.rowIndex };
-                signalManager().fireSelectionChangedSignal(payload);
+        return itemPropsObj;
+    }
+
+    private onKeyDown(event: CellKeyDownEvent) {
+        const gridApi = event.api;
+        const keyEvent = event.event as KeyboardEvent;
+        const rowIndex = event.rowIndex;
+        this.enableIndexSelection = true;
+        const currTimestamp = (this.timestampCol && event.data) ? event.data[this.timestampCol] : undefined;
+
+        if (gridApi) {
+            const currentRow = gridApi.getRowNode(String(rowIndex));
+
+            let isContiguous = true;
+            if (keyEvent.shiftKey) {
+                if (keyEvent.code === 'ArrowDown') {
+                    if (!currentRow.isSelected()) {
+                        gridApi.deselectAll();
+                        isContiguous = false;
+                    }
+
+                    const nextRow = gridApi.getRowNode(String(rowIndex + 1));
+                    if (isContiguous === false) {
+                        if (this.timestampCol && nextRow.data) {
+                            this.startTimestamp = this.endTimestamp = Number(nextRow.data[this.timestampCol]);
+                        }
+                        this.selectStartIndex = this.selectEndIndex = nextRow.rowIndex;
+                    } else {
+                        if (this.selectEndIndex < this.selectStartIndex) {
+                            if (currentRow && currentRow.id) {
+                                currentRow.setSelected(false);
+                            }
+                        } else {
+                            if (nextRow && nextRow.id) {
+                                nextRow.setSelected(true);
+                            }
+                        }
+                        this.selectEndIndex += 1;
+                        if (this.timestampCol && nextRow.data) {
+                            this.endTimestamp = Number(nextRow.data[this.timestampCol]);
+                        }
+                    }
+                    this.handleRowSelectionChange();
+                } else if (keyEvent.code === 'ArrowUp') {
+                    if (!currentRow.isSelected()) {
+                        gridApi.deselectAll();
+                        isContiguous = false;
+                    }
+                    const nextRow = gridApi.getRowNode(String(rowIndex - 1));
+
+                    if (isContiguous === false) {
+                        if (this.timestampCol && nextRow.data) {
+                            this.startTimestamp = this.endTimestamp = Number(nextRow.data[this.timestampCol]);
+                        }
+                        this.selectStartIndex = this.selectEndIndex = nextRow.rowIndex;
+                    } else {
+                        if (this.selectStartIndex < this.selectEndIndex) {
+                            if (currentRow && currentRow.id) {
+                                currentRow.setSelected(false);
+                            }
+                        } else {
+                            if (nextRow && nextRow.id) {
+                                nextRow.setSelected(true);
+                            }
+                        }
+                        this.selectEndIndex -= 1;
+                        if (this.timestampCol && nextRow.data) {
+                            this.endTimestamp = Number(nextRow.data[this.timestampCol]);
+                        }
+                    }
+                    this.handleRowSelectionChange();
+
+                } else if (keyEvent.code === 'Space') {
+                    this.selectEndIndex = rowIndex;
+                    if (currTimestamp) {
+                        this.endTimestamp = Number(currTimestamp);
+                    }
+                    this.selectRows();
+                    this.handleRowSelectionChange();
+                }
+            } else if (keyEvent.code === 'Space') {
+                gridApi.deselectAll();
+                if (currentRow && currentRow.id) {
+                    currentRow.setSelected(true);
+                }
+                this.selectStartIndex = this.selectEndIndex = rowIndex;
+                this.startTimestamp = this.endTimestamp = Number(currTimestamp);
+                this.handleRowSelectionChange();
             }
         }
     }
@@ -169,7 +295,6 @@ export class TableOutputComponent extends AbstractOutputComponent<TableOutputPro
                     const itemCopy = cloneDeep(item);
                     rowsThisPage[i] = itemCopy;
                 }
-
                 params.successCallback(rowsThisPage, this.props.nbEvents);
             }
         };
@@ -221,25 +346,58 @@ export class TableOutputComponent extends AbstractOutputComponent<TableOutputPro
         this.setState({
             tableColumns: this.columnArray
         });
+
+        if (this.columnApi) {
+            const columns = this.columnApi.getAllColumns();
+            const timestampHeader = columns.find(column => column.getColDef().headerName === 'Timestamp ns');
+            if (timestampHeader) {
+                this.timestampCol = timestampHeader.getColDef().field;
+            }
+        }
+    }
+
+    private handleRowSelectionChange() {
+        if (this.timestampCol) {
+            const startTimestamp = String(this.startTimestamp);
+            const endTimestamp = String(this.endTimestamp);
+            const payload = { startTimestamp, endTimestamp };
+            this.prevStartTimestamp = Number(startTimestamp);
+            this.eventSignal = true;
+            signalManager().fireSelectionChangedSignal(payload);
+        }
     }
 
     private async handleTimeSelectionChange(range?: TimelineChart.TimeGraphRange) {
         if (range) {
-            const start = Math.trunc(this.props.range.getstart() + range.start);
-            if (this.lastIndex.timestamp === start) {
+            if (this.eventSignal) {
+                this.eventSignal = false;
                 return;
             }
-            const index = await this.fetchTableIndex(start);
-            if (index && this.gridApi) {
-                this.lastIndex = { timestamp: start, index: index };
+
+            this.startTimestamp = Math.trunc(this.props.range.getstart() + range.start);
+            this.endTimestamp = Math.trunc(this.props.range.getstart() + range.end);
+
+            if (this.startTimestamp === this.endTimestamp || !this.timestampCol) {
+                this.enableIndexSelection = true;
+            } else {
+                this.enableIndexSelection = false;
+            }
+
+            if (this.gridApi && this.startTimestamp !== this.prevStartTimestamp) {
+                this.prevStartTimestamp = this.startTimestamp;
+                this.selectStartIndex = this.selectEndIndex = -1;
                 this.gridApi.deselectAll();
-                this.gridApi.ensureIndexVisible(index);
-                const node = this.gridApi.getDisplayedRowAtIndex(index);
-                if (node.id) {
-                    node.setSelected(true);
-                } else {
-                    this.pendingIndex = index;
+
+                const index = await this.fetchTableIndex(this.startTimestamp > this.endTimestamp ? this.startTimestamp+1 : this.startTimestamp);
+                if (index) {
+                    const startIndex = this.startTimestamp > this.endTimestamp ? index-1 : index;
+                    this.selectStartIndex = this.selectStartIndex === -1 ? startIndex : this.selectStartIndex;
+                    this.selectEndIndex = (this.enableIndexSelection && this.selectEndIndex === -1) ? startIndex : this.selectEndIndex;
+                    this.gridApi.ensureIndexVisible(this.selectStartIndex);
+                    this.selectRows();
                 }
+            } else {
+                this.selectRows();
             }
         }
     }
@@ -263,15 +421,31 @@ export class TableOutputComponent extends AbstractOutputComponent<TableOutputPro
     }
 
     private onModelUpdated = async () => {
-        if (this.pendingIndex && this.gridApi) {
-            if (this.gridApi.getSelectedNodes().length === 0) {
-                this.gridApi.getDisplayedRowAtIndex(this.pendingIndex).setSelected(true);
-            }
-            this.pendingIndex = undefined;
-        }
+        this.selectRows();
+
         if (this.columnArray.length > 0 && !this.columnsPacked && this.columnApi) {
             this.columnApi.autoSizeAllColumns();
             this.columnsPacked = true;
         }
     };
+
+    private isValidRowSelection(rowNode: RowNode): boolean {
+        if ((this.enableIndexSelection && this.selectStartIndex !== -1 && this.selectEndIndex !== -1 && rowNode.rowIndex >= Math.min(this.selectStartIndex, this.selectEndIndex)
+            && rowNode.rowIndex <= Math.max(this.selectStartIndex, this.selectEndIndex)) || (!this.enableIndexSelection
+                && this.timestampCol && rowNode.data[this.timestampCol] >= Math.min(this.startTimestamp, this.endTimestamp)
+                && rowNode.data[this.timestampCol] <= Math.max(this.startTimestamp, this.endTimestamp))) {
+            return true;
+        }
+        return false;
+    }
+
+    private selectRows(): void {
+        if (this.gridApi) {
+            this.gridApi.forEachNode(rowNode => {
+                if (rowNode.id) {
+                    rowNode.setSelected(this.isValidRowSelection(rowNode));
+                }
+            });
+        }
+    }
 }
--- a/packages/react-components/src/components/timegraph-output-component.tsx
+++ b/packages/react-components/src/components/timegraph-output-component.tsx
@@ -50,9 +50,9 @@ export class TimegraphOutputComponent extends AbstractTreeOutputComponent<Timegr
 
     private selectedElement: TimeGraphStateComponent | undefined;
     private tooltipElement: TimeGraphStateComponent | undefined;
-    private tooltipInfo: {[key: string]: string} | undefined;
+    private tooltipInfo: { [key: string]: string } | undefined;
 
-    private onSelectionChanged = ( payload: { [key: string]: string; } ) => this.doHandleSelectionChangedSigna(payload);
+    private onSelectionChanged = (payload: { [key: string]: string; }) => this.doHandleSelectionChangedSignal(payload);
 
     constructor(props: TimegraphOutputProps) {
         super(props);
@@ -138,10 +138,10 @@ export class TimegraphOutputComponent extends AbstractTreeOutputComponent<Timegr
                 const columns = [];
                 if (headers && headers.length > 0) {
                     headers.forEach(header => {
-                        columns.push({title: header.name, sortable: true, tooltip: header.tooltip});
+                        columns.push({ title: header.name, sortable: true, tooltip: header.tooltip });
                     });
                 } else {
-                    columns.push({title: 'Name', sortable: true});
+                    columns.push({ title: 'Name', sortable: true });
                 }
                 this.setState({
                     outputStatus: treeResponse.status,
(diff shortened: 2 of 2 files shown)


muddana-satish:
Kapture 2021-04-15 at 01 21 33

bhufmann:
Thanks Satish, I'm going to review it.

packages/react-components/src/components/table-output-component.tsx
```diff
@@ -252,12 +272,16 @@ export class TableOutputComponent extends AbstractOutputComponent<TableOutputPro
...
     private onModelUpdated = async () => {
-        if (this.pendingIndex && this.gridApi) {
-            if (this.gridApi.getSelectedNodes().length === 0) {
-                this.gridApi.getDisplayedRowAtIndex(this.pendingIndex).setSelected(true);
+        this.pendingIndices.forEach(index => {
+            if (this.gridApi) {
+                const node = this.gridApi.getDisplayedRowAtIndex(index);
+                if (node.id) {
+                    node.setSelected(true);
+                    this.gridApi.ensureIndexVisible(index);
```
PatrickTasse, line 280: You only need to ensure that the endIndex is visible. This should be the last (most recent) selected time, e.g with Shift+right it is the last selected row, with Shift-left it is the first selected row.

packages/react-components/src/components/table-output-component.tsx
```diff
@@ -252,12 +272,16 @@ export class TableOutputComponent extends AbstractOutputComponent<TableOutputPro
...
+                }
             }
-            this.pendingIndex = undefined;
-        }
+        });
+        this.pendingIndices = [];
```
PatrickTasse, line 284: Is it sure that all pending indices have been loaded? Otherwise you need to only remove those that were found.

packages/react-components/src/components/table-output-component.tsx
```diff
@@ -216,20 +226,30 @@ export class TableOutputComponent extends AbstractOutputComponent<TableOutputPro
...
+                endIndex: endIndex
+            };
             this.gridApi.deselectAll();
-            this.gridApi.ensureIndexVisible(index);
-            const node = this.gridApi.getDisplayedRowAtIndex(index);
-            if (node.id) {
-                node.setSelected(true);
-            } else {
-                this.pendingIndex = index;
+            for (let index = startIndex; index <= endIndex; index++) {
```
PatrickTasse, line 243: This does not work when the selection is right-to-left (e.g. Shift+left) and endIndex < startIndex.

PatrickTasse (changes requested):

packages/react-components/src/components/table-output-component.tsx
```diff
@@ -107,7 +112,12 @@ export class TableOutputComponent extends AbstractOutputComponent<TableOutputPro
             if (timestampCol) {
                 const timestamp = event.data[timestampCol];
                 const payload = { 'timestamp': timestamp };
-                this.lastIndex = { timestamp: Math.trunc(Number(timestamp)), index: event.rowIndex };
+                this.lastIndex = {
+                    startTimestamp: Math.trunc(Number(timestamp)),
+                    endTimestamp: Math.trunc((Number(timestamp))),
+                    startIndex: event.rowIndex,
+                    endIndex: event.rowIndex
+                };
```
bhufmann, line 120: right now only the start timestamp is broadcasted below (fireSelectionChangedSignal). If user selects multiple rows, the you need to broadcast the start and end time of the range so that other views (e.g. timeline-chart) can draw the selection change)

packages/react-components/src/components/table-output-component.tsx
bhufmann, line 280: I don't think we need to ensureIndexVisible for an index that is not the start time. This means that it will show the last index and not the start index, right? Or do I miss something.

bhufmann (changes requested):
Some feedback from my side.  @PatrickTasse will review since he is original author of the single time sync.

packages/react-components/src/components/table-output-component.tsx
```diff
@@ -30,9 +30,17 @@ export class TableOutputComponent extends AbstractOutputComponent<TableOutputPro
     private components: any;
     private gridApi: GridApi | undefined = undefined;
     private columnApi: ColumnApi | undefined = undefined;
-    private pendingIndex: number | undefined = undefined;
-    private lastIndex = { timestamp: Number.MIN_VALUE, index: 0 };
+    private lastSelectedRange = {
+        startTimestamp: Number.MIN_VALUE,
+        endTimestamp: Number.MIN_VALUE,
+        startIndex: 0,
```
PatrickTasse, line 36: We don't seem to be using the startIndex/endIndex anywhere?

packages/react-components/src/components/table-output-component.tsx
```diff
@@ -212,23 +221,64 @@ export class TableOutputComponent extends AbstractOutputComponent<TableOutputPro
...
+                    this.startVisibleIndex = undefined;
+                }
             } else {
-                this.pendingIndex = index;
+                this.gridApi.forEachNode(rowNode => {
+                    if (rowNode.data && this.timestampCol && rowNode.data[this.timestampCol] >= this.startTimeStamp && rowNode.data[this.timestampCol] <= this.endTimeStamp) {
+                        rowNode.setSelected(true);
+                    }
+                });
+                if (this.startVisibleIndex) {
```
PatrickTasse, line 278: I think the ensureVisibleIndex() handling can be done only once in handleTimeSelectionChange, is this correct?

packages/react-components/src/components/table-output-component.tsx
```diff
@@ -212,23 +221,64 @@ export class TableOutputComponent extends AbstractOutputComponent<TableOutputPro
...
-                node.setSelected(true);
+            this.startVisibleIndex = startIndex;
+            this.selectRows();
+        }
+    }
+
+    private selectRows(): void {
+        if (this.gridApi) {
+
+            if (this.startTimeStamp === this.endTimeStamp && this.startVisibleIndex) {
```
PatrickTasse, line 265: Interesting side effect of this code:
If you click between (the timestamps of) events A and B on the time graph, B is selected in the table (next timestamp). But if you select a small time range that is completely between A and B, nothings gets selected.
This may be what we want, I'm not sure.

packages/react-components/src/components/table-output-component.tsx
```diff
@@ -212,23 +221,64 @@ export class TableOutputComponent extends AbstractOutputComponent<TableOutputPro
...
+                const node = this.gridApi.getDisplayedRowAtIndex(this.startVisibleIndex);
+                if (node.id) {
+                    node.setSelected(true);
+                    this.gridApi.ensureIndexVisible(this.startVisibleIndex);
+                    this.startVisibleIndex = undefined;
+                }
             } else {
-                this.pendingIndex = index;
+                this.gridApi.forEachNode(rowNode => {
+                    if (rowNode.data && this.timestampCol && rowNode.data[this.timestampCol] >= this.startTimeStamp && rowNode.data[this.timestampCol] <= this.endTimeStamp) {
```
PatrickTasse, line 274: I'm a bit worried that we are dependent on a "Timestamp ns" column for this to work properly, but it's OK until we find a better solution...

I was able to get a non-contiguous table selection, by using a text trace that had some events out-of-order. Wouldn't have noticed this is we did the selection from startIndex/endIndex...

packages/react-components/src/components/table-output-component.tsx
```diff
@@ -212,23 +221,64 @@ export class TableOutputComponent extends AbstractOutputComponent<TableOutputPro
...
-        if (this.lastIndex.timestamp === start) {
+        const end = Math.trunc(this.props.range.getstart() + range.end);
+
+        this.startTimeStamp = Math.min(start, end);
+        this.endTimeStamp = Math.max(start, end);
+        const startIndex = await this.fetchTableIndex(start);
+        const endIndex = await this.fetchTableIndex(end);
+
+        if (this.lastSelectedRange.startTimestamp === start &&
+            this.lastSelectedRange.endTimestamp === end &&
```
PatrickTasse, line 244: In the old code, we only cared about start time. When you drag a selection, you receive many callbacks but only the end time changes, the start time is the same. The old code would return early.

In the new code, all those callbacks make you go to the trace-server twice, once for startIndex and once for endIndex. This can overload the server, which sometimes you can see the table takes a few seconds to be updated, and sometimes the table bounces around processing all those pending server responses coming back.

In addition, it doesn't seem that the endIndex is used at all, since you select rows by their timestamp.

If in a subsequent patch we make it select the endIndex when extending a selection, we should find a way to throttle that or make it happen when the selection is finished (mouse released).

PatrickTasse (changes requested):

packages/react-components/src/components/table-output-component.tsx
muddana-satish: I agree we may not need the endIndex at all for now since I am using the timestamp for range selection. I removed it. Thanks for the suggestion.

packages/react-components/src/components/table-output-component.tsx
muddana-satish: Modified. Thanks.

packages/react-components/src/components/table-output-component.tsx
muddana-satish: updated.

packages/react-components/src/components/table-output-component.tsx
```diff
@@ -212,23 +219,59 @@ export class TableOutputComponent extends AbstractOutputComponent<TableOutputPro
...
     private async handleTimeSelectionChange(range: TimelineChart.TimeGraphRange) {
         const start = Math.trunc(this.props.range.getstart() + range.start);
-        if (this.lastIndex.timestamp === start) {
+        const end = Math.trunc(this.props.range.getstart() + range.end);
+
+        this.startTimeStamp = Math.min(start, end);
+        this.endTimeStamp = Math.max(start, end);
+        const startIndex = await this.fetchTableIndex(start);
+
+        if (this.lastSelectedRange.startTimestamp === start &&
```
PatrickTasse, line 240: We should do it a bit differently: If and only if the startTimestamp has changed, do the fetchTableIndex and ensure it's visible.

Then regardless of the above, always update the selected rows.

This will avoid many requests to the server while dragging a selection.

packages/react-components/src/components/table-output-component.tsx
```diff
@@ -212,23 +219,59 @@ export class TableOutputComponent extends AbstractOutputComponent<TableOutputPro
...
+            this.lastSelectedRange.endTimestamp === end &&
+            startIndex === startIndex) {
```
PatrickTasse, line 242: Good luck getting this to be false ;)

packages/react-components/src/components/table-output-component.tsx
```diff
@@ -212,23 +219,59 @@ export class TableOutputComponent extends AbstractOutputComponent<TableOutputPro
...
-                node.setSelected(true);
+            this.pendingIndex = startIndex;
+            this.gridApi.ensureIndexVisible(startIndex);
+            this.selectRows();
+        }
+    }
+
+    private selectRows(): void {
+        if (this.gridApi) {
+            if (this.startTimeStamp === this.endTimeStamp && this.pendingIndex) {
```
PatrickTasse, line 261: When you click or shift-click on the event table, the selection made by the user is not recorded. However it triggers a handleTimeSelectionChange event. There, the startTimeStamp and endTimeStamp is recorded. The pending index is undefined after the row has been selected.

If you then scroll out of range to trigger a model update, and then scroll back, you will see that the selection is now only based on timestamps, and it might have changed from the user selection (this only happens when there are multiple events on the same start/end timestamp).

This means that the selection start/end index should be stored and remembered, at least when it is set by the user clicking on the table or selecting events with (Shift+)arrow up/down.

PatrickTasse (changes requested):

packages/react-components/src/components/table-output-component.tsx
muddana-satish: Updated the code to only fetch the index from the server when the previous start time is different than the current start time.

muddana-satish:
@PatrickTasse I have covered the scenarios you mentioned. I only update the rowSelection if the signal is from outside. The changes now holds good for the two way synchronization. Could you please review the latest changes. Thanks.

packages/react-components/src/components/table-output-component.tsx
```diff
@@ -74,9 +81,9 @@ export class TableOutputComponent extends AbstractOutputComponent<TableOutputPro
                 onGridReady={this.onGridReady}
                 components={this.components}
                 onCellClicked={this.onEventClick}
-                rowSelection='single'
+                rowSelection='multiple'
```
PatrickTasse, line 84: Two issues with table selection:

- When pressing space bar it adds the row to current selection, same as Ctrl+click. This might be the default 'multiple' behaviour. We might want to only allow contiguous selection. In Eclipse TC, Ctrl+click replaces the selection, maybe should be the same, also for space bar.
- When using the scroll bar to go far enough that model is updated, and the current selection index is out of range of the current virtual window, then it is not possible to extend selection with Shift+click. You need to start a new selection with normal click first before you can extend it again.

PatrickTasse (changes requested):

packages/react-components/src/components/table-output-component.tsx
muddana-satish: I have made changes for the first scenario. Both space and ctrl+click now has the right behavior. They only select a single node and deselects all the previously selected nodes. 

Should we handle the second scenario explicitly? The grid is supposed to handle it by default.

muddana-satish: I fixed both the issues now. could you please review the latest. Thanks.

packages/react-components/src/components/table-output-component.tsx
```diff
@@ -101,23 +111,131 @@ export class TableOutputComponent extends AbstractOutputComponent<TableOutputPro
...
+        const keyEvent = event.event as KeyboardEvent;
+        const rowIndex = event.rowIndex;
+        this.enableIndexSelection = keyEvent.shiftKey;
+
+        if (gridApi && this.timestampCol) {
+            const currentRow = gridApi.getRowNode(String(rowIndex));
+            const currTimestamp = event.data[this.timestampCol];
+
+            let isContiguous = true;
+            if (keyEvent.shiftKey) {
```
PatrickTasse, line 165: I would expect, if you first select row 5, then shift-up to row 3, then shift-down to row 7, that rows 3,4 would de-select and final selection would be [5,7]. So it should remember row 5 as selection start.

packages/react-components/src/components/table-output-component.tsx
```diff
@@ -101,23 +111,131 @@ export class TableOutputComponent extends AbstractOutputComponent<TableOutputPro
...
+                        gridApi.deselectAll();
+                        if (currentRow && currentRow.id) {
+                            currentRow.setSelected(true);
+                        }
+                    }
+                    this.selectRowsByIndex();
+                    this.handleRowSelectionChange();
+                }
+            } else if (keyEvent.code === 'Space') {
+                gridApi.deselectAll();
```
PatrickTasse, line 216: There are a few places like here where there is code that you would expect to execute (related to index selection) even if the timestampCol is not defined. We should be careful to only make timestamp handling conditional on that variable.

PatrickTasse (changes requested):

PatrickTasse (approved):
Thanks for the contribution!

bhufmann (changes requested):
It works until doing multiple quick selection changes.

I get all kinds of TypeError exception with AgGrid when doing several range selection changes from a timeline-chart and scroll in the table. After having one of the exception the Table is  not functioning anymore and doesn't update when scrolling or getting range updates. When closing the table also the timeline-chart closes at the same time. It looks like there are issues with concurrent updates. See below of a copy of the exceptions (Note that they don't happen in this order). I think there needs to be some updates to avoid such problems.

Please update the following components: AgGridReact
root ERROR The above error occurred in the <AgGridReact> component:
    in AgGridReact (created by TableOutputComponent)
    in div (created by TableOutputComponent)
    in div (created by TableOutputComponent)
    in div (created by TableOutputComponent)
    in TableOutputComponent (created by TraceContextComponent)
    in Resizable (created by GridItem)
    in DraggableCore (created by GridItem)
    in GridItem (created by ReactGridLayout)
    in div (created by ReactGridLayout)
    in ReactGridLayout (created by ResponsiveReactGridLayout)
    in ResponsiveReactGridLayout (created by WidthProvider)
    in WidthProvider (created by TraceContextComponent)
    in div (created by TraceContextComponent)
    in TraceContextComponent
    in div

Consider adding an error boundary to your tree to customize error handling behavior.
Visit https://fb.me/react-error-boundaries to learn more about error boundaries.
root ERROR TypeError: Cannot read property 'getBlockNumber' of undefined
    at InfiniteCache.push.../../node_modules/ag-grid-community/dist/lib/rowModels/cache/rowNodeCache.js.RowNodeCache.destroyBlock (http://localhost:3000/43.bundle.js:102192:34)
    at http://localhost:3000/43.bundle.js:102026:66
    at http://localhost:3000/43.bundle.js:102174:13
    at Array.forEach (<anonymous>)
    at InfiniteCache.push.../../node_modules/ag-grid-community/dist/lib/rowModels/cache/rowNodeCache.js.RowNodeCache.forEachBlockId (http://localhost:3000/43.bundle.js:102172:13)
    at InfiniteCache.push.../../node_modules/ag-grid-community/dist/lib/rowModels/cache/rowNodeCache.js.RowNodeCache.forEachBlockInOrder (http://localhost:3000/43.bundle.js:102164:14)
    at InfiniteCache.push.../../node_modules/ag-grid-community/dist/lib/rowModels/cache/rowNodeCache.js.RowNodeCache.destroy (http://localhost:3000/43.bundle.js:102026:14)
    at InfiniteRowModel.push.../../node_modules/ag-grid-community/dist/lib/rowModels/infinite/infiniteRowModel.js.InfiniteRowModel.destroyCache (http://localhost:3000/43.bundle.js:104522:32)
    at http://localhost:3000/43.bundle.js:104358:56
    at http://localhost:3000/43.bundle.js:77870:64
        

rowNodeCache.js:93 Uncaught (in promise) TypeError: Cannot read property 'getState' of undefined
    at rowNodeCache.js:93
    at Array.forEach (<anonymous>)
    at InfiniteCache.push.../../node_modules/ag-grid-community/dist/lib/rowModels/cache/rowNodeCache.js.RowNodeCache.purgeBlocksIfNeeded (rowNodeCache.js:92)
    at InfiniteCache.push.../../node_modules/ag-grid-community/dist/lib/rowModels/cache/rowNodeCache.js.RowNodeCache.postCreateBlock (rowNodeCache.js:112)
    at InfiniteCache.push.../../node_modules/ag-grid-community/dist/lib/rowModels/infinite/infiniteCache.js.InfiniteCache.createBlock (infiniteCache.js:138)
    at InfiniteCache.push.../../node_modules/ag-grid-community/dist/lib/rowModels/infinite/infiniteCache.js.InfiniteCache.getRow (infiniteCache.js:130)
    at InfiniteRowModel.push.../../node_modules/ag-grid-community/dist/lib/rowModels/infinite/infiniteRowModel.js.InfiniteRowModel.getRow (infiniteRowModel.js:245)
    at GridApi.push.../../node_modules/ag-grid-community/dist/lib/gridApi.js.GridApi.getDisplayedRowAtIndex (gridApi.js:988)
    at TableOutputComponent.push.../../packages/react-components/lib/components/table-output-component.js.TableOutputComponent.selectRowsByTimestamp (table-output-component.tsx:427)
    at TableOutputComponent.<anonymous> (table-output-component.tsx:391)

rowNodeCache.js:172 Uncaught (in promise) TypeError: Cannot read property 'forEachNodeDeep' of undefined
    at rowNodeCache.js:172
    at rowNodeCache.js:187
    at Array.forEach (<anonymous>)
    at InfiniteCache.push.../../node_modules/ag-grid-community/dist/lib/rowModels/cache/rowNodeCache.js.RowNodeCache.forEachBlockId (rowNodeCache.js:185)
    at InfiniteCache.push.../../node_modules/ag-grid-community/dist/lib/rowModels/cache/rowNodeCache.js.RowNodeCache.forEachBlockInOrder (rowNodeCache.js:177)
    at InfiniteCache.push.../../node_modules/ag-grid-community/dist/lib/rowModels/cache/rowNodeCache.js.RowNodeCache.forEachNodeDeep (rowNodeCache.js:171)
    at InfiniteRowModel.push.../../node_modules/ag-grid-community/dist/lib/rowModels/infinite/infiniteRowModel.js.InfiniteRowModel.forEachNode (infiniteRowModel.js:258)
    at GridApi.push.../../node_modules/ag-grid-community/dist/lib/gridApi.js.GridApi.forEachNode (gridApi.js:517)
    at TableOutputComponent.push.../../packages/react-components/lib/components/table-output-component.js.TableOutputComponent.selectRowsByTimestamp (table-output-component.tsx:436)
    at TableOutputComponent.<anonymous> (table-output-component.tsx:391)    
    
8rowRenderer.js:279 Uncaught Error: ag-Grid: cannot get grid to draw rows when it is in the middle of drawing rows. Your code probably called a grid API method while the grid was in the render stage. To overcome this, put the API call into a timeout, eg instead of api.refreshView(), call setTimeout(function(){api.refreshView(),0}). To see what part of your code that caused the refresh check this stacktrace.
    at RowRenderer.push.../../node_modules/ag-grid-community/dist/lib/rendering/rowRenderer.js.RowRenderer.getLockOnRefresh (rowRenderer.js:279)
    at RowRenderer.push.../../node_modules/ag-grid-community/dist/lib/rendering/rowRenderer.js.RowRenderer.redrawAfterScroll (rowRenderer.js:479)
    at eventService.js:133
    at Array.forEach (<anonymous>)
    at Function.push.../../node_modules/ag-grid-community/dist/lib/utils.js.Utils.forEachSnapshotFirst (utils.js:1219)
    at EventService.push.../../node_modules/ag-grid-community/dist/lib/eventService.js.EventService.dispatchToListeners (eventService.js:128)
    at EventService.push.../../node_modules/ag-grid-community/dist/lib/eventService.js.EventService.dispatchEvent (eventService.js:105)
    at GridPanel.push.../../node_modules/ag-grid-community/dist/lib/gridPanel/gridPanel.js.GridPanel.redrawRowsAfterScroll (gridPanel.js:1102)
    at GridPanel.push.../../node_modules/ag-grid-community/dist/lib/gridPanel/gridPanel.js.GridPanel.onVerticalScroll (gridPanel.js:1039)

packages/react-components/src/components/table-output-component.tsx
```diff
@@ -101,23 +111,134 @@ export class TableOutputComponent extends AbstractOutputComponent<TableOutputPro
...
+    }
+
+    private fetchItemProperties(columns: Column[], data: any) {
+        const itemPropsObj: { [key: string]: string } = {};
         columns.forEach(column => {
             const headerName = column.getColDef().headerName;
             const colField = column.getColDef().field;
-            if (headerName && colField && event.data[colField]) {
-                tooltipObj[headerName] = event.data[colField];
+            if (headerName && colField && data[colField]) {
```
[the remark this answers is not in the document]
bhufmann: @muddana-satish could you please look into these issues?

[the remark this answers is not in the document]
muddana-satish: Thanks @bhufmann for catching this. The data field can be undefined until the data is loaded into the row. I have made changes to fix this error.

packages/react-components/src/components/table-output-component.tsx
```diff
@@ -263,15 +411,63 @@ export class TableOutputComponent extends AbstractOutputComponent<TableOutputPro
...
+                                this.selectEndIndex = Math.max(this.selectEndIndex, rowNode.rowIndex);
+                            }
+                        } else {
+                            rowNode.setSelected(false);
+                        }
+                    }
+                });
+
+                if (this.gridApi.getSelectedNodes.length === 0) {
+                    this.selectStartIndex = Number.MAX_VALUE;
```
PatrickTasse, line 452: So this line can cause an exception when selecting by index.

Here is a log showing the error:

user starts a new single selection by clicking in time graph

root INFO handleTimeSelectionChange() startTimestamp=4699075186 endTimestamp=4699075186  >>> single selection
root INFO handleTimeSelectionChange fetchTableIndex...  >>> this now blocked in an await, UI continues handling other events

user extends the selection by dragging the mouse to the right

root INFO handleTimeSelectionChange() startTimestamp=4699075186 endTimestamp=4699088315  >>> selection is extended, start time does not change
root INFO handleTimeSelectionChange calling selectRowsByTimestamp()
root INFO selectRowsByTimestamp enableTimestampSelection=true startTimestamp=4699075186 endTimestamp=4699088315 selectStartIndex=1.7976931348623157e+308
root INFO selectRowsByTimestamp selectStartIndex=Number.MAX_VALUE  >>> because selectedNodes.length=0

the above happens a few times as mouse is extending selection with new end time

now the trace-server comes back with the start index, the await returns

root INFO handleTimeSelectionChange fetchTableIndex => returned 1238
root INFO handleTimeSelectionChange gridApi.ensureIndexVisible(1238)  >>> this triggers a fetching of new model lines from trace-server in the background
root INFO handleTimeSelectionChange fetchTableIndex selectStartIndex=1238
root INFO handleTimeSelectionChange calling selectRowsByTimestamp()
root INFO selectRowsByTimestamp enableTimestampSelection=true startTimestamp=4699075186 endTimestamp=4699170844 selectStartIndex=1238  <<< when we fetched start index it was a single selection, not anymore
root INFO selectRowsByTimestamp selectStartIndex=Number.MAX_VALUE  >>> because selectedNodes.length=0

now user starts a new single selection by clicking in time graph

root INFO handleTimeSelectionChange() startTimestamp=4699470950 endTimestamp=4699470950  >>> single selection
root INFO handleTimeSelectionChange fetchTableIndex...  >>> this now blocked in an await, UI continues handling other events

now the trace-server comes back with the lines from start index 1238 and updates the model

root INFO onModelUpdated calling selectRowsByTimestamp()
root INFO selectRowsByTimestamp enableTimestampSelection=true startTimestamp=4699470950 endTimestamp=4699470950 selectStartIndex=1.7976931348623157e+308
root INFO selectRowsByTimestamp getDisplayedRowAtIndex(1.7976931348623157e+308)  <<< *** EXCEPTION IS THROWN ***

PatrickTasse (changes requested):

packages/react-components/src/components/table-output-component.tsx
muddana-satish: Thanks for this explanation. I have switched back to looping through the active rows and then selecting the row if there is a match. That way we would never select an index that doesn't exist. I don't see the exception now.

packages/react-components/src/components/table-output-component.tsx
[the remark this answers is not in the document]
muddana-satish: @bhufmann  I fixed the other errors that you pointed out. could you please review the latest.

bhufmann (dismissed):
Looks good to me. Thanks for this improvement.

packages/react-components/src/components/table-output-component.tsx
PatrickTasse: The conditions to have an invalid selectStartIndex are still there, except that now they just don't select anything instead of cause an exception. So in the following example, the table ends up with no selection:

root INFO *** handleTimeSelectionChange({"start":1341337214,"end":1341337214})
root INFO vvv fetching table index @4738416805
root INFO *** handleTimeSelectionChange({"start":1341337214,"end":1341610057})
root INFO *** handleTimeSelectionChange({"start":1341337214,"end":1341630876})
root INFO *** handleTimeSelectionChange({"start":1341337214,"end":1342068391})
root INFO ^^^ fetching table index returns 4269
root INFO vvv fetching table lines @4200
root INFO *** handleTimeSelectionChange({"start":1341337214,"end":1341630876})
root INFO *** handleTimeSelectionChange({"start":1341337214,"end":1341610057})
root INFO *** handleTimeSelectionChange({"start":1341337214,"end":1341337214})
root INFO selectStartIndex=1.7976931348623157e+308
root INFO ^^^ fetching table lines returns
root INFO selectStartIndex=1.7976931348623157e+308

packages/react-components/src/components/table-output-component.tsx
```diff
@@ -30,9 +31,17 @@ export class TableOutputComponent extends AbstractOutputComponent<TableOutputPro
...
-    private lastIndex = { timestamp: Number.MIN_VALUE, index: 0 };
+    private prevStartTimestamp: number = Number.MIN_VALUE;
+    private startTimestamp: number = Number.MAX_VALUE;
+    private endTimestamp: number = Number.MIN_VALUE;
+    private isIndexChanged = false;
     private columnsPacked = false;
+    private timestampCol: string | undefined = undefined;
+    private eventSignal = false;
+    private enableTimestampSelection = false;
+    private enableIndexSelection = false;
```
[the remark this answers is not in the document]
muddana-satish: Updated. Thanks.

packages/react-components/src/components/table-output-component.tsx
```diff
@@ -263,15 +414,48 @@ export class TableOutputComponent extends AbstractOutputComponent<TableOutputPro
...
+                            this.selectEndIndex = Math.min(this.selectEndIndex, rowNode.rowIndex);
+                        } else {
+                            this.selectEndIndex = Math.max(this.selectEndIndex, rowNode.rowIndex);
+                        }
+                    }
+                }
+            });
+
+            if (this.gridApi.getSelectedNodes.length === 0 && !this.enableIndexSelection && isNodeSelected === false) {
+                this.selectStartIndex = this.selectEndIndex = -1;
```
PatrickTasse, line 457: The same problem still happens. See the log below.

I think this line is the problem. I don't think it's the job of selectRows() to clear the start/end index when enableIndexSelection is false.

I believe the intention is to have correct start/end index in case the user extends a time-based selection into an index-based selection, by shift-clicking on the event table after having selected a time range in a time graph.

In that case I believe that the right place to clear the start/end index is at the reception of a new time range in handleTimeSelectionRange(), but only when the start time has changed e.g. in the branch where it fetches the new start index, just before calling fetchTableIndex().

Another issue, above it is updating the end index when using time range selection. I believe that in some cases it also needs to set the start index (e.g. if it has never been set following a time range change). As far as I can tell, the effect of not setting it is that, after a time range selection, you can extend selection end by shift+clicking the table on an unselected event below the current multiple selection, but it does not work if you try to extend the selection start by shift-clicking the table on an unselected event above the current multiple selection.

root INFO *** handleTimeSelectionChange({"start":1341337214,"end":1341337214})
root INFO vvv fetching table index @4738416805
root INFO *** handleTimeSelectionChange({"start":1341337214,"end":1341545475})
root INFO selectRows() enableIndexSelection=false selectStartIndex=80 selectEndIndex=80
root INFO *** setting selectStartIndex = selectEndIndex = -1
root INFO *** handleTimeSelectionChange({"start":1341337214,"end":1341566293})
root INFO selectRows() enableIndexSelection=false selectStartIndex=-1 selectEndIndex=-1
root INFO *** setting selectStartIndex = selectEndIndex = -1
root INFO *** handleTimeSelectionChange({"start":1341337214,"end":1342005891})
root INFO selectRows() enableIndexSelection=false selectStartIndex=-1 selectEndIndex=-1
root INFO *** setting selectStartIndex = selectEndIndex = -1
root INFO ^^^ fetching table index returns 4269
root INFO handleTimeSelectionChange() setting selectStartIndex = selectEndIndex = 4269
root INFO selectRows() enableIndexSelection=false selectStartIndex=4269 selectEndIndex=4269
root INFO *** setting selectStartIndex = selectEndIndex = -1
root INFO vvv fetching table lines @4200
root INFO *** handleTimeSelectionChange({"start":1341337214,"end":1341566293})
root INFO selectRows() enableIndexSelection=false selectStartIndex=-1 selectEndIndex=-1
root INFO *** setting selectStartIndex = selectEndIndex = -1
root INFO *** handleTimeSelectionChange({"start":1341337214,"end":1341545475})
root INFO selectRows() enableIndexSelection=false selectStartIndex=-1 selectEndIndex=-1
root INFO *** setting selectStartIndex = selectEndIndex = -1
root INFO *** handleTimeSelectionChange({"start":1341337214,"end":1341337214})
root INFO selectRows() enableIndexSelection=true selectStartIndex=-1 selectEndIndex=-1
root INFO ^^^ fetching table lines returns
root INFO selectRows() enableIndexSelection=true selectStartIndex=-1 selectEndIndex=-1

PatrickTasse (changes requested):

packages/react-components/src/components/table-output-component.tsx
PatrickTasse: If you look at the log above, when fetchTableIndex() returns the selection start/end is set to 4269. But then when the time range selection is selected, the start/end is cleared to -1 because the table currently doesn't have a selection. So when the time range selection comes back to a single time and index-based selection is set, the start/end has been lost.

I also wonder, in selectRows() there is code to increase the end index when the range selection end time is increased, but it does not decrease the end index when the time range selection end time is decreased. For example, extending from [10, 20] to [10, 30] then back to [10, 25], the end index is stuck at 30. So if we really need the start/end index to be updated there, it will not work properly.

Also the updating of end index in selectRows() should only happen for selected row nodes! Otherwise it updates the end index just because a row exists, regardless of whether it is part of the selection. See example below, I was extending time selection to cover [1, 1] then [1, 2] then [1, 1] again, but end index becomes 199.

root INFO vvv fetching table lines @0
root INFO *** handleTimeSelectionChange({"start":11.833151488666372,"end":11.833151488666372})
root INFO vvv fetching table index @3397079602
root INFO ^^^ fetching table lines returns
root INFO selectRows() enableIndexSelection=true selectStartIndex=-1 selectEndIndex=-1
root INFO ^^^ fetching table index returns 1
root INFO handleTimeSelectionChange() setting selectStartIndex = selectEndIndex = 1
root INFO selectRows() enableIndexSelection=true selectStartIndex=1 selectEndIndex=1
root INFO setSelected(true) #1
root INFO *** handleTimeSelectionChange({"start":11.833151488666372,"end":747.4607357007592})
root INFO selectRows() enableIndexSelection=false selectStartIndex=1 selectEndIndex=1
root INFO setSelected(true) #1
root INFO setSelected(true) #2
root INFO *** handleTimeSelectionChange({"start":11.833151488666372,"end":464.4511959301551})
root INFO selectRows() enableIndexSelection=false selectStartIndex=1 selectEndIndex=199
root INFO setSelected(true) #1

bhufmann (dismissed):
It still looks good to me after the updates. Thanks alot. @PatrickTasse will have the final approval.

PatrickTasse (approved):
Looks good to me, thanks for your contribution!

bhufmann (approved):

PR #344, opened 2021-04-21
Support style hierarchy and composition in StyleProvider

PatrickTasse:
The OutputElementStyle's values map properties can now override the
properties defined for the style identified by the parent key.

A multi-level hierarchy of parent styles is supported. Style composition
by using a comma-delimited list of parent keys is supported.

The 'background-color', 'height', 'border-style', 'border-color' and
'border-width' properties are supported for time graph states.

The 'symbol-type', 'color', 'height' and 'vertical-align' properties are
supported for time graph annotations.

The 'opacity' property and '-blend' modifier are supported for color
styles.

The '-factor' modifier is supported for number styles.

### DIFF
--- a/packages/react-components/src/components/data-providers/style-properties.ts
+++ b/packages/react-components/src/components/data-providers/style-properties.ts
@@ -1,0 +1,13 @@
+export const StyleProperties = {
+    BACKGROUND_COLOR: 'background-color',
+    BLEND: '-blend',
+    BORDER_STYLE: 'border-style',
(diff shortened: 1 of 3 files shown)


bhufmann (approved):
It looks good to me. Please update the description in the TSP specification https://github.com/theia-ide/trace-server-protocol for that.

MatthewKhouzam (approved):
I was able to see the styles. Approved.

PR #345, opened 2021-04-27
Check time selection range for undefined value

arfio:
Set selectionRange can take an undefined value, so onSelectionRangeChange should have an optional value for the range variable.
Useful for https://github.com/theia-ide/timeline-chart/pull/123


MatthewKhouzam (approved):
LGTM

bhufmann (approved):

PR #347, opened 2021-04-28
Add support for range events.

ankusht-work:
This PR is dependent on: https://github.com/theia-ide/timeline-chart/pull/124

### DIFF
--- a/packages/react-components/src/components/data-providers/tsp-data-provider.ts
+++ b/packages/react-components/src/components/data-providers/tsp-data-provider.ts
@@ -29,13 +29,14 @@ export class TspDataProvider {
     }
 
     async getData(ids: number[], entries: TimeGraphEntry[], totalTimeRange: TimeRange,
-            viewRange?: TimelineChart.TimeGraphRange, resolution?: number): Promise<TimelineChart.TimeGraphModel> {
+        viewRange?: TimelineChart.TimeGraphRange, resolution?: number): Promise<TimelineChart.TimeGraphModel> {
         this.timeGraphEntries = [...entries];
         if (!this.timeGraphEntries.length) {
             return {
                 id: 'model',
                 totalLength: this.totalRange,
                 rows: [],
+                rangeEvents: [],
                 arrows: [],
                 data: {}
             };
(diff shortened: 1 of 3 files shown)


MatthewKhouzam (approved):
Depends, not depent in the pr description.

packages/react-components/src/components/data-providers/tsp-data-provider.ts
```diff
@@ -94,6 +100,11 @@ export class TspDataProvider {
             }
         }
         const arrows = await this.getArrows(ids, viewRange, resolution);
+
+        rows.forEach(row => {
+            row.annotations = row.annotations.concat(rangeEvents);
```
bhufmann, line 105: annotations in that span the whole view should not be stored in each row. They are an additional layer to be drawn on the chart. I think, there should be a new layer in the timeline-chart library.

Also, I noticed that the view gets very unresponsive. in my test trace I have only 5 such annotations and that should not impact the drawing performance. Could it be that in the timeline-chart each such marker is draw for each row>

bhufmann (changes requested):

packages/react-components/src/components/data-providers/tsp-data-provider.ts
ankusht-work: Added a new layer for lost event. Thank you for the feedback and suggestion!!

bhufmann (approved):
Looks good to me. Thanks.

packages/react-components/src/components/timegraph-output-component.tsx
```diff
@@ -472,7 +478,9 @@ export class TimegraphOutputComponent extends AbstractTreeOutputComponent<Timegr
                     size: symbolSize,
                     color: color ? color.color : 0x000000,
                     alpha: color ? color.alpha : 1.0,
-                    verticalAlign: vAlign ? vAlign : 'middle'
+                    verticalAlign: vAlign ? vAlign : 'middle',
+                    opacity: color ? color.alpha : 0.2
```
PatrickTasse, line 482: There is already support for alpha above, but I notice that it's not part of the TimeGraphAnnotationStyle definition and therefore not used by timeline-chart. It's the same for alpha in TimeGraphStateStyle.

I think we'll need to update this, but needs to be fixed in timeline-chart first.

The default should still be 1.0 if the style from TSP doesn't have a color. That will be fixed at the same time.

PatrickTasse (approved):

PR #357, opened 2021-05-07
Populate properties window on click of a row in events table

### LINKED ISSUE #348: Generate properties window for Events table
When user clicks on a row in the events table, the properties window should show the relevant information of the selected row.

image

muddana-satish:
fixes #348 


bhufmann (approved):
Looks good to me. Thanks for the improvement

bhufmann:
> For later, let's handle hyperlinks?

I think in the Properties view, if the key is "Source" the  TraceExplorerTooltipWidget assumes that the value is a source code location. It tries to match filename and line number and adds a handler to open the file. The assumption is not always correct, because the Source column could be, for example, source IP address or similar things. So, this source code lookup feature will need to be addressed differently (more standardize), which is outside the scope of this PR. I'll open a new PR for to track this.

Issue #240, opened 2021-01-25
Change title bar colour on mouse over


ebugden:
The entire title bar should change colour when the cursor is over it. This shows the user that they can click-drag on any part of the title bar to move the panel. 

![image](https://user-images.githubusercontent.com/28311615/95929770-c2d41f80-0d92-11eb-87d2-119627a940e5.gif)

Related: #239

CC: @ssmagula

IbrahimFradj:
I'll be working on it

Issue #239, opened 2021-01-25
Allow click and drag to move across the entire title bar


ebugden:
The click-drag possibility should be associated with the panel's whole title bar. This is intuitive and also makes the UI more robust since users won't need to be as precise with their mouvements.

Right now it's not possible to click and drag on some sections of the title bar below and above the panel title. In the video, look for where the cursor does not switch to the cross with arrows. This means the move action is not possible here. 

https://user-images.githubusercontent.com/28311615/105771274-4f63b180-5f2e-11eb-9a8b-da786ae8b94d.mp4

Example of desired behaviour from Grafana:
![image](https://user-images.githubusercontent.com/28311615/95929770-c2d41f80-0d92-11eb-87d2-119627a940e5.gif)

CC: @ssmagula

PR #359, opened 2021-05-07
Color change of the title bar on mouse over 

IbrahimFradj:
Related to #240 

The title bar changes color when the cursor is on it.

I added a hover color in the css.

Here is the result.
color

### DIFF
--- a/packages/react-components/style/output-components-style.css
+++ b/packages/react-components/style/output-components-style.css
@@ -10,6 +10,9 @@
     display: grid;
     grid-template-rows: 25px 1fr;
 }
+.widget-handle :hover {
+    background-color: var(--theia-list-hoverBackground); 
+}
 .title-bar-label {
     text-align: center;
     writing-mode: vertical-rl;


packages/react-components/style/output-components-style.css
```diff
@@ -10,6 +10,9 @@
     display: grid;
     grid-template-rows: 25px 1fr;
 }
+.widget-handle :hover {
+    background-color: var(--theia-selection-background);
```
ebugden, line 14: Please change the colour to something very subtle (ex. slightly lighten the bar colour) like in the example in the issue description #240. Here's the information priority that should be respected in this case:
1. **What action is possible:** The user should first notice that the cursor changes to a cross with arrows (this tells them they can click to move the panel). The cursor change is the most important piece of information.
2. **Where the action is possible:** The reason the title bar changes colour is just to show the the user the full area where they can hover to move the panel. It is secondary information so the colour change should be subtle. Right now the bright blue means the user is unlikely to notice the cursor change.

ebugden (changes requested):
Behaves as expected for me! The selection area doesn't look elegant, but that's because we still need #239 so in my opinion elegance isn't required for merging this PR. 

After the colour adjustment, I'm happy to have this merged! 🌻

packages/react-components/style/output-components-style.css
IbrahimFradj: Alright, I understand. Thank you for your comment.  changes are coming soon

ebugden: Thanks for the update. I like the new colour you chose!

ebugden (changes requested):
Before we merge, you need to squash (combine) the two commits in this PR into one commit. You can do this by interactively rebasing the commits using git.

New colour looks good! 🌈

bhufmann (approved):
Looks good to me. Thanks for the UX improvement.

PatrickTasse (approved):

PR #369, opened 2021-05-18
Click and drag to move across the entire title bar

### LINKED ISSUE #239: Allow click and drag to move across the entire title bar

IbrahimFradj:
We can now click and drag to move across the entire title bar 

Fixes #239
**Here is the result**

issue entire bar

### DIFF
--- a/packages/react-components/style/output-components-style.css
+++ b/packages/react-components/style/output-components-style.css
@@ -13,7 +13,7 @@
 .title-bar-label {
     text-align: center;
     writing-mode: vertical-rl;
-    height: 50%;
+    height: 100%;
     min-height: 150px;
     transform: rotate(180deg);
     user-select: none;


ebugden (approved):
Looks good to me! Making the whole title bar clickable gives the user a larger target and predictably sized target 🌳

MatthewKhouzam (approved):

bhufmann (approved):
LGTM! Thanks alot

PR #374, opened 2021-05-19
Reduce the resizable area of the widget

### LINKED ISSUE #265: Resize widget effective area is too large
When the mouse cursor is a certain distance from the resize widget, but not over it, the cursor changes to the resize cursor and dragging will resize the component.

This prevents clicking or dragging widgets that are under the cursor, such as the Event Table horizontal or vertical scroll bar.

muddana-satish:
fixes #265


bhufmann (approved):
Looks good to me!

Issue #352, opened 2021-05-04
Provide pre-compiled application packaged with trace server


ebugden:
In some cases, if users want to visualize their own traces, they still need to do the full complicated dev setup. The live demo allows users to upload traces, but this is not safe for private traces and not possible for very large traces. Providing a downloadable binary is a nice midway point between the live demo and the full dev setup.

- [x] **Build electron app packaged with server for Linux**
- [x] **Shrink packaged application to reasonable size range** - @arfio completed a quick package (it is 600 MB which is a large for a typical electron app #378) 
- [x] **Test app works without requiring external dependencies** - @arfio and @thefinaljob successfully tested the app on a clean install of Ubuntu. 
   - The only missing dependency is Java which needs to be installed manually before using the app #377. 
   - Packaged app only works on a full Linux VM or native Linux (does not work using the Windows subsystem for Linux (wsl) command)
- [x] **Host app permanently on publicly available server** (Github has a file size limit of 100 MB) - The app package is hosted [here](https://www.dorsal.polymtl.ca/files/other/electron-theia-trace-example-app-0.0.1.AppImage) on a DORSAL lab server (a main project contributor). @thefinaljob says it takes a few minutes to download (~5 on slower wifi).
- [x] **Upstream changes required to package application** with the server (PR #594) - This makes it easy to periodically generate an updated version of the package that can then be made publicly available
- [x] **Update Github readme** with: (PR #392)
   * **Link to application**, **build date** (so users know how old the build is), specify that installing **Java is required**
   * Specify that currently it's not possible to view your own traces with the live demo, and that to view your own traces you need to install the tool.
- [ ] **Automate?** building, packaging and updating the app (update twice a month?)

* ~~**Put everything in a Docker container?** So that users don't have to install Java (would have to install Docker instead)~~ Java is easier to install than Docker for Linux users so that will be the dependency (arfio [comment](https://github.com/theia-ide/theia-trace-extension/pull/392#discussion_r669860963))
   * ~~This will be more of an issue once we can build packages for Windows (because Java is awkward to install on Windows)~~
   * ~~On Linux Java is not difficult to install, but people have an aversion to it~~

----

Related: #59

ebugden:
- [x] **Build electron app packaged with server for Linux** - @arfio completed a quick package (it is 600 MB which is a large for a typical electron app) 
[...]

PR #387, opened 2021-06-08
Include trace server with packaged app

arfio:
Modifies the way the path is obtained to work with the AppImage application. Also adds meaningful error messages when the path is not correct.

Related to #352 

Motivation: Currently, users need to compile the Trace Viewer from source to view their own traces. To avoid the time-consuming dev setup and compilation we now offer a downloadable package (electron desktop application + trace server). Since this package currently needs to be generated and uploaded manually, this PR includes the packaging changes required so that any dev can easily create an updated package in the future.

### DIFF
--- a/theia-extensions/viewer-prototype/src/node/trace-server-service.ts
+++ b/theia-extensions/viewer-prototype/src/node/trace-server-service.ts
@@ -1,13 +1,23 @@
 import { spawn } from 'child_process';
+import { join } from 'path';
+import { existsSync } from 'fs';
 import { injectable } from 'inversify';
 import { PortBusy, TraceServerConfigService } from '../common/trace-server-config';
+import { FileOperationError, FileOperationResult } from '@theia/filesystem/lib/common/files';
 import treeKill = require('tree-kill');
 @injectable()
 export class TraceServerServiceImpl implements TraceServerConfigService {
     private processId: number;
 
     startTraceServer(path: string | undefined, port: number | undefined): Promise<string> {
-        const server = spawn(`${path}`, ['-vmargs', `-Dtraceserver.port=${port}`]);
+        if (typeof path === 'undefined') {
+            throw new FileOperationError('The Trace Viwer path is not defined.', FileOperationResult.FILE_INVALID_PATH);
+        }
+        const serverPath = join(__dirname, '../../../../', path);
+        if (existsSync(serverPath) === false) {
+            throw new FileOperationError(`The Trace Viewer path (${serverPath}) was not found.`, FileOperationResult.FILE_NOT_FOUND);
+        }
+        const server = spawn(`${serverPath}`, ['-vmargs', `-Dtraceserver.port=${port}`]);
         this.processId = server.pid;
         const timeouts: NodeJS.Timeout[] = [];
 
--- a/examples/browser/package.json
+++ b/examples/browser/package.json
@@ -9,7 +9,7 @@
         "applicationName": "Theia-Trace Example Application",
         "preferences": {
           "editor.autoSave": "on",
-          "trace-viewer.path" : "../../trace-compass-server/tracecompass-server",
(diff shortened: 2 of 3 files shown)


theia-extensions/viewer-prototype/src/node/trace-server-service.ts
```diff
@@ -1,13 +1,23 @@
...
 @injectable()
 export class TraceServerServiceImpl implements TraceServerConfigService {
     private processId: number;
 
     startTraceServer(path: string | undefined, port: number | undefined): Promise<string> {
-        const server = spawn(`${path}`, ['-vmargs', `-Dtraceserver.port=${port}`]);
+        if (typeof path === 'undefined') {
+            throw new FileOperationError('The Trace Viwer path is not defined.', FileOperationResult.FILE_INVALID_PATH);
+        }
+        const serverPath = join(__dirname, '../../../../', path);
```
bhufmann, line 16: with this it will be required that the server is relative to the patch that is configure for it. This will only work in what you want to achieve in this patch, but breaks other setups.

The path is configured in a preference so that it can be changed by the user.

bhufmann (changes requested):

theia-extensions/viewer-prototype/src/node/trace-server-service.ts
arfio: you're right if the path is absolute, it will not work. However, for all setups this will default to the main directory of where the theia-trace-extension is executed. I can change the implementation so that it takes care of absolute path correctly. Are there any thing else to be taken into account ?

theia-extensions/viewer-prototype/src/node/trace-server-service.ts
```diff
@@ -1,13 +1,23 @@
...
+        if (existsSync(serverPath) === false) {
+            throw new FileOperationError(`The Trace Viewer path (${serverPath}) was not found.`, FileOperationResult.FILE_NOT_FOUND);
+        }
+        const server = spawn(`${serverPath}`, ['-vmargs', `-Dtraceserver.port=${port}`]);
```
ankusht-work, line 20: I suggest to keep this implementation as is (read values straight from the preferences) and let the user specify the exact path to the server.

arfio: The usage of __dirname is mandatory if I want the application to work with an AppImage. Because by default the relative path is towards the location of the AppImage file, and not the mounted directory.

theia-extensions/viewer-prototype/src/node/trace-server-service.ts
colin-grant-work: We need to be able to handle a couple of contingencies here. Whether or not the server is bundled with the app image, it should be possible for the user to specify a path to their preferred executable.

I suggest the following:
 - Remove the default for the preference specifying the location of the server executable when the application is packaged, but not when it is built locally.
 - Allow `path` to be `undefined` - i.e. don't throw an invalid path error immediately.
 - If `path` is `undefined` or an empty string, use the expected packaged directory.

colin-grant-work: And I'm suggesting that that value be removed from the version of the `package.json` included with the packaged electron application. When the application is packaged, the default will be to use the built-in binary, rather than an external program. Since that behavior is different from the unpackaged version, a change needs to be made during the packaging process.

arfio:
A workaround to not break any other AppImages that would have relative path to the AppImage location (and not the temporary path of the AppImage), would be to set an environment variable that toggles this behaviour.
The only way I found to include that environment variable at package time with the electron-builder is to make a script that hooks onto the [AfterPackContext hook](https://www.electron.build/configuration/configuration#afterpackcontext). This script will then change the packaged app to set an env variable when launching the AppImage.

arfio:
This pull request have been superseeded by #594. Closing this for now

PR #391, opened 2021-06-11
Hide analysis that is not implemented on the client side

### LINKED ISSUE #296: Hide "analyses" that do not provide value to the user

IbrahimFradj:
I have filtered out the view type that is not implemented in the client side (DATA_TREE). The expected result is that views of this type are not displayed in the available views 

Contributes towards fixing #296

**Before**

image

**After**

image

In these two images, we notice that the Futex contention analysis is not shown in the next image because this view has a DATA_TREE type. 

### DIFF
--- a/packages/react-components/src/trace-explorer/trace-explorer-views-widget.tsx
+++ b/packages/react-components/src/trace-explorer/trace-explorer-views-widget.tsx
@@ -158,6 +158,6 @@ export class ReactAvailableViewsWidget extends React.Component<ReactAvailableVie
         if (descriptors && descriptors.length) {
             outputDescriptors.push(...descriptors);
         }
-        return outputDescriptors;
+        return outputDescriptors.filter(value => value.type !== 'DATA_TREE');
     }
 }


ebugden (changes requested):
Requested changes:
* **Change "Fixes #296" to something like "Contributes towards fixing #296"** - This is the client side of the work (still need server fixes) so we don't want to close the issue when this is merged. 

🌈 Thank you for this contribution! Non functional analyses were visibly frustrating for users during UX tests so this will have a significant positive impact.  

Question: How many views did this filter eliminate?

MatthewKhouzam (approved):
lgtm

bhufmann (approved):
Thanks. Totally makes sense.

bhufmann:
> Requested changes:
> 
>     * **Change "Fixes #296" to something like "Contributes towards fixing #296"** - This is the client side of the work (still need server fixes) so we don't want to close the issue when this is merged.
> 
> 
This has been done by Ibrahim and it's ready to be merged.

ebugden (approved):
Thanks for the update @IbrahimFradj! And thanks for the approval reminder @bhufmann!

PR #394, opened 2021-06-14
Add Toolbar Icons for Zoom In, Zoom Out and Reset.

muddana-satish:
depends on #https://github.com/theia-ide/timeline-chart/pull/140

### DIFF
--- a/theia-extensions/viewer-prototype/src/browser/trace-viewer/trace-viewer-toolbar-commands.ts
+++ b/theia-extensions/viewer-prototype/src/browser/trace-viewer/trace-viewer-toolbar-commands.ts
@@ -1,0 +1,22 @@
+import { Command } from '@theia/core';
+
+export namespace TraceViewerToolbarCommands {
+
+    export const ZOOM_IN: Command = {
+        id: 'trace.viewer.zoomin',
+        label: 'Zoom In',
+        iconClass: 'fa fa-plus-square-o fa-lg',
+    };
+
+    export const ZOOM_OUT: Command = {
+        id: 'trace.viewer.toolbar.zoomout',
+        label: 'Zoom Out',
+        iconClass: 'fa fa-minus-square-o fa-lg',
+    };
+
+    export const RESET: Command = {
+        id: 'trace.viewer.toolbar.reset',
+        label: 'Reset',
+        iconClass: 'fa fa-home fa-lg',
+    };
+}
--- a/packages/base/src/signals/signal-manager.ts
+++ b/packages/base/src/signals/signal-manager.ts
@@ -17,6 +17,8 @@ export declare interface SignalManager {
     fireSelectionChangedSignal(payload: { [key: string]: string }): void;
     fireCloseTraceViewerTabSignal(traceUUID: string): void;
     fireTraceViewerTabActivatedSignal(experiment: Experiment): void;
+    fireZoomTimeGraphSignal(hasZoomedIn: boolean): void;
+    fireResetTimeGraphSignal(): void;
 }
 
 export const Signals = {
@@ -32,7 +34,9 @@ export const Signals = {
     THEME_CHANGED: 'theme changed',
     SELECTION_CHANGED: 'selection changed',
     CLOSE_TRACEVIEWERTAB: 'tab closed',
-    TRACEVIEWERTAB_ACTIVATED: 'widget activated'
+    TRACEVIEWERTAB_ACTIVATED: 'widget activated',
+    TIMEGRAPH_ZOOMED: 'timegraph zoomed',
+    TIMEGRAPH_RESET: 'timegraph reset'
 };
 
 export class SignalManager extends EventEmitter implements SignalManager {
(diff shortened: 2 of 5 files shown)


muddana-satish:
Added three icons for zoom in, zoom out and reset. 

**Zoom In/ Zoom Out:** Zooms at the center of the chart if the selection range is not specified. Else zooms at the middle of the selection range.

**Reset:** Scales to the default level.


Kapture 2021-06-14 at 13 46 44

MatthewKhouzam:
I would like Stefan or Erica's opinion about the icons. I instinctively thougth the + or - was to add or remove entries.

MatthewKhouzam:
Also reset is backwards reload. ;)

theia-extensions/viewer-prototype/src/browser/trace-viewer/trace-viewer-toolbar-commands.ts
```diff
@@ -0,0 +1,22 @@
+import { Command } from '@theia/core';
+
+export namespace TraceViewerToolbarCommands {
+
+    export const ZOOM_IN: Command = {
+        id: 'trace.viewer.zoomin',
+        label: 'Zoom In',
+        iconClass: 'fa fa-plus',
```
MatthewKhouzam, line 8: https://fontawesome.com/v4.7/icon/search-plus ?

theia-extensions/viewer-prototype/src/browser/trace-viewer/trace-viewer-toolbar-commands.ts
```diff
@@ -0,0 +1,22 @@
...
+    };
+
+    export const ZOOM_OUT: Command = {
+        id: 'trace.viewer.toolbar.zoomout',
+        label: 'Zoom Out',
+        iconClass: 'fa fa-minus',
```
MatthewKhouzam, line 14: fa-search-minus ?

theia-extensions/viewer-prototype/src/browser/trace-viewer/trace-viewer-toolbar-commands.ts
```diff
@@ -0,0 +1,22 @@
...
+    };
+
+    export const RESET: Command = {
+        id: 'trace.viewer.toolbar.reset',
+        label: 'Reset',
+        iconClass: 'fa fa-undo',
```
MatthewKhouzam, line 20: https://fontawesome.com/v4.7/icon/map

MatthewKhouzam (commented):
Not a hard minus, I have questions and suggestions though for the icon selection

bhufmann:
I think the idea was to use: 
image

MatthewKhouzam:
@bhufmann  yes, the box makes a world of difference IMO. It gives a context that the windows is zoomed.

theia-extensions/viewer-prototype/src/browser/trace-viewer/trace-viewer-toolbar-commands.ts
MatthewKhouzam: https://fontawesome.com/v4.7/icon/plus-square-o !!! this one! this one!

theia-extensions/viewer-prototype/src/browser/trace-viewer/trace-viewer-toolbar-commands.ts
MatthewKhouzam: https://fontawesome.com/v4.7/icon/minus-square-o

theia-extensions/viewer-prototype/src/browser/trace-viewer/trace-viewer-toolbar-commands.ts
muddana-satish: image
 
They're very dull for some reason :/

ebugden:
> **bhufmann:** I think the idea was to use:
> image

Without reading this exchange in detail, my understanding is that the icons added in this PR should be the exact icons used in the wireframe that @ssmagula tested on users recently. One of the questions asked was what the icons meant (without having a chance to try the buttons) and users had no issues identifying the actions associated with them. Other than that I haven't been much involved in the icon selection process.

MatthewKhouzam:
I found two of the three icons... see the comments. 3rd one is tricky.

ebugden:
Is it not in the Figma project? I don't think the icons were created by hand.

MatthewKhouzam (approved):
I am OK with this, we need to do two things:

1. Update the home button when we find the right icon
2. make it work on the time axis in the future so it will affect line charts too.

bhufmann (approved):
Looks good to me. Thanks!

ebugden:
> I am OK with this, we need to do two things:
> 
> 1. Update the home button when we find the right icon
> 2. make it work on the time axis in the future so it will affect line charts too.

Can these Todos be made into issues, so they don't get forgotten?

Issue #385, opened 2021-06-08
Flame Chart stays in theme it was created with


ebugden:
The Flame Chart doesn't switch themes. It stays in the theme it was created with.

https://user-images.githubusercontent.com/28311615/121206787-be3f0880-c846-11eb-989e-8fdae1cd9120.mp4

PR #396, opened 2021-06-21
Support searching in events table

muddana-satish:


bhufmann (changes requested):
Overall the feature looks already good and you're on good path. I started with testing the functionality and I noticed a few things:
1- the yellow background color makes the text hard to read in dark theme (white on yellow background). The highlighting color should be different depending on the theme. For light theme the yellow is good.  In Trace Compass dark green (0x008000) works well.
2- I can't type in to search box, after clicking once into a search box. I have to click twice
search-2-click
3- When searching for a string, the highlighted text is put at the beginning of the cell text 
wrong-replace
4-The next and close button is not visible when the column is outside the window range (not sure how to fix it). 
image
5-The progress wheel (when searching or reloading a new page) is shown now in all cells and not in the first column.
table-progress-wheel

Will continue with reviewing the code while can look at the things above.

MatthewKhouzam (commented):
In the commit message, please explain why we bump ag-grid community. Also why move fetchTableLines

muddana-satish:
> In the commit message, please explain why we bump ag-grid community. Also why move fetchTableLines

Updated the commit message.

muddana-satish:
@bhufmann. Thanks for the feedback and comments. I have updated the code that fixes few of the issues you pointed out.

> 1- the yellow background color makes the text hard to read in dark theme (white on yellow background). The highlighting color should be different depending on the theme. For light theme the yellow is good. In Trace Compass dark green (0x008000) works well.

Update the background color of the matching text in the dark theme. However for the change to reflect, the table has to be relaunched. 

> 2- I can't type in to search box, after clicking once into a search box. I have to click twice

Set the auto focus to the input field after clicking the search filter.

> 3- When searching for a string, the highlighted text is put at the beginning of the cell text

I didn't realize it's a substring match. Updated the code to highlight all the matches of the filtered text.

> 5-The progress wheel (when searching or reloading a new page) is shown now in all cells and not in the first column. Minor thing. Can be also addressed afterwards if we agree on to do so.

This change was on purpose. I felt having the spinner at all cells is a good UX.

MatthewKhouzam (approved):
Works and I like it. please address Bernd's comments. Code looks good to me.

bhufmann:
> @bhufmann. Thanks for the feedback and comments. I have updated the code that fixes few of the issues you pointed out.
> 
> > 1- the yellow background color makes the text hard to read in dark theme (white on yellow background). The highlighting color should be different depending on the theme. For light theme the yellow is good. In Trace Compass dark green (0x008000) works well.
> 
> Update the background color of the matching text in the dark theme. However for the change to reflect, the table has to be relaunched.
> 
That's ok for now. This is similar #385 and should be fixed together.

> > 2- I can't type in to search box, after clicking once into a search box. I have to click twice
> 
> Set the auto focus to the input field after clicking the search filter.
> 
It works now.
> > 3- When searching for a string, the highlighted text is put at the beginning of the cell text
> 
> I didn't realize it's a substring match. Updated the code to highlight all the matches of the filtered text.
> 
It works now.
> > 5-The progress wheel (when searching or reloading a new page) is shown now in all cells and not in the first column. Minor thing. Can be also addressed afterwards if we agree on to do so.
> 
> This change was on purpose. I felt having the spinner at all cells is a good UX.
I'm fine with it.

bhufmann (dismissed):
@muddana-satish Thanks for the quick updates and fixes. I noticed that the search starts always from the beginning of the trace. In the example gif below, I moved the timestamp in the table using the Resource view, then I started a search and the table gets populated from the beginning of the trace. The scrollbar of the tables indicates that. Moreover, the Resources view doesn't update it's time selection. I would expect that the search starts from the selected event (start even if a range is selected) instead of the beginning of the trace. If no event is selected, then start from the beginning of the trace.

out-of-sync-table

packages/react-components/src/components/table-output-component.tsx
```diff
@@ -331,13 +301,31 @@ export class TableOutputComponent extends AbstractOutputComponent<TableOutputPro
...
+                        searchResultsColor: this.props.backgroundTheme === 'light' ? '#cccc00' : '#008000'
+                    },
+                    suppressMenu: true,
+                    filter: 'agTextColumnFilter',
+                    floatingFilter: true,
+                    floatingFilterComponent: 'searchFilterRenderer',
+                    floatingFilterComponentParams: {
+                        suppressFilterButton: true,
+                        onFilterChange: this.searchEvents,
+                        onclickNext: this.findNextMatchedEvent,
```
PatrickTasse, line 317: I expected it to find the next matched event incrementally as I type in the text box, not only after pressing Enter or clicking the next button.

It creates this strange behavior:

1. Type 'x' where x matches in that column at row 100.
The table is updated to dim the visible rows 0-10.
I expected it to go to row 100.

2. Press Enter or click Next button.
The table goes to and selects row 100.

3. Delete 'x' so that regex is empty
The table goes back to row 0 and undims all the rows.
I expected it to stay at row 100.
(note, row 100 is still selected even though it is not visible)

4. Type 'x' again
The table is updated to dim visible rows 0-10

5. Press Enter or click Next button
The table goes to and selects the next match at row 200.
This is OK since the selection was row 100, but it's confusing because the table had scrolled back to row 0.

PatrickTasse (changes requested):

PatrickTasse (approved):
Issues to be opened to fix later.

Issue #45, opened 2020-05-25
XY Chart: Support all styles in Trace Compass


tahini:
The XY chart part should support all styles:

* scatter
* line
* bar
* stack
* area

PR #399, opened 2021-06-22
XY Chart: Add area and scatter chart styles

thefinaljob:
Added area and scatter chart styles through a dropdown menu

Contributes towards fixing #45

### DIFF
--- a/packages/react-components/src/components/xy-output-component.tsx
+++ b/packages/react-components/src/components/xy-output-component.tsx
@@ -3,6 +3,7 @@ import { AbstractOutputProps, AbstractOutputState } from './abstract-output-comp
 import { AbstractTreeOutputComponent } from './abstract-tree-output-component';
 import * as React from 'react';
 import { Line } from 'react-chartjs-2';
+import {Scatter} from 'react-chartjs-2';
 import { QueryHelper } from 'tsp-typescript-client/lib/models/query/query-helper';
 import { Entry } from 'tsp-typescript-client/lib/models/entry';
 import { ResponseStatus } from 'tsp-typescript-client/lib/models/response/responses';
@@ -12,6 +13,7 @@ import { EntryTree } from './utils/filtrer-tree/entry-tree';
 import { getAllExpandedNodeIds } from './utils/filtrer-tree/utils';
 import { TreeNode } from './utils/filtrer-tree/tree-node';
 import ColumnHeader from './utils/filtrer-tree/column-header';
+import { ChangeEvent } from 'react';
 
 type XYOuputState = AbstractOutputState & {
     selectedSeriesId: number[];
@@ -19,16 +21,27 @@ type XYOuputState = AbstractOutputState & {
     checkedSeries: number[];
     collapsedNodes: number[];
     orderedNodes: number[];
+    chartStyle: any;
     // FIXME Type this properly
     xyData: any;
     columns: ColumnHeader[];
 };
 
+class xyPair {
+    x: number;
+    y: number;
+    constructor(x: number, y: number) {
+        this.x = x;
+        this.y = y;
+      }
+}
+
 export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutputProps, XYOuputState> {
     private currentColorIndex = 0;
     private colorMap: Map<string, number> = new Map();
 
-    private lineChartRef: any;
+    protected desiredChartStyle: any;
+    private chartRef: any;
     private mouseIsDown = false;
     private posPixelSelect = 0;
     private plugin = {
@@ -40,7 +53,7 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
             const scale = this.props.viewRange.getEnd() - this.props.viewRange.getstart();
             this.props.unitController.selectionRange = {
                 start: xStartPos,
-                end: xStartPos + ((event.screenX - this.posPixelSelect) / this.lineChartRef.current.chartInstance.width) * scale
+                end: xStartPos + ((event.screenX - this.posPixelSelect) / this.chartRef.current.chartInstance.width) * scale
             };
         }
     };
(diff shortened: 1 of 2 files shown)


MatthewKhouzam (commented):
Initial comment

https://www.freecodecamp.org/news/writing-good-commit-messages-a-practical-guide/

MatthewKhouzam:
Initial hot take:

* Setting the default to scatter is not a great idea. offering scatter is.
* Using a dropdown may not be as interesting as using toggle buttons
* Scatters show latencies better 
image
* The UI freezes in Linux/Firefox with this patch. Not sure if it's related.
* You need to sign off your patches if you want them to pass DCO
* Stacked area makes sense for very few views, I would omit it from the 1st patch.
* Most importantly, this is really cool! 

Initial action items:
* Fix your commit message
* Make line the default
* Use toggles (icons can be <i class="far fa-chart-scatter"></i> https://fontawesome.com/v5.15/icons/chart-line?style=regular)
* Remove the stacked area as it implies a total which is not the case often
* Have a github issue associated to the commit message

I only looked at the functionality, not the code, so this can be considered a "sanity check"

MatthewKhouzam:
Also funny bug, the more I resize the view and play with it, the smaller the dropdown gets.
image

ebugden:
I recently learned that each data series in an xy chart specifies the chart type/style (line, scatter) that should be used to visualize it. I think this PR should be adjusted to automatically **display data based on the chart type specified in the series label** rather than offering a drop down menu currently proposed. @MatthewKhouzam @tahini thoughts?

tahini:
> I recently learned that each data series in an xy chart specifies the chart type/style (line, scatter) that should be used to visualize it. I think this PR should be adjusted to automatically **display data based on the chart type specified in the series label** rather than offering a drop down like I initially suggested to @thefinaljob. Allowing all chart types implies that any option makes sense which isn't the case (ex. you wouldn't use a scatter plot to visualize continuous time-based data). @MatthewKhouzam @tahini thoughts?

Yes! That was indeed discussed offline with Nikolai. Note that the data type from the server comes with the series (for instance, I think the CPU usage has the total as a line and individual threads as area), not the whole chart, while iiuc, with chartjs, the series type will have to be for the whole chart. Would you default to the first one you get?

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -303,11 +437,80 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
         const tspClientResponse = await this.props.tspClient.fetchXY(this.props.traceId, this.props.outputDescriptor.id, xyDataParameters);
         const xyDataResponse = tspClientResponse.getModel();
         if (tspClientResponse.isOk() && xyDataResponse) {
-            this.buildXYData(xyDataResponse.model.series);
+            switch (this.desiredChartStyle) {
+                case 'scatter':
+                    this.buildScatterData(xyDataResponse.model.series);
```
tahini, line 442: As discussed offline, if different types of chart require the data to be formatted differently, they should be encapsulated together in a single class so that there can be no mismatch. You could add separate widget classes for each series type (in react, it is suggested to multiply small stateless classes instead of big ones with ton of stuff in it). So you could have ScatterXy, LineXy, AreaXy, which would take the series as props and would format the data appropriately and render the proper line type. 

Here, you could definitely have a race condition between changing the chart type and zooming in with the old chart type and get the view to temporarily render wrong, until it is rendered right ;-)

But I'll let the maintainers of the extension decide if my comment is purely informative or if you should act on it

thefinaljob:
Reopeing after a snafu

thefinaljob:
This is a mess, this is what happens when you try to fix solder.... several times.

thefinaljob:
OK looks like the submission is ready. To address @MatthewKhouzam's comments.

- After discussing with Erica, we are keeping the dropdown as is, using a toggle for each individual chart style would take up too much real estate. Also, having a toggle between a chart and table style would not be optimal given that a lot of the time there isn't a table view available
- We cannot have the chart style be chosen automatically yet as the API is sending a "null" value. The endpoint exists but it is not usable as of yet. I will issue a bug report regarding this.
- Line has been made default
- The UI freeze typically happens when too many data points are chosen, this could be fixed by allowing for a max number of data points to visualized
- After discussing with Erica, the stacked area chart will stay. I forgot the reasons though @ebugden can chime in here :) 

Thank you for your feedback Matthew!! This was my first serious PR and it was great to get a lot of good pointers in the right direction. 

Known bugs
- Border is not set to 0 when changing from line to stacked area, after talking to @tahini it seems that only dividing each of these chart styles into separate components would fix the issue as they would be destroyed and re-initialized upon a change in the chart style.

MatthewKhouzam:
Hey, before getting into the review proper, please remember to update the commit message to follow the guidelines.

The reasoning being if and when something goes wrong, it's jarring to have a Hodge-podge of commit messages to go through, standardized ones will make things clearer in our time of duress. :)

MatthewKhouzam:
Stacked area charts freeze on my computer. Could you profile it with the apt trace CPU usage and clicking on all?

ebugden:
> Hey, before getting into the review proper, please remember to update the commit message to follow the guidelines.
> 
> The reasoning being if and when something goes wrong, it's jarring to have a Hodge-podge of commit messages to go through, standardized ones will make things clearer in our time of duress. :)

@thefinaljob the commit guidelines were recently added to the main readme in the [How to contribute code](https://github.com/theia-ide/theia-trace-extension#how-to-contribute-code) section.

MatthewKhouzam (changes requested):
Please update the commit message and fix the performance issue.

mirsky-work:
I just tested the functionality on Gitpod - very cool!
A nitpicking - right now the [X, Y] values' label is shown when selecting a scatter chart element. Can we show it on mouse hover?

image

ebugden:
Exceptionally, I'm going to make edits to the title and description of this PR to correct style issues.

ebugden:
> any news on this one.

@arfio could take a look at this if they have free time.

bhufmann:
Closing this one after #595 was merged. Thanks for the effort it helped with the current implementation.

PR #403, opened 2021-06-24
Retrieve port, path values from PreferenceService as required

colin-grant-work:
The current code in the `TraceViewerContribution` tracks the value of the preferences for port and path by listening to preference change events emitted by the `PreferenceService`. This can have unintended side effects because if a user resets a setting specified in a `settings.json` file, the [preference provider for that scope will emit an event](https://github.com/eclipse-theia/theia/blob/62116bfc756fcf61622f64646a831af78695beda/packages/preferences/src/browser/abstract-resource-preference-provider.ts#L230-L234) indicating that its local value has been set to `undefined`. The desired behavior of the `TraceViewerContribution` is that we should then use the default value, but the current code will use the `undefined` value.

This PR dispenses with the value tracking code in favor of simply retrieving the values from the preference service when they're required, ensuring that we fall back to the default as appropriate without getting confused by the noise of change events.

### Bad behavior:
 1. On current master
 2. Change the setting for trace server path to some non-default value using the preferences UI
 3. Click the cog next to the setting and select 'reset setting'
 4. Attempt to open a trace
 5. Observe that the server fails to start.
 6. (If you'd like, debug the listener code in `TraceViewerContribution` and confirm that `this.path` is `undefined`.)

### New behavior:
 1. Pull this branch
 2. Perform the steps outlined above
 3. Assuming that you have a server at the default path, observe that that server is started.

### DIFF
--- a/theia-extensions/viewer-prototype/src/browser/trace-viewer/trace-viewer-contribution.ts
+++ b/theia-extensions/viewer-prototype/src/browser/trace-viewer/trace-viewer-contribution.ts
@@ -1,4 +1,4 @@
-import { injectable, inject, postConstruct } from 'inversify';
+import { injectable, inject } from 'inversify';
 import { CommandRegistry, CommandContribution, MessageService } from '@theia/core';
 import { WidgetOpenerOptions, WidgetOpenHandler } from '@theia/core/lib/browser';
 import URI from '@theia/core/lib/common/uri';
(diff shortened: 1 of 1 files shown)


MatthewKhouzam (approved):
Get the build to work please

MatthewKhouzam:
https://github.com/theia-ide/theia-trace-extension/pull/403/checks?check_run_id=2906183868

colin-grant-work:
@MatthewKhouzam, I pushed during Eclipse's recent server outage, so the build failed at the plugin download stage. Hopefully a fresh push will have more success.

ankusht-work (approved):
LGTM. Thank you colin for the fix.

bhufmann (approved):
Locks good to me! Fixes the bug as described. Thanks for the contribution.

MatthewKhouzam (approved):
LGTM

PR #408, opened 2021-07-01
Allow events table search from the current selected row

### LINKED ISSUE #400: Events table search always start from the first row.
https://github.com/theia-ide/theia-trace-extension/pull/396#pullrequestreview-689757368

muddana-satish:
fixes #400

### DIFF
--- a/packages/react-components/src/components/table-output-component.tsx
+++ b/packages/react-components/src/components/table-output-component.tsx
@@ -476,7 +476,18 @@ export class TableOutputComponent extends AbstractOutputComponent<TableOutputPro
         } else {
             this.filterModel.set(colName, filterValue);
         }
-        this.gridApi?.setDatasource(this.dataSource);
+        if (this.gridApi) {
+            this.gridApi.forEachNode(rowNode => {
+                let isMatched = true;
+                this.filterModel.forEach((value, key) => {
+                    if (!rowNode.data[key].includes(value)) {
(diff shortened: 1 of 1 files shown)


MatthewKhouzam (commented):
This does what it should. But the commit message should be

"Allow events table search from the selected row" as you can move with the scrollbar and it still works well.

bhufmann (dismissed):
I tested it and the UX seems good to me. 

When entering a search string, then matched strings in the visible rows are updated. When pressing enter, it goes to the next matching event (starting from the selected row) regardless if it's in the current cache or if it needs to fetch new data from the back-end).

Thanks for the improvement

PatrickTasse (dismissed):

muddana-satish:
> "Allow events table search from the selected row" as you can move with the scrollbar and it still works well.

It actually works even when no row is selected. :)

muddana-satish:
Updated the commit message. Created #410 for regular expression searching.

bhufmann:
> Updated the commit message. Created #410 for regular expression searching.

I actually like the initial commit message (and PR title) better than the new one. It's more obvious what the improvement is. 

Maybe only replace "visible" with "selected":
"Allow events table search from the current selected row"

bhufmann (approved):
Thanks!

PatrickTasse (approved):

PR #417, opened 2021-07-14
Do not open TraceViewerWidget when initializing layout

### LINKED ISSUE #411: `TraceViewerWidget` with no ID created when application started with empty layout
# Bug

If the application is started with no layout to restore, a `TraceViewerWidget` with no ID will be created, which is considered an error in Theia. 

## To reproduce

1. Open the example application.
2. Open the command palette and `View: Reset Workbench Layout`
3. In the console, you will see an error with this message:
image
4. Open the dev tools and open `application-shell.ts` and add a breakpoint around line 714:
image
5. Rerun the comand, and when the breakpoint is tripped, you'll see this:
image

colin-grant-work:
This PR fixes #411 by removing the bindings that were causing a `TraceViewerWidget` to be created when the layout was restored.

The `TraceViewerToolbarContribution` was `extending` `AbstractViewContribution` and implementing `initializeLayout`, but it doesn't need any `ViewContribution` functionality and should not try to open a widget on startup.

To test:
1. Open a trace and view some trace graphs.
2. You should have access to the same commands and toolbar items as on `master`.
3. Run the `View: Reset Workbench Layout` command.
4. You should not see any logs about widgets missing id's or inability to read `traceURI` of `null`.

> I see an error `Cannot read property 'annotationCategories' of null` logged by `trace-explorer-views-widget.tsx`, but I believe that that error is unrelated to these changes.

### DIFF
--- a/theia-extensions/viewer-prototype/src/browser/trace-viewer/trace-viewer-frontend-module.ts
+++ b/theia-extensions/viewer-prototype/src/browser/trace-viewer/trace-viewer-frontend-module.ts
@@ -28,9 +28,10 @@ export default new ContainerModule(bind => {
     bind(TspClientProvider).toSelf().inSingletonScope();
     bind(TheiaMessageManager).toSelf().inSingletonScope();
 
-    bindViewContribution(bind, TraceViewerToolbarContribution);
+    bind(TraceViewerToolbarContribution).toSelf().inSingletonScope();
     bind(FrontendApplicationContribution).toService(TraceViewerToolbarContribution);
     bind(TabBarToolbarContribution).toService(TraceViewerToolbarContribution);
+    bind(CommandContribution).toService(TraceViewerToolbarContribution);
 
     bind(TraceViewerWidget).toSelf();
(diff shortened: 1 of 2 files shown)


colin-grant-work:
@ankusht-work, would you mind taking a look?

mirsky-work:
@colin-grant-work thanks for the patch! I was able to verify the fix. @PatrickTasse, can you please take a look?
Also, looks like the Gitpod check is stuck, but I couldn't find a way to retrigger it, other than pushing an amended commit (with no changes).

colin-grant-work:
> Also, looks like the Gitpod check is stuck, but I couldn't find a way to retrigger it, other than pushing an amended commit (with no changes).

@mirsky-work, for better or worse my commit message had a typo, so an empty push was necessary anyway :-)

bhufmann (approved):
LGTM. No more error message: "ERROR Widgets added to the application shell must have a unique id property"

MatthewKhouzam (approved):
Looks fine to me!

PR #419, opened 2021-07-15
Implement navigation for XY charts

IbrahimFradj:
**Navigation implemented :** 

- **Zoom-in by following the mouse cursor :** "w","W" and "ctrl+mouse wheel up".
- **Zoom-out by following the mouse cursor :** "s", "S", and ""ctrl+mouse wheel down".
- **Zooming with range :** click right and drag.
-  **Panning left :** "a","A", "arrowLeft" and "shift+mouse wheel up".
-  **Panning right :**  "d","D", "arrowRight" and "shift+mouse wheel down".

Fixes theia-ide#389

**Here is the result**
zoom

### DIFF
--- a/packages/react-components/src/components/xy-output-component.tsx
+++ b/packages/react-components/src/components/xy-output-component.tsx
@@ -32,6 +32,7 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
     private currentColorIndex = 0;
     private colorMap: Map<string, number> = new Map();
     private lineChartRef: any;
+    private chartRef: any;
     private mouseIsDown = false;
     private positionXMove = 0;
     private posPixelSelect = 0;
@@ -57,6 +58,8 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
         document.removeEventListener('mouseup', this.endSelection);
     };
 
+    private preventDefaultHandler: ((event: WheelEvent) => void) | undefined;
+
     constructor(props: AbstractOutputProps) {
         super(props);
         this.state = {
@@ -72,6 +75,7 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
 
         this.afterChartDraw = this.afterChartDraw.bind(this);
         this.lineChartRef = React.createRef();
+        this.chartRef = React.createRef();
     }
 
     componentDidMount(): void {
@@ -115,6 +119,14 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
             this.updateXY();
         }
         if (this.lineChartRef.current) {
+            if (this.preventDefaultHandler === undefined) {
+                this.preventDefaultHandler = (event: WheelEvent) => {
+                    if (event.ctrlKey) {
+                        event.preventDefault();
+                    }
+                };
+                this.chartRef.current.addEventListener('wheel', this.preventDefaultHandler);
+            }
             this.lineChartRef.current.chartInstance.render();
         }
     }
@@ -172,7 +184,9 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
                     onWheel={event => this.onWheel(event)}
                     onMouseMove={event => this.onMouseMove(event)}
                     onMouseDown={event => this.onMouseDown(event)}
-                    style={{ height: this.props.style.height }}>
+                    style={{ height: this.props.style.height }}
+                    ref={this.chartRef}
+                    >
                     <Line
                         data={this.state.xyData}
                         height={parseInt(this.props.style.height.toString())}
@@ -362,6 +376,12 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
             }
             else if (wheel.deltaY > 0) {
                 this.pan(PAN_RIGHT);
+            }
+        } else if (wheel.ctrlKey) {
+            if (wheel.deltaY < 0) {
+                this.zoom(ZOOM_IN);
+            } else if (wheel.deltaY > 0) {
+                this.zoom(ZOOM_OUT);
             }
         }
     }
--- a/packages/react-components/src/components/xy-output-component.tsx
+++ b/packages/react-components/src/components/xy-output-component.tsx
@@ -23,6 +23,7 @@ type XYOuputState = AbstractOutputState & {
     xyData: any;
     columns: ColumnHeader[];
 };
+const RIGHT_CLICK_NUMBER = 2;
 const ZOOM_IN = true;
 const ZOOM_OUT = false;
 const PAN_LEFT = true;
@@ -35,14 +36,17 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
     private chartRef: any;
     private mouseIsDown = false;
     private positionXMove = 0;
+    private isRightClick = false;
     private posPixelSelect = 0;
+    private isMouseLeave = false;
+    private startPositionMouseRightClick = 0;
     private plugin = {
         afterDraw: (chartInstance: Chart, _easing: Chart.Easing, _options?: any) => { this.afterChartDraw(chartInstance); }
     };
     private updateSelection = (event: MouseEvent) => {
-        if (this.mouseIsDown && this.props.unitController.selectionRange) {
+        const scale = this.props.viewRange.getEnd() - this.props.viewRange.getstart();
+        if (this.mouseIsDown && this.props.unitController.selectionRange && !this.isRightClick) {
             const xStartPos = this.props.unitController.selectionRange.start;
-            const scale = this.props.viewRange.getEnd() - this.props.viewRange.getstart();
             let end = xStartPos + ((event.screenX - this.posPixelSelect) / this.lineChartRef.current.chartInstance.width) * scale;
             end = Math.min(Math.max(end, 0), this.props.unitController.absoluteRange);
             this.props.unitController.selectionRange = {
@@ -53,8 +57,12 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
     };
 
     private endSelection = () => {
+        if (this.isRightClick) {
+           this.applySelectionZoom();
+        } else {
+            document.removeEventListener('mousemove', this.updateSelection);
+        }
         this.mouseIsDown = false;
-        document.removeEventListener('mousemove', this.updateSelection);
         document.removeEventListener('mouseup', this.endSelection);
     };
 
@@ -70,7 +78,7 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
             collapsedNodes: [],
             orderedNodes: [],
             xyData: {},
-            columns: [{title: 'Name', sortable: true}]
+            columns: [{title: 'Name', sortable: true}],
         };
 
         this.afterChartDraw = this.afterChartDraw.bind(this);
@@ -183,6 +191,8 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
                     onKeyDown={event => this.onKeyDown(event)}
                     onWheel={event => this.onWheel(event)}
                     onMouseMove={event => this.onMouseMove(event)}
+                    onContextMenu={event => event.preventDefault()}
+                    onMouseLeave={event => this.onMouseLeave(event)}
                     onMouseDown={event => this.onMouseDown(event)}
                     style={{ height: this.props.style.height }}
                     ref={this.chartRef}
@@ -213,33 +223,58 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
         const ctx = chart.ctx;
         const xScale = (chart as any).scales['time-axis'];
         const ticks: number[] = xScale.ticks;
-        if (ctx && this.props.selectionRange) {
-            const min = Math.min(this.props.selectionRange.getstart(), this.props.selectionRange.getEnd());
-            const max = Math.max(this.props.selectionRange.getstart(), this.props.selectionRange.getEnd());
-            // If the selection is out of range
-            if (min > this.props.viewRange.getEnd() || max < this.props.viewRange.getstart()) {
-                return;
+        if (ctx) {
+            let start = 0;
+            if (this.props.selectionRange) {
+                const min = Math.min(this.props.selectionRange.getstart(), this.props.selectionRange.getEnd());
+                const max = Math.max(this.props.selectionRange.getstart(), this.props.selectionRange.getEnd());
+                const minValue = this.findNearestValue(min, ticks);
+                let minPixel = xScale.getPixelForValue(min, minValue);
+                const maxValue = this.findNearestValue(max, ticks);
+                let maxPixel = xScale.getPixelForValue(max, maxValue);
+                // In the case the selection is going out of bounds, the pixelValue needs to be in the displayed range.
+                if (maxPixel === 0) {
+                    maxPixel = chart.chartArea.right;
+                }
+                if (minPixel === 0) {
+                    minPixel = chart.chartArea.right;
+                }
+                ctx.strokeStyle = '#259fd8';
+                ctx.fillStyle = '#259fd8';
+                this.drawSelection(chart, minPixel, maxPixel);
             }
-            const minValue = this.findNearestValue(min, ticks);
-            const minPixel = xScale.getPixelForValue(min, minValue);
-            const maxValue = this.findNearestValue(max, ticks);
-            let maxPixel = xScale.getPixelForValue(max, maxValue);
-            // In the case the selection is going out of bounds, the pixelValue needs to be in the displayed range.
-            if (maxPixel === 0) {
-                maxPixel = chart.chartArea.right;
+            if (this.isRightClick) {
+                const offset = this.props.viewRange.getOffset() ?? 0;
+                start = this.startPositionMouseRightClick + offset;
+                const startValue = this.findNearestValue(start, ticks);
+                let startPixel = xScale.getPixelForValue(start, startValue);
+                const endPixel = this.positionXMove;
+                if (startPixel === 0) {
+                    startPixel = chart.chartArea.right;
+                }
+                ctx.strokeStyle = '#9f9f9f';
+                ctx.fillStyle = '#9f9f9f';
+                this.drawSelection(chart, startPixel, endPixel);
             }
+        }
+    }
+
+    private drawSelection(chart: Chart, startPixel: number, endPixel: number) {
+        const ctx = chart.ctx;
+        const minPixel = Math.min(startPixel, endPixel);
+        const maxPixel = Math.max(startPixel, endPixel);
+        if (ctx) {
             ctx.save();
 
             ctx.lineWidth = 1;
-            ctx.strokeStyle = '#259fd8';
             // Selection borders
-            if (min > this.props.viewRange.getstart()) {
+            if (startPixel > chart.chartArea.left) {
                 ctx.beginPath();
                 ctx.moveTo(minPixel, 0);
                 ctx.lineTo(minPixel, chart.chartArea.bottom);
                 ctx.stroke();
             }
-            if (max < this.props.viewRange.getEnd()) {
+            if (endPixel < this.props.viewRange.getEnd()) {
                 ctx.beginPath();
                 ctx.moveTo(maxPixel, 0);
                 ctx.lineTo(maxPixel, chart.chartArea.bottom);
@@ -247,9 +282,7 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
             }
             // Selection fill
             ctx.globalAlpha = 0.2;
-            ctx.fillStyle = '#259fd8';
             ctx.fillRect(minPixel, 0, maxPixel - minPixel, chart.chartArea.bottom);
-
             ctx.restore();
         }
     }
@@ -299,14 +332,21 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
     }
 
     private onMouseDown(event: React.MouseEvent<HTMLDivElement, MouseEvent>) {
+        this.isMouseLeave = false;
         this.mouseIsDown = true;
         this.posPixelSelect = event.nativeEvent.screenX;
         const startTime = this.getTimeX(event.nativeEvent.offsetX);
-        this.props.unitController.selectionRange = {
-            start: startTime,
-            end: startTime
-        };
-        document.addEventListener('mousemove', this.updateSelection);
+        if (event.button === RIGHT_CLICK_NUMBER) {
+            this.isRightClick = true;
+            this.startPositionMouseRightClick = startTime;
+        } else {
+            this.isRightClick = false;
+            this.props.unitController.selectionRange = {
+                start: startTime,
+                end: startTime
+            };
+            document.addEventListener('mousemove', this.updateSelection);
+        }
         document.addEventListener('mouseup', this.endSelection);
     }
 
@@ -370,6 +410,7 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
     }
 
     private onWheel(wheel: React.WheelEvent) {
+        this.isMouseLeave = false;
         if (wheel.shiftKey) {
             if (wheel.deltaY < 0) {
                 this.pan(PAN_LEFT);
@@ -396,31 +437,52 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
 
     private onMouseMove(event: React.MouseEvent) {
         this.positionXMove = event.nativeEvent.offsetX;
+        this.isMouseLeave = false;
+        if (this.mouseIsDown && this.isRightClick) {
+            this.forceUpdate();
+        }
+    }
+
+    private onMouseLeave(event: React.MouseEvent) {
+        this.isMouseLeave = true;
+        if (this.isRightClick) {
+            this.positionXMove = Math.max(0, Math.min(event.nativeEvent.offsetX, this.lineChartRef.current.chartInstance.width));
+            this.forceUpdate();
+        }
+    }
+
+    private applySelectionZoom() {
+        const newStartRange = this.startPositionMouseRightClick;
+        const newEndRange = this.getTimeX(this.positionXMove);
+        this.updateRange(newStartRange, newEndRange);
+        this.isRightClick = false;
     }
 
     private onKeyDown(key: React.KeyboardEvent) {
-        switch (key.key) {
-            case 'W':
-            case 'w': {
-                this.zoom(ZOOM_IN);
-                break;
-            }
-            case 'S':
-            case 's': {
-                this.zoom(ZOOM_OUT);
-                break;
-            }
-            case 'A':
-            case 'a':
-            case 'ArrowLeft': {
-                this.pan(PAN_LEFT);
-                break;
-            }
-            case 'D':
-            case 'd':
-            case 'ArrowRight': {
-                this.pan(PAN_RIGHT);
-                break;
+        if (!this.isMouseLeave) {
+            switch (key.key) {
+                case 'W':
+                case 'w': {
+                    this.zoom(ZOOM_IN);
+                    break;
+                }
+                case 'S':
+                case 's': {
+                    this.zoom(ZOOM_OUT);
+                    break;
+                }
+                case 'A':
+                case 'a':
+                case 'ArrowLeft': {
+                    this.pan(PAN_LEFT);
+                    break;
+                }
+                case 'D':
+                case 'd':
+                case 'ArrowRight': {
+                    this.pan(PAN_RIGHT);
+                    break;
+                }
             }
         }
     }
--- a/packages/react-components/src/components/xy-output-component.tsx
+++ b/packages/react-components/src/components/xy-output-component.tsx
@@ -23,13 +23,17 @@ type XYOuputState = AbstractOutputState & {
     xyData: any;
     columns: ColumnHeader[];
 };
+const ZOOM_IN = true;
+const ZOOM_OUT = false;
+const PAN_LEFT = true;
+const PAN_RIGHT = false;
 
 export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutputProps, XYOuputState> {
     private currentColorIndex = 0;
     private colorMap: Map<string, number> = new Map();
-
     private lineChartRef: any;
     private mouseIsDown = false;
+    private positionXMove = 0;
     private posPixelSelect = 0;
     private plugin = {
         afterDraw: (chartInstance: Chart, _easing: Chart.Easing, _options?: any) => { this.afterChartDraw(chartInstance); }
@@ -163,7 +167,12 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
         // width={this.props.style.chartWidth}
         return <React.Fragment>
             {this.state.outputStatus === ResponseStatus.COMPLETED ?
-                <div id='xy-main' onMouseDown={event => this.beginSelection(event)} style={{ height: this.props.style.height }} >
+                <div id='xy-main' tabIndex={0}
+                    onKeyDown={event => this.onKeyDown(event)}
+                    onWheel={event => this.onWheel(event)}
+                    onMouseMove={event => this.onMouseMove(event)}
+                    onMouseDown={event => this.onMouseDown(event)}
+                    style={{ height: this.props.style.height }}>
                     <Line
                         data={this.state.xyData}
                         height={parseInt(this.props.style.height.toString())}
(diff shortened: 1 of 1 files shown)


packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -284,10 +287,128 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
     }
 
+    private zoomIn(diffStartMouse: number, diffEndMouse: number, percentRange: number) {
+        if (this.isRangeMove) {
+            this.props.unitController.viewRange = {
+                start: this.props.unitController.viewRange.start + percentRange,
+                end: this.props.unitController.viewRange.end - percentRange
+            };
+        } else {
+            const newStartRange = this.xStartPos - (diffStartMouse - (diffStartMouse * 0.1));
```
MatthewKhouzam, line 302: extract .1 as a constant

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -284,10 +287,128 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
+        const diffEndMouse = this.props.unitController.viewRange.end - this.xStartPos;
+        const percentRange = this.props.unitController.viewRangeLength * 0.1;
+        if (this.xStartPos >= this.props.unitController.viewRange.end || this.xStartPos <= this.props.unitController.viewRange.start) {
+            this.isRangeMove = true;
+        } else {
+            this.isRangeMove = false;
+        }
+        switch (key.key) {
+            case 'W':
+            case '+':
```
MatthewKhouzam, line 388: please add '=' as it is no shift '+'

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -284,10 +287,128 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
+            case 'w': {
+                this.zoomIn(diffStartMouse, diffEndMouse, percentRange);
+                break;
+            }
+            case 'S':
+            case '-':
```
MatthewKhouzam, line 394: add '_'

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -284,10 +287,128 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
+            case 's': {
+                this.zoomOut(diffStartMouse, diffEndMouse, percentRange);
+                break;
+            }
+            case 'A':
```
MatthewKhouzam, line 399: question... not a suggestion: should we add right arrow and left arrow too?

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -284,10 +287,128 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
+                start: Math.max(0, newStartRange),
+                end: Math.min(newEndRange, this.props.unitController.absoluteRange)
+            };
+        }
+    }
+
+    private shiftLeft(percentRange: number) {
+        let startRange = this.props.unitController.viewRange.start - percentRange;
+        let endRange = this.props.unitController.viewRange.end - percentRange;
+        if (startRange <= 0) {
```
MatthewKhouzam, line 331: make a helper function

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -284,10 +287,128 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
+                end: endRange
+            };
+        }
+
+    }
+
+    private shiftRight(percentRange: number) {
+        let startRange = this.props.unitController.viewRange.start + percentRange;
+        let endRange = this.props.unitController.viewRange.end + percentRange;
+        if (endRange >= this.props.unitController.absoluteRange) {
```
MatthewKhouzam, line 350: use the helper function here

MatthewKhouzam (commented):
Commenting, these are requested changes but I don't want to slow you down.

packages/react-components/src/components/xy-output-component.tsx
MatthewKhouzam: Holding down on 'W' or 'w' should zoom on the cursor, + on the center.

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -284,10 +287,128 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
+    private zoomOut(diffStartMouse: number, diffEndMouse: number, percentRange: number) {
+        if (this.isRangeMove) {
+            this.props.unitController.viewRange = {
+                start: this.props.unitController.viewRange.start - percentRange,
+                end: this.props.unitController.viewRange.end + percentRange
+            };
+        }
+        else {
+            const newStartRange = this.xStartPos - (diffStartMouse + (diffStartMouse * 0.1));
+            const newEndRange = this.xStartPos + (diffEndMouse + (diffEndMouse * 0.1));
```
muddana-satish, line 320: Does xy chart zooming and timegraph chart zooming behave the same way? I think the zoom factor used is different. 
Refer to this: https://github.com/theia-ide/timeline-chart/blob/2d3f4d88d8f5f214b5b76136ccdfa25d2d645297/timeline-chart/src/layer/time-graph-chart.ts#L81

IbrahimFradj: Should I use the same zoom factor? When I tried to zoom in and out in the timeline view, the zoom in and out is sometimes larger than I want.

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -61,7 +63,7 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
             collapsedNodes: [],
             orderedNodes: [],
             xyData: {},
-            columns: [{title: 'Name', sortable: true}]
+            columns: [{ title: 'Name', sortable: true }]
```
ebugden, line 66: There seem to be a lot of lines in the diff that are whitespace changes unrelated to adding navigation to XY charts. Unrelated code cleanup changes should at least be in a separate commit (and probably in a different PR). Including unrelated changes adds a lot of noise to the diff and makes the code more difficult to review.

When creating commits with your changes, I recommend reading over the diff and manually staging functional changes using a GUI git tool:
* If the internal git tool in your IDE doesn't allow staging diff sections or individual lines, you can use gitk to create commits.
* First, create a commit with the functional changes. 
* Then, any whitespace or style changes leftover can be committed separately. One way to create a separate PR for leftover style changes would be to "git stash" them, create a new branch from master, then reapply the changes (I think it's "git stash pop").

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -171,14 +176,14 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
-                {(
+                    {(
                         <i
                             className='fa fa-refresh fa-spin'
                             style={{ marginRight: '5px' }}
                         />
                     )}
                     {
-                    'Analysis running'
+                        'Analysis running'
```
ebugden, line 186: Unrelated whitespace change?

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -171,14 +176,14 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
                     </Line>
                 </div> :
                 <div className='analysis-running'>
-                {(
+                    {(
```
ebugden, line 179: Unrelated whitespace change?

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -252,7 +257,7 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
                 newList = newList.concat(id);
             }
         });
-        this.setState({checkedSeries: newList});
+        this.setState({ checkedSeries: newList });
```
ebugden, line 260: Unrelated whitespace change?

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -266,20 +271,17 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
             newList = newList.concat(id);
         }
         const orderedIds = getAllExpandedNodeIds(nodes, newList);
-        this.setState({collapsedNodes: newList, orderedNodes: orderedIds});
+        this.setState({ collapsedNodes: newList, orderedNodes: orderedIds });
```
ebugden, line 274: Unrelated whitespace change?

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -266,20 +271,17 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
     }
 
     private onOrderChange(ids: number[]) {
-        this.setState({orderedNodes: ids});
+        this.setState({ orderedNodes: ids });
```
ebugden, line 278: Unrelated whitespace change?

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -156,12 +158,15 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
                 yAxes: [{ display: false }]
             },
             animation: { duration: 0 },
-            events: [ 'mousedown' ],
+            events: ['mousedown'],
```
ebugden, line 161: Unrelated whitespace change?

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -82,12 +84,12 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
                     headers.forEach(header => {
-                        columns.push({title: header.name, sortable: true, tooltip: header.tooltip});
+                        columns.push({ title: header.name, sortable: true, tooltip: header.tooltip });
                     });
                 } else {
-                    columns.push({title: 'Name', sortable: true});
+                    columns.push({ title: 'Name', sortable: true });
                 }
-                columns.push({title: 'Legend', sortable: false});
+                columns.push({ title: 'Legend', sortable: false });
```
ebugden, line 92: Unrelated whitespace change?

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -82,12 +84,12 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
                 const columns = [];
                 if (headers && headers.length > 0) {
                     headers.forEach(header => {
-                        columns.push({title: header.name, sortable: true, tooltip: header.tooltip});
+                        columns.push({ title: header.name, sortable: true, tooltip: header.tooltip });
                     });
                 } else {
-                    columns.push({title: 'Name', sortable: true});
+                    columns.push({ title: 'Name', sortable: true });
```
ebugden, line 90: Unrelated whitespace change?

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -82,12 +84,12 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
                 const columns = [];
                 if (headers && headers.length > 0) {
                     headers.forEach(header => {
-                        columns.push({title: header.name, sortable: true, tooltip: header.tooltip});
+                        columns.push({ title: header.name, sortable: true, tooltip: header.tooltip });
```
ebugden, line 87: Unrelated whitespace change?

ebugden (commented):
I have a lot of feelings about the amount of unrelated whitespace changes mixed in with the commit 😅 I think this is worth correcting before merging for learning purposes (creating clean and specific commits) and also to set a good style example with the commits that are merged.

That being said, the code seems to do what it claims to do! Thanks @IbrahimFradj ☀️

packages/react-components/src/components/xy-output-component.tsx
muddana-satish: Okay. I am not sure about that. It would be nice to keep the zooming consistent across the charts. I think we can create a new issue for that.

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -33,9 +33,9 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
     private lineChartRef: any;
     private mouseIsDown = false;
     private positionZoomMouse = 0;
-    private positionZommkeyboard = 0;
+    private positionZomCenter = 0;
```
ebugden, line 36: Zom? "positionZomCenter"

IbrahimFradj: this is the position from where I want to zoom . Because when i zoom with "+" is not the same position when i zoom with "w".
The zoom with "+" follow the center of the view and the zoom with " w" it follows the mouse cursor

ebugden: I mean why is it "Zom" rather than "Zoom"?

IbrahimFradj: ah sorry thank you i didn't notice

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -61,11 +83,16 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
+            columns: [{title: 'Name', sortable: true}],
+            zoomSelectionRangeStart: 0,
+            zoomSelectionRangeEnd: 0,
         };
 
         this.afterChartDraw = this.afterChartDraw.bind(this);
         this.lineChartRef = React.createRef();
+        this.instanceId = XYOutputComponent.instanceCounter++;
+        signalManager().on(Signals.TIMEGRAPH_ZOOMED, this.onTimeGraphZoomed);
+        signalManager().on(Signals.TIMEGRAPH_RESET, this.onTimeGraphReset);
```
muddana-satish, line 95: Can we listen to these signals in the parent class?

IbrahimFradj: I tried and I could not put the signals in AbstractTreeOutputComponent

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -288,6 +291,191 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
+    private getPosition(event: React.MouseEvent): number {
+        const offset = this.props.viewRange.getOffset() ?? 0;
+        const scale = this.props.viewRange.getEnd() - this.props.viewRange.getstart();
+        const xPos = this.props.viewRange.getstart() - offset +
+            (event.nativeEvent.offsetX / this.lineChartRef.current.chartInstance.width) * scale;
+        return xPos;
+    }
+
+    private MoveEvent(event: React.MouseEvent) {
+        if (!this.mouseIsDown) {
```
PatrickTasse: Should update the mouse position unconditionally, otherwise for example if you select a range with mouse then W/S, it zooms on the time of the selection start.

I would also only store the mouse x-coordinate, and only calculate the corresponding time when it's used to zoom, that way you don't have to keep updating the mouse 'time' after panning etc..

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -288,6 +291,191 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
+        const diffStartMouse = this.positionZoomMouse - this.props.unitController.viewRange.start;
+        const percentMousePosition = diffStartMouse / this.props.unitController.viewRangeLength;
+        return percentMousePosition;
+    }
+
+    private zoomKey(key: React.KeyboardEvent) {
+        switch (key.key) {
+            case '+':
+            case '=': {
+                this.isZoomCenter = true;
```
PatrickTasse: Not a good idea to set an instance variable so that it can be used by a method you call, it should just be a method parameter.

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -288,6 +291,191 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
+                if (this.props.unitController.viewRangeLength >= 1) {
+                    this.zoomIn(this.initializePosition()[0], this.initializePosition()[1], this.mousePositionPercent());
```
PatrickTasse: These parameters could be calculated from within the called method?

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -288,6 +291,191 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
+        let position = 0;
+        if (this.isZoomCenter) {
+            this.positionZoomCenter = (this.props.unitController.viewRange.end + this.props.unitController.viewRange.start) / 2;
+            position = this.positionZoomCenter;
+        } else {
+            position = this.positionZoomMouse;
+        }
+        const startPosition = position - this.props.unitController.viewRange.start;
+        const endPosition = this.props.unitController.viewRange.end - position;
+        return [startPosition, endPosition];
```
PatrickTasse, line 568: Returns an array that is [duration between window start and mouse, duration between mouse and window end]. That is a very strange API, I would suspect another way would be more understandable...

Also maybe would be cleaner to pass objects instead of array of values.

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -288,6 +291,191 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
+                start: this.NewstartEndRangePanning(right, percentRange)[0],
+                end: this.NewstartEndRangePanning(right, percentRange)[1]
+            };
+        }
+    }
+
+    private shiftLeft() {
+        const numberLeft = -1;
+        const distance = ( this.props.unitController.viewRange.start);
+        this.shift(numberLeft, distance);
```
PatrickTasse, line 479: Would be cleaner to pass a boolean and these two values could be calculated within the called method instead?

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -288,6 +291,191 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
+    }
+
+    private shiftRight() {
+        const numberRight = 1;
+        const distance = (this.props.unitController.absoluteRange - this.props.unitController.viewRange.end);
+        this.shift(numberRight, distance);
+    }
+
+    private wheelEvent(wheel: React.WheelEvent) {
+        if (wheel.shiftKey) {
```
PatrickTasse, line 490: We probably want to do this if only Shift is pressed, mouse wheel with Ctrl+Shift down has weird effect.

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -288,6 +291,191 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
+    }
+
+    private updatePositionAfterPann(right: number, distance: number){
+        this.positionZoomCenter = this.positionZoomCenter + right * distance;
+        this.positionZoomMouse = this.positionZoomMouse + right * distance;
+    }
+
+    private shift(right: number, distance: number) {
+        const percentRange = this.props.unitController.viewRangeLength * TEN_PERCENT;
+        if (this.NewstartEndRangePanning(right, percentRange)[0] < 0) {
```
PatrickTasse, line 455: I would just calculate the expected new range once and handle the limits, not call the same method repeatedly.

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -288,6 +291,191 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
+            case '+':
+            case '=': {
+                this.isZoomCenter = true;
+                if (this.props.unitController.viewRangeLength >= 1) {
+                    this.zoomIn(this.initializePosition()[0], this.initializePosition()[1], this.mousePositionPercent());
+                }
+                break;
+            }
+            case 'W':
+            case 'w': {
```
PatrickTasse: I wonder if we support arrow left/right to pan, if we should also support arrow up/down to zoom?

PatrickTasse (changes requested):

packages/react-components/src/components/xy-output-component.tsx
IbrahimFradj: I asked this question but the answer to this question is that we left the up and down arrows to scroll the page.

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -288,6 +291,191 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
+        }
+        this.updateRange(startPosition, endPosition, percentMousePosition);
+    }
+
+    private zoomOut(startPosition: number, endPosition: number, percentMousePosition: number) {
+        const zoom = -1;
+        this.updateRange( this.newRangeZoom(zoom, startPosition, endPosition)[0], this.newRangeZoom(zoom, startPosition, endPosition)[1],percentMousePosition);
+    }
+
+    private NewstartEndRangePanning(shift: number, percentRange: number): number[] {
```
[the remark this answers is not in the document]
IbrahimFradj: yes thank you

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -288,6 +291,191 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
+            if (wheel.deltaY < 0) {
+                this.shiftLeft();
+            }
+            else if (wheel.deltaY > 0) {
+                this.shiftRight();
+            }
+        }
+    }
+
+    private getPosition(event: React.MouseEvent): number {
```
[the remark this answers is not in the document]
IbrahimFradj: i changed it with getTimeX is it okay ?

packages/react-components/src/components/xy-output-component.tsx
IbrahimFradj: yes I made these changes in another commit "Zoom in and out by click right drag "when I had this problem.

packages/react-components/src/components/xy-output-component.tsx
IbrahimFradj: yes

packages/react-components/src/components/xy-output-component.tsx
IbrahimFradj: i didn't have a weird effect when i tried this. it took always the shift

mirsky-work:
Thanks for this nice improvement!
I did notice some misalignment issues between the timeline and XY charts. For example when panning the timeline chart at a high zoom level using 'a' or 'd' shortcuts, the selection range on the XY chart dances around and sometimes ends up misaligned with the selection range on the timeline chart. There are no issues the other way around, and the way to correct the misalignment is to pan the XY chart. See the gif attached.

xy-chart-nav

PatrickTasse:
> I did notice some misalignment issues between the timeline and XY charts. For example when panning the timeline chart at a high zoom level using 'a' or 'd' shortcuts, the selection range on the XY chart dances around and sometimes ends up misaligned with the selection range on the timeline chart.

When the view range is changed rapidly in succession, each small step triggers fetching a new model from the trace-server over the network. These requests are all asynchronous and processed in separate concurrent threads in the trace-server, and the responses do not necessarily come back in the same order that the requests were sent. So if that last response received is not for the last view range that was requested, it can cause a misalignment. This is a design flaw that should be addressed, probably in a separate issue.

IbrahimFradj:
> I did notice some misalignment issues between the timeline and XY charts. For example when panning the timeline chart at a high zoom level using 'a' or 'd' shortcuts, the selection range on the XY chart dances around and sometimes ends up misaligned with the selection range on the timeline chart. There are no issues the other way around, and the way to correct the misalignment is to pan the XY chart. See the gif attached.

I tried what you did in the master branch and found that there was the same issue. I will create an issue in github for this problem. Thank you for your comment
**Here is the result in master branch :**

bug master pan

mirsky-work:
> I tried what you did in the master branch and found that there was the same issue. I will create an issue in github for this problem. Thank you for your comment

Thanks for verifying this and reporting #453!

mirsky-work:
Sorry, closed by mistake (Monday morning...).

PatrickTasse (approved):

PatrickTasse (commented):
The commit message mentions support for "+" and "-". This now will only come later.

The lines in commit message should wrap after 72 chars.

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -273,21 +283,123 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
+            let zoomNumber = 1;
+            const position = this.getTimeX(this.positionXMove);
+            if (!iszoomIn) {
+                zoomNumber = -1;
+            }
+            const startPos = position - this.props.unitController.viewRange.start;
+            const endPos = this.props.unitController.viewRange.end - position;
+            newStartRange = position - (startPos - zoomNumber * (startPos * TEN_PERCENT));
+            newEndRange = position + (endPos - zoomNumber * (endPos * TEN_PERCENT));
+            this.updateRange(newStartRange, newEndRange);
```
PatrickTasse, line 395: This creates a strange behavior in the following use case:
- Zoom to half the trace range
- Pan to beginning of trace
- Try to zoom out with 'S" while the mouse is very close to the right edge.

What happens is that the window range start is beyond the start time (< 0) and gets snapped back to 0 in time-graph-unit-controller. Then the effect is a very slow zoom out step.

What I would suggest is to calculate the new view range length based on zoom in/out (capped to absolute range), then apply this new window range centered on the mouse position. After this, snap the new window range to the bounds [0, absolute range] like it's done below for the pan case.

PatrickTasse (changes requested):

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -273,21 +283,125 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
+            end: rangeEnd
+        };
+    }
+
+    private zoom(iszoomIn: boolean) {
+        if (this.props.unitController.viewRangeLength >= 1) {
+            let newStartRange = 0;
+            let newEndRange = 0;
+            const zoomNumber = iszoomIn ? 1 : -1;
+                const position = this.getTimeX(this.positionXMove);
```
PatrickTasse, line 386: Indentation

PatrickTasse (changes requested):

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -273,21 +283,125 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
+                const startPos = position - this.props.unitController.viewRange.start;
+                const endPos = this.props.unitController.viewRange.end - position;
+                newStartRange = position - (startPos - zoomNumber * (startPos * TEN_PERCENT));
+                newEndRange = position + (endPos - zoomNumber * (endPos * TEN_PERCENT));
```
PatrickTasse, line 390: Seems like the end result of this code is that range is changed by a factor of 0.9 or 1.1.
But to be the mathematical inverse, the factor should be 0.9 or 1/0.9=1.11111111111.

PatrickTasse (changes requested):

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -114,10 +119,33 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
+    componentWillUnmount(): void {
+        if (this.preventDefaultHandler !== undefined) {
+            const xyMainElement = document.getElementById(this.getUniqueDivId());
+            if (xyMainElement) {
+                xyMainElement.removeEventListener('wheel', this.preventDefaultHandler);
+            }
+            this.preventDefaultHandler = undefined;
+        }
+    }
+
```
PatrickTasse, line 161: The timegraph is able to prevent the mouse wheel document zoom without resorting to adding/removing the handler and needing uniqueDivId. I think it would be worth to try and simplify the code by using a similar strategy.

PatrickTasse (changes requested):

packages/react-components/src/components/xy-output-component.tsx
Rodrigoplp-work: Patrick suggested using `React.createRef()` for this, and I agree. It could be implemented by:

1. Adding `private chartRef: any;` below line 41;

2. Adding `this.chartRef = React.createRef();` below line 91;

3. Replacing most of the code block above with `this.chartRef.current.addEventListener('wheel', this.preventDefaultHandler);`. No need for `xyMainElement`.

Also, in this case it seems that the eventListener is being cleared by the garbage collector. No need to remove it manually.

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -113,6 +120,14 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
         }
         if (this.lineChartRef.current) {
+            if (this.preventDefaultHandler === undefined) {
+                this.preventDefaultHandler = (event: WheelEvent) => {
+                    if (event.ctrlKey) {
+                        event.preventDefault();
+                    }
+                };
+                this.chartRef.current.addEventListener('wheel', this.preventDefaultHandler);
+            }
```
PatrickTasse, line 143: Could the listener be added in componentDidMount instead?

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -165,12 +180,14 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
         // width={this.props.style.chartWidth}
         return <React.Fragment>
             {this.state.outputStatus === ResponseStatus.COMPLETED ?
-                <div id='xy-main' tabIndex={0}
+                <div id={this.getUniqueDivId()} tabIndex={0}
```
PatrickTasse, line 196: I am not sure if this change is required anymore now that we have the chartRef. Could you check if it works without it?

PatrickTasse (changes requested):

packages/react-components/src/components/xy-output-component.tsx
IbrahimFradj: I tried it but it didn't work. the view didn't want to display.

packages/react-components/src/components/xy-output-component.tsx
IbrahimFradj: i just tried now and it works.

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -394,31 +443,69 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
+
+    private updateSelectionZoom() {
+        const xStartPos = this.state.zoomSelectionRangeStart;
+        this.setState({
+            zoomSelectionRangeStart: xStartPos,
+            zoomSelectionRangeEnd: this.getTimeX(this.positionXMove)
+        });
+    }
+
+    private resetSelectionZoom() {
```
PatrickTasse, line 475: Since it actually does the zoom, I suggest calling it applySelectionZoom().

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -394,31 +443,69 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
             }
-            case 'D':
-            case 'd':
-            case 'ArrowRight': {
-                this.pan(PAN_RIGHT);
-                break;
+        }
+    }
+
+    private updateSelectionZoom() {
```
PatrickTasse, line 467: Since you already keep track of the mouse x position, you could use that to draw the selection range, then you only need to convert it to a time when applying the zoom selection.

This way you wouldn't need to update the zoom end time as you pan/zoom.

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -320,33 +355,42 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
+                newStartRange = Math.max(newStartRange, 0);
+                newEndRange = Math.min(newEndRange, this.props.unitController.absoluteRange);
             }
-            newStartRange = Math.max(newStartRange, 0);
-            newEndRange = Math.min(newEndRange, this.props.unitController.absoluteRange);
             this.updateRange(newStartRange, newEndRange);
         }
     }
 
     private pan(panLeft: boolean) {
```
PatrickTasse, line 386: Changes in this method might not be necessary if you just use the x position to draw the zoom selection.

PatrickTasse (changes requested):

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -394,31 +435,56 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
-        switch (key.key) {
-            case 'W':
-            case 'w': {
-                this.zoom(ZOOM_IN);
-                break;
+    private onMouseLeave(event: React.MouseEvent) {
+        this.isMouseLeave = true;
+        if (this.isRightClick) {
+            if (this.getTimeX(event.nativeEvent.offsetX) < this.props.unitController.viewRange.start) {
+                this.positionXMove = this.props.unitController.viewRange.start;
```
PatrickTasse, line 448: Mixing timestamps with x-coordinates...

PatrickTasse (changes requested):

PatrickTasse (approved):

bhufmann (approved):
LGTM. Great improvement. Thank you very much!

PatrickTasse (approved):

PR #431, opened 2021-07-27
Support for searching for previous event in the Events Table

### LINKED ISSUE #412: Support searching for previous events in the events table
The current event table implementation only supports searching for next events, but searching for previous events is not support. A button for searching next exists in the search cells as well as keyboard shortcut ENTER is implemented for searching next.

- Add a previous button in column search cell right beside the next button
- Use keyboard short-cut SHIFT+ENTER to search previous event

The trace server protocol needs to be updated for that as well the trace server.

bhufmann:
- Add a 'previous' button in the column search cells right beside the
'next' button
- Use keyboard short-cut SHIFT+ENTER to search previous event

fixes #412

### DIFF
--- a/packages/react-components/src/components/table-output-component.tsx
+++ b/packages/react-components/src/components/table-output-component.tsx
@@ -24,6 +24,11 @@ type TableOutputProps = AbstractOutputProps & {
     tableWidth?: string;
 };
 
+enum Direction {
+    NEXT,
+    PREVIOUS
+}
+
 export class TableOutputComponent extends AbstractOutputComponent<TableOutputProps, TableOuputState> {
     private debugMode = false;
     private columnIds: Array<number> = [];
@@ -83,7 +88,7 @@ export class TableOutputComponent extends AbstractOutputComponent<TableOutputPro
         this.onModelUpdated = this.onModelUpdated.bind(this);
         this.onKeyDown = this.onKeyDown.bind(this);
         this.searchEvents = this.searchEvents.bind(this);
-        this.findNextMatchedEvent = this.findNextMatchedEvent.bind(this);
+        this.findMatchedEvent = this.findMatchedEvent.bind(this);
     }
 
     renderMainArea(): React.ReactNode {
@@ -314,7 +319,8 @@ export class TableOutputComponent extends AbstractOutputComponent<TableOutputPro
                     floatingFilterComponentParams: {
                         suppressFilterButton: true,
                         onFilterChange: this.searchEvents,
-                        onclickNext: this.findNextMatchedEvent,
+                        onclickNext: () => this.findMatchedEvent(Direction.NEXT),
+                        onclickPrevious: () =>  this.findMatchedEvent(Direction.PREVIOUS),
                         colName: columnHeader.id.toString()
                     },
                     icons: {
@@ -409,7 +415,7 @@ export class TableOutputComponent extends AbstractOutputComponent<TableOutputPro
         return lines[0].index;
     }
 
-    private fetchAdditionalParams(isFiltered = false): ({ [key: string]: any }) {
+    private fetchAdditionalParams(direction?: Direction): ({ [key: string]: any }) {
         let additionalParams: { [key: string]: any } = {};
         const filterObj: { [key: number]: string } = {};
         if (this.filterModel) {
@@ -419,8 +425,10 @@ export class TableOutputComponent extends AbstractOutputComponent<TableOutputPro
             });
             additionalParams = {
                 ['table_search_expressions']: filterObj,
-                ['isFiltered']: isFiltered
             };
+            if (direction !== undefined) {
+                additionalParams['table_search_direction'] = Direction[direction];
+            }
         }
         return additionalParams;
 
@@ -490,11 +498,11 @@ export class TableOutputComponent extends AbstractOutputComponent<TableOutputPro
         }
     }
 
-    private async findNextMatchIndex(currRowIndex: number) {
+    private async findMatchIndex(currRowIndex: number, direction = Direction.NEXT) {
         const traceUUID = this.props.traceId;
         const tspClient = this.props.tspClient;
         const outputId = this.props.outputDescriptor.id;
-        const additionalParams = this.fetchAdditionalParams(true);
+        const additionalParams = this.fetchAdditionalParams(direction);
         const tspClientResponse = await tspClient.fetchTableLines(traceUUID, outputId, QueryHelper.tableQuery(this.columnIds, currRowIndex, 1, additionalParams));
         const lineResponse = tspClientResponse.getModel();
         if (!tspClientResponse.isOk() || !lineResponse) {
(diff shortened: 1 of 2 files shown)


packages/react-components/src/components/table-output-component.tsx
```diff
@@ -421,6 +428,9 @@ export class TableOutputComponent extends AbstractOutputComponent<TableOutputPro
                 ['table_search_expressions']: filterObj,
                 ['isFiltered']: isFiltered
             };
+            if (direction !== undefined) {
+                additionalParams['table_search_direction'] = direction.toString;
```
PatrickTasse, line 432: This actually puts the toString function reference in the map ;) it's missing () after toString...

PatrickTasse (changes requested):

packages/react-components/src/components/table-output-component.tsx
bhufmann: good catch

packages/react-components/src/components/table-output-component.tsx
```diff
@@ -490,11 +500,11 @@ export class TableOutputComponent extends AbstractOutputComponent<TableOutputPro
         }
     }
 
-    private async findNextMatchIndex(currRowIndex: number) {
+    private async findMatchIndex(currRowIndex: number, direction = Direction.NEXT) {
```
PatrickTasse, line 503: This isn't working as expected... Direction.NEXT is actually the number 0 and Direction.PREVIOUS is the number 1. So later the toString() in the map puts "0" or "1"...

PatrickTasse (changes requested):

packages/react-components/src/components/table-output-component.tsx
bhufmann: Direction[direction] seems to return the string

PatrickTasse (approved):

bhufmann:
@muddana-satish could you please review this change? The latest nightly build has the corresponding back-end code implementation.

https://download.eclipse.org/tracecompass.incubator/trace-server/rcp/

packages/react-components/src/components/table-output-component.tsx
```diff
@@ -528,7 +538,42 @@ export class TableOutputComponent extends AbstractOutputComponent<TableOutputPro
...
+                    this.handleRowSelectionChange();
+                    rowNode.setSelected(true);
+                    isFound = true;
+                }
+            });
+
+            if (isFound) {
+                return;
+            }
+
```
muddana-satish, line 575: findPreviousMatchedEvent and findNextMatchedEvent share a lot of things. Can we reuse the logic?

muddana-satish (dismissed):
Looks good to me.

I don't know if we want to fix it in this but we no longer need the 'isFiltered' flag. Also we don't need additional information from the api like 'tags'.

packages/react-components/src/components/table-output-component.tsx
bhufmann: I'll keep the tags API because there might be other client implementations than the theia-trace-extension that relies on the server side filtering. 

I'll remove the code for 'isFiltered' flag since the server is updated. I'll also try to minimize the code duplication findPreviousMatchedEvent and findNextMatchedEvent.

bhufmann:
I updated the patch according to the review comments. 

I also noticed a general bug with the searching. The AgGrid keeps a cache of size maxBlocksInCache*cacheBlockSize (in our case (5*200)=1000), where each block of cacheBlockSize events are fetched from the back-end. These 5 blocks, however, are not necessarily consecutive blocks, it depends on user's navigation. In the search feature, where we do this.gridApi.forEachNode(...) to move to the next match, the whole cache is looped through. It could happen that the first matched event in the cache is not inside the current block and the table moves somewhere else in the table where the next matched event in the cache is found. I corrected that in this PR as well.

packages/react-components/src/components/table-output-component.tsx
```diff
@@ -508,13 +516,28 @@ export class TableOutputComponent extends AbstractOutputComponent<TableOutputPro
...
     }
 
-    private async findNextMatchedEvent() {
+    private async findMatchedEvent(direction: Direction) {
         let isFound = false;
         if (this.gridApi) {
             this.gridApi.deselectAll();
+            let rowNodes: RowNode[] = [];
+            let currRowIndex = this.selectStartIndex === -1 ? this.gridApi?.getLastDisplayedRow() + 1 : this.selectStartIndex;
+            const blockSize = this.props.cacheBlockSize ? this.props.cacheBlockSize : 200;
```
PatrickTasse, line 525: Just checking, will the 'current block' fix work correctly if the cachBlockSize is undefined and we use 200?

packages/react-components/src/components/table-output-component.tsx
```diff
@@ -508,13 +516,28 @@ export class TableOutputComponent extends AbstractOutputComponent<TableOutputPro
...
+                    rowNodes.push(rowNode);
+                }
+            });
+            if (direction === Direction.PREVIOUS) {
+                rowNodes = rowNodes.reverse();
+            }
+            rowNodes.forEach(rowNode => {
+                if ((rowNode.data && rowNode.data['isMatched'] && !isFound) &&
+                    (((direction === Direction.PREVIOUS) &&
+                        (rowNode.rowIndex < Math.max(this.selectStartIndex, this.selectEndIndex))) ||
```
PatrickTasse, line 538: Will this prevent searching backwards if there is no selection (both are -1, right?)

packages/react-components/src/components/table-output-component.tsx
```diff
@@ -526,9 +549,11 @@ export class TableOutputComponent extends AbstractOutputComponent<TableOutputPro
             if (isFound) {
                 return;
             }
-
-            const currRowIndex = this.gridApi?.getLastDisplayedRow() + 1;
-            const lineIndex = await this.findNextMatchIndex(currRowIndex);
+            if (direction === Direction.NEXT) {
+                currRowIndex = this.gridApi?.getLastDisplayedRow() + 1;
```
PatrickTasse, line 553: I'm not sure if necessary, but maybe it needs to be the first index following the last row node index that was considered by the 'current block' fix above?

And for PREVIOUS it should be the first index before the first considered row node index?

PatrickTasse (changes requested):
Some more suggestions perhaps for future enhancement:

- It should be possible to use Enter/Shift+Enter when the table itself has focus
- There should be a visual feedback while we are searching the next/previous index, not only when we are waiting to fill the table at the found index
- It should not lose current table selection if there is no found index

packages/react-components/src/components/table-output-component.tsx
bhufmann: I just chose 200 because that's the current blocksize. I don't think there is a "good" value if the cacheBlockSize is undefined.

packages/react-components/src/components/table-output-component.tsx
```diff
@@ -508,13 +516,28 @@ export class TableOutputComponent extends AbstractOutputComponent<TableOutputPro
         return lines[0].index;
     }
 
-    private async findNextMatchedEvent() {
+    private async findMatchedEvent(direction: Direction) {
         let isFound = false;
         if (this.gridApi) {
             this.gridApi.deselectAll();
+            let rowNodes: RowNode[] = [];
+            let currRowIndex = this.selectStartIndex === -1 ? this.gridApi?.getLastDisplayedRow() + 1 : this.selectStartIndex;
```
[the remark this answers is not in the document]
bhufmann: I'll check if this works

packages/react-components/src/components/table-output-component.tsx
bhufmann: mhh, maybe it should compare with currRowIndex instead? I'll check

packages/react-components/src/components/table-output-component.tsx
bhufmann: I'll check

packages/react-components/src/components/table-output-component.tsx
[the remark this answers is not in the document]
PatrickTasse: If there is no selection I would expect the search would find the first match in the visible rows if there is any?

packages/react-components/src/components/table-output-component.tsx
PatrickTasse: To clarify, what I'm worried is if we accurately compute the 'block number' for comparison in that case.

packages/react-components/src/components/table-output-component.tsx
```diff
@@ -508,13 +516,59 @@ export class TableOutputComponent extends AbstractOutputComponent<TableOutputPro
...
+
+            if (direction === Direction.PREVIOUS) {
+                rowNodes = rowNodes.reverse();
+            }
+
+            // search inside current block where current block is where the selection is either
+            rowNodes.forEach(rowNode => {
+                let condition = this.selectStartIndex === -1 ? currentBlock * blockSize : Math.max(this.selectStartIndex, this.selectEndIndex);
+                if (direction === Direction.PREVIOUS) {
+                    condition = this.selectStartIndex === -1 ? (currentBlock + 1) * blockSize - 1 : Math.min(this.selectStartIndex, this.selectEndIndex);
```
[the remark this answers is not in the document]
PatrickTasse: I propose that forward search start at index 0 when there is no selection, and backward search be a noop when there is no selection.

packages/react-components/src/components/table-output-component.tsx
```diff
@@ -508,13 +516,59 @@ export class TableOutputComponent extends AbstractOutputComponent<TableOutputPro
...
+                if (direction === Direction.PREVIOUS) {
+                    Math.min(this.selectStartIndex, this.selectEndIndex);
+                }
+            }
+
             this.gridApi.forEachNode(rowNode => {
-                if (rowNode.rowIndex > Math.max(this.selectStartIndex, this.selectEndIndex) && rowNode.data
-                    && rowNode.data['isMatched'] && !isFound) {
+                // make sure to search only in current block and not in full cache
+                if (currentBlock === Math.floor(rowNode.rowIndex / blockSize)) {
```
PatrickTasse, line 551: This prevents considering contiguous blocks. I propose to use a scheme that only considers contiguous row index regardless of their block.

packages/react-components/src/components/table-output-component.tsx
```diff
@@ -524,14 +578,19 @@ export class TableOutputComponent extends AbstractOutputComponent<TableOutputPro
...
             if (isFound) {
+                // match found in current block
                 return;
             }
 
-            const currRowIndex = this.gridApi?.getLastDisplayedRow() + 1;
-            const lineIndex = await this.findNextMatchIndex(currRowIndex);
-            if (lineIndex) {
+            // find match outside the current block
+            const lineIndex = await this.findMatchIndex(currRowIndex, direction);
```
PatrickTasse, line 586: This sends the index of a row that should not be searched, then the server skips that index and starts at the next/previous index.
I propose that the client should send the first index that should be searched (so after/before the selection), and the server be changed to start at that index. See https://git.eclipse.org/r/c/tracecompass/org.eclipse.tracecompass/+/183874.

PatrickTasse (changes requested):

PatrickTasse (approved):

Issue #282, opened 2021-03-01
Make chart and filter field backgrounds a darker colour in dark theme


ebugden:
In dark theme, the contrast between the chart background and the tab background should be reduced (same for the filter background). This makes the data stand out more and draw the user's eye towards the data.
* New dark theme filter and background colours should have a similar difference in brightness to the existing light theme background colours.
* Tip: [Colorizer.org](http://colorizer.org/) is useful for easily adjusting colours along different scales 

| Current Light Theme  | Proposed Dark Theme |
|-------------------|--------------------------|
| **Difference in brightness (HSB):** ~2 ![image](https://user-images.githubusercontent.com/28311615/121205273-84213700-c845-11eb-9209-2a0757369e58.png) | **Difference in brightness (HSB):** ~2 ![image](https://user-images.githubusercontent.com/28311615/109556427-cf88b400-7aa4-11eb-8c48-e7857c86a2c0.png) |
| Background: #ffffff ![image](https://user-images.githubusercontent.com/28311615/121209393-d0ba4180-c848-11eb-9770-53711848718f.png) Chart background: #f4f7fb ![image](https://user-images.githubusercontent.com/28311615/121209511-e3347b00-c848-11eb-91fc-40c79c75fd6e.png) | **Chart background:** #232323 ![image](https://user-images.githubusercontent.com/28311615/121209699-0f4ffc00-c849-11eb-836d-e6f2b30e50da.png) **Tab background:** #1e1e1e ![image](https://user-images.githubusercontent.com/28311615/121208524-1fb3a700-c848-11eb-8b9d-ff6910666b48.png) |

![image](https://user-images.githubusercontent.com/28311615/109555660-d105ac80-7aa3-11eb-86f2-8c38d45cc9c7.png)

Comparison of darker and lighter backgrounds:
| Current backgrounds (lighter) | Proposed backgrounds (darker) |
|-------------------|--------------------------|
| **Difference in brightness (HSB):** ~13 ![image](https://user-images.githubusercontent.com/28311615/109554802-b7179a00-7aa2-11eb-83ac-4a0c68e62500.png)| **Difference in brightness (HSB):** ~2 ![image](https://user-images.githubusercontent.com/28311615/109556427-cf88b400-7aa4-11eb-8c48-e7857c86a2c0.png) |
[...]

PR #442, opened 2021-08-05
Reduction contrast of grid lines

### LINKED ISSUE #441: Reduce contrast of grid lines
Grid lines are much more common yet much less important than the data being visualized, and therefore grid lines should be about "one noticeable difference" in contrast from the background of the chart. The current chart background is #232323. I'm proposing we use these colors for horizontal and vertical grid lines:

- #35373B for horizontal grid lines
- #2B2E31 for vertical grid lines

This will yield a more subdued grid similar to this screenshot (click to view something closer to actual pixels):
image

Let's compare and contrast (pun intended!) that grid with this grid in the current implementation:
image

In the above example, the grid lines are much, _much_ more prominent than the data itself. 

Because the proposed grid lines are more subdued, in a future iteration we may want to supplement the grid by highlighting grid lines when the user mouses near a line or an intersection of lines. In the current iteration the user must click.

IbrahimFradj:
change the color of lineColor with #35373B for horizontal grid lines and #2B2E31 for vertical grid lines

**Here is the result in the screenshot :**

image

Fixes #441

### DIFF
--- a/packages/react-components/src/components/timegraph-output-component.tsx
+++ b/packages/react-components/src/components/timegraph-output-component.tsx
@@ -81,7 +81,7 @@ export class TimegraphOutputComponent extends AbstractTreeOutputComponent<Timegr
             rowStyleProvider: (row: TimelineChart.TimeGraphRowModel) => ({
                 backgroundColor: 0x979797,// 0xaaaaff,
                 backgroundOpacity: row.selected ? 0.1 : 0,
-                lineColor: 0xdddddd, // hasStates ? 0xdddddd : 0xaa4444, // row.data && row.data.hasStates
+                lineColor: this.props.backgroundTheme === 'light' ? 0xD3D3D3 : 0x3F4146 , // hasStates ? 0xdddddd : 0xaa4444, // row.data && row.data.hasStates
                 lineThickness: 1, // hasStates ? 1 : 3 // row.data && row.data.hasStates
             })
         };
@@ -328,7 +328,7 @@ export class TimegraphOutputComponent extends AbstractTreeOutputComponent<Timegr
     }
 
     private getChartContainer() {
-        const grid = new TimeGraphChartGrid('timeGraphGrid', this.props.style.rowHeight, this.props.style.lineColor);
+        const grid = new TimeGraphChartGrid('timeGraphGrid', this.props.style.rowHeight, this.props.backgroundTheme === 'light' ? 0xdddddd : 0x34383C);
         const selectionRange = new TimeGraphChartSelectionRange('chart-selection-range', { color: this.props.style.cursorColor });
         return <ReactTimeGraphContainer
             options={
@@ -337,7 +337,7 @@ export class TimegraphOutputComponent extends AbstractTreeOutputComponent<Timegr
                     height: parseInt(this.props.style.height.toString()),
                     width: this.props.style.chartWidth, // this.props.style.mainWidth,
                     backgroundColor: this.props.style.chartBackgroundColor,
-                    lineColor: this.props.style.lineColor,
+                    lineColor: this.props.backgroundTheme === 'light' ? 0xdddddd : 0x34383C,
                     classNames: 'horizontal-canvas'
                 }
             }
--- a/packages/react-components/src/components/trace-context-component.tsx
+++ b/packages/react-components/src/components/trace-context-component.tsx
@@ -101,7 +101,7 @@ export class TraceContextComponent extends React.Component<TraceContextProps, Tr
                 height: this.DEFAULT_COMPONENT_HEIGHT,
                 rowHeight: this.DEFAULT_COMPONENT_ROWHEIGHT,
                 naviBackgroundColor: this.props.backgroundTheme === 'light' ? 0xf4f7fb : 0x3f3f3f,
-                chartBackgroundColor: this.props.backgroundTheme === 'light' ? 0xf4f7fb : 0x3f3f3f,
+                chartBackgroundColor: this.props.backgroundTheme === 'light' ? 0xf4f7fb : 0x232323,
                 cursorColor: 0x259fd8,
                 lineColor: this.props.backgroundTheme === 'light' ? 0x757575 : 0xbbbbbb
             },
@@ -135,7 +135,7 @@ export class TraceContextComponent extends React.Component<TraceContextProps, Tr
                 height: this.DEFAULT_COMPONENT_HEIGHT,
                 rowHeight: this.DEFAULT_COMPONENT_ROWHEIGHT,
                 naviBackgroundColor: theme === 'light' ? 0xf4f7fb : 0x3f3f3f,
-                chartBackgroundColor: theme === 'light' ? 0xf4f7fb : 0x3f3f3f,
+                chartBackgroundColor: theme === 'light' ? 0xf4f7fb : 0x232323,
                 cursorColor: 0x259fd8,
                 lineColor: theme === 'light' ? 0x757575 : 0xbbbbbb
             },


packages/react-components/src/components/timegraph-output-component.tsx
```diff
@@ -81,7 +81,7 @@ export class TimegraphOutputComponent extends AbstractTreeOutputComponent<Timegr
             rowStyleProvider: (row: TimelineChart.TimeGraphRowModel) => ({
                 backgroundColor: 0x979797,// 0xaaaaff,
                 backgroundOpacity: row.selected ? 0.1 : 0,
-                lineColor: 0xdddddd, // hasStates ? 0xdddddd : 0xaa4444, // row.data && row.data.hasStates
+                lineColor: 0x35373B , // hasStates ? 0xdddddd : 0xaa4444, // row.data && row.data.hasStates
```
PatrickTasse, line 84: These are only an improvement in dark theme, in light theme it increases the contrast, so the grid line colors should probably be theme dependent...

PatrickTasse (changes requested):

ssmagula:
Reducing contrast of grid lines requires the darker color for the chart background (#232323). See "Make chart and filter field backgrounds a darker" in https://github.com/theia-ide/theia-trace-extension/issues/282

packages/react-components/src/components/timegraph-output-component.tsx
bhufmann: I agree with @PatrickTasse that the color should depend on dark or light theme. Personally, I find the colors of the gridlines to dark and not visible, especially the vertical gridlines.

bhufmann (changes requested):
I found a bug, when resizing the chart, then the gridline colors are not kept and the original colors are back.

Besides, I noticed that the time axis still has the original color for the vertical ticks. They should have the same color, right?

IbrahimFradj:
> 
> 
> I found a bug, when resizing the chart, then the gridline colors are not kept and the original colors are back.
> 
> Besides, I noticed that the time axis still has the original color for the vertical ticks. They should have the same color, right?

for the time axis, if I change the color, the time will not be visible
image

bhufmann:
> > I found a bug, when resizing the chart, then the gridline colors are not kept and the original colors are back.
> > Besides, I noticed that the time axis still has the original color for the vertical ticks. They should have the same color, right?
> 
> for the time axis, if I change the color, the time will not be visible
> image

Thanks for trying. Can you open a separate issue tracker to be able to configure the tick and text color separately for the time axis?

bhufmann (changes requested):
There seems to be an initialization error for the gridlines. The new color is only applied to a certain window range. Outside the window it uses the old color. See circled areas below:

image

The color in the light theme still needs to be fixed.

bhufmann:
@ssmagula the selected color for the gridlines feels to dark and is barely visible in my opinion. Any other suggestion?

ssmagula:
I feel we need to take a holistic approach to color. I'm trying to tackle this challenge with a high fidelity prototype that I hope will help us preview a variety of options in a holistic way.

Meanwhile, below is a screenshot of a chart with grid lines that offer slightly more contrast against the chart background of #232323:
image

If that offers just enough contrast, then these new grid line colors are:
Chart Background: #232323
New Horiz line #3F4146
New Vert line #34383C

mirsky-work:
Just throwing an idea - should the vertical gridlines be more visible than the horizontal ones, since those are probably of more interest to the users? I believe that visually placing events on the same row should be easy enough without very noticeable gridlines.

I also noticed a related bug when testing this PR, but it's also present on the main branch, so opened #455

bhufmann:
@ssmagula the selected color for the gridlines feels to dark and is barely visible in my opinion. Any other suggestion?

> I feel we need to take a holistic approach to color. I'm trying to tackle this challenge with a high fidelity prototype that I hope will help us preview a variety of options in a holistic way.
> 
> Meanwhile, below is a screenshot of a chart with grid lines that offer slightly more contrast against the chart background of #232323:
> image
> 
> If that offers just enough contrast, then these new grid line colors are:
> Chart Background: #232323
> New Horiz line #3F4146
> New Vert line #34383C

New colors are more visible and I'm ok with them.

packages/react-components/src/components/timegraph-output-component.tsx
IbrahimFradj: I changed the colors with what Stefan said.

bhufmann:
> There seems to be an initialization error for the gridlines. The new color is only applied to a certain window range. Outside the window it uses the old color. See circled areas below:
> 
> image
> 
> The color in the light theme still needs to be fixed.

I can't reproduce this anymore with the latest updates. It looks your latest change fixed it.

packages/react-components/src/components/timegraph-output-component.tsx
```diff
@@ -81,7 +81,7 @@ export class TimegraphOutputComponent extends AbstractTreeOutputComponent<Timegr
             rowStyleProvider: (row: TimelineChart.TimeGraphRowModel) => ({
                 backgroundColor: 0x979797,// 0xaaaaff,
                 backgroundOpacity: row.selected ? 0.1 : 0,
-                lineColor: 0xdddddd, // hasStates ? 0xdddddd : 0xaa4444, // row.data && row.data.hasStates
+                lineColor: this.props.backgroundTheme === 'light' ? 0xD3D3D3 : 0x3F4146 , // hasStates ? 0xdddddd : 0xaa4444, // row.data && row.data.hasStates
```
bhufmann, line 84: This is executed once when the row is created. When changing the theme this is not called again until a new row is created. This can happen when the few is closed and reopend or when zooming to different time range where the rows are rebuild with new data from the back-end.  I would suggest to create a new issue tracker to fix this and we work on merging this PR.

packages/react-components/src/components/timegraph-output-component.tsx
```diff
@@ -328,7 +328,7 @@ export class TimegraphOutputComponent extends AbstractTreeOutputComponent<Timegr
     }
 
     private getChartContainer() {
-        const grid = new TimeGraphChartGrid('timeGraphGrid', this.props.style.rowHeight, this.props.style.lineColor);
+        const grid = new TimeGraphChartGrid('timeGraphGrid', this.props.style.rowHeight, this.props.backgroundTheme === 'light' ? 0xdddddd : 0x34383C);
```
bhufmann, line 331: I think we should keep the original code here and update the TimeAxisComponent instead to have the previous color. Please open an issue tracker for providing 2 different colors for line and text for the TimeAxisComponent. So, that we can configure the color of the TimeAxisComponent as well.

packages/react-components/src/components/timegraph-output-component.tsx
```diff
@@ -337,7 +337,7 @@ export class TimegraphOutputComponent extends AbstractTreeOutputComponent<Timegr
                     height: parseInt(this.props.style.height.toString()),
                     width: this.props.style.chartWidth, // this.props.style.mainWidth,
                     backgroundColor: this.props.style.chartBackgroundColor,
-                    lineColor: this.props.style.lineColor,
+                    lineColor: this.props.backgroundTheme === 'light' ? 0xdddddd : 0x34383C,
```
bhufmann, line 340: I think we should keep the original code here and update the TimeAxisComponent instead to have the previous color.

bhufmann (changes requested):

MatthewKhouzam (commented):
removing approve to make Bernd's comments more visible.

bhufmann (approved):
I think it's good to go. I'll open 2 issue trackers to follow-up:
- Fix the theme switch
- Provide different properties in OutputComponentStyle for horizontal line, vertical line and text color

Issue #433, opened 2021-07-30
Time Graph: Align row labels with rows (recent Chrome only)


ebugden:
On recent Chrome browsers, there is a large vertical offset between where the chart starts and where line labels start so labels do not line up with the associated rows. (Row information is still available using the tooltip, but this isn't convenient)

* Observed in both Resources Status and Thread Status views, so bug assumed present in all timegraph views.
* Browser: Chrome, Version 92.0.4515.107

> **bhufmann:** This problem only started with recent Chrome versions. Older Chrome versions or Firefox (v90.0) don't have that issue. I don't know which version it started, but with Chrome version 92.0.4515.107 it fails.

https://user-images.githubusercontent.com/28311615/127364700-05f6899b-fbe7-43cc-a4df-f8d43a9cd7ff.mp4

ebugden:
Let me know if this is too specific to the timeline chart and should be moved to the [timeline-chart repo issues](https://github.com/theia-ide/timeline-chart/issues).

bhufmann:
> Let me know if this is too specific to the timeline chart and should be moved to the [timeline-chart repo issues](https://github.com/theia-ide/timeline-chart/issues).

I think that the problem needs to be fixed here because the tree implementation and row synchronization is in the theia-trace-extension. So, let's keep the tracker here.

bhufmann:
@ebugden This problem only started with recent Chrome versions. Older Chrome versions or Firefox (v90.0) don't have that issue. I don't know which version it started, but with Chrome version 92.0.4515.107 it fails.

ebugden:
Ok, thanks for the context! I'll add it in the issue description.

PR #443, opened 2021-08-10
Fix misalignment between labels and rows

Rodrigoplp-work:
Fix misalignment between labels and rows in the Resources Status and Thread Status views.

In Chrome, the first line of the labels in the side bar of these views would not be aligned with the first row of the data. It would be located much lower in the view.

See attached images of the Resources Status view in Chrome and Firefox (added for comparison), before and after the fix.

Chrome before fix:
Chrome-before

Chrome after fix:
Chrome-after

Firefox before fix:
Firefox-before

Firefox after fix:
Firefox-after

### DIFF
--- a/packages/react-components/style/output-components-style.css
+++ b/packages/react-components/style/output-components-style.css
@@ -101,7 +101,6 @@ canvas {
     line-height: 18px;
     position: relative;
     white-space: nowrap;
-    top: 50%;
     padding: 0 0;
 }
 


IbrahimFradj (changes requested):
it looks good, it just needs some modifications :

- [ ] Can you please associate the issue with this PR example : Fixes #433 and put it in your description and in your commit.
- [ ] Put your Signed-off in your commit and in your description https://github.com/theia-ide/theia-trace-extension/pull/443/checks?check_run_id=3292720065
- [ ] Please update the commit message. Use a short description in the first line because you can't read it correctly, it's like having a line break in the middle of the word.

Thank you for this contribution!

PatrickTasse (approved):

IbrahimFradj (approved):

PR #447, opened 2021-08-12
Show view message when there is no data to provide

### LINKED ISSUE #406: Inform user when there is no data to perform analysis
At the moment, certain views (listed below) spin infinitely when the data required for the analysis isn't in the traces. 
* This needs to happen for all chart types. Right now there is a message that displays for XY charts only #447

## To fix this
- [ ] Fix that in the fetchTree() method, the state of the component (xy-output-component.tsx) is never updated when the response indicates COMPLETED and the model is undefined.
- [ ] Update render() method to handle empty xyTree and show NO_DATA instead.
   * Ideally the message displayed would tell the user why the view is empty (ex. "Nothing to see: The data (events) required for this view are not in the trace.")

(Updated steps based on @bhufmann's [comment](https://github.com/theia-ide/theia-trace-extension/issues/406#issuecomment-874809317) below)

More context in #296 ([comment](https://github.com/theia-ide/theia-trace-extension/issues/296#issuecomment-844495341))

### Related example patches:
* [Return an empty list](https://git.eclipse.org/r/c/tracecompass/org.eclipse.tracecompass/+/182636/1)
* [Display "No Data"](https://git.eclipse.org/r/c/tracecompass/org.eclipse.tracecompass/+/182633) (Trace Compass patch)

## Views with this problem
* Ust memory latency vs time
* Memory usage
* Event Matching Latency - Latency vs Time

Files and function to possibly adjust:
image

jeffGodonou:
Some views weren't able to display any data after the completion of the trace's analysis.
The problem was due to the fact that some provided models are empty server side because
they were no relevents trace events to display.

We choose to display a message indicating that there are no data to display in those cases
and update the state of the analysis client side. It will allow a better clarity of the user
experience.

Fixes #406
Signed-off-by: Jeff Godonou <jeffalecgodonou@gmail.com>

### DIFF
--- a/packages/react-components/src/components/xy-output-component.tsx
+++ b/packages/react-components/src/components/xy-output-component.tsx
@@ -93,6 +93,10 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
                     xyTree: treeResponse.model.entries,
                     columns
                 });
+            } else {
+                this.setState({
+                    outputStatus: treeResponse.status
+                });
             }
             return treeResponse.status;
         }
@@ -159,6 +163,17 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
             events: [ 'mousedown' ],
         };
         // width={this.props.style.chartWidth}
+        if (this.state.outputStatus === ResponseStatus.COMPLETED && this.state.xyTree.length === 0 ) {
+            return <React.Fragment>
+                {
+                    <p className='no-data' style={{fontSize: 20, marginRight: '5px', marginLeft: '5px', justifyContent:'center', alignItems:'center'}}>
+                        Trace analysis completed.
+                        No results: Trace missing required events.
+                    </p>
+                }
+            </React.Fragment>;
+        }
+
         return <React.Fragment>
             {this.state.outputStatus === ResponseStatus.COMPLETED ?
                 <div id='xy-main' onMouseDown={event => this.beginSelection(event)} style={{ height: this.props.style.height }} >
--- a/packages/react-components/style/output-components-style.css
+++ b/packages/react-components/style/output-components-style.css
@@ -81,6 +81,16 @@ canvas {
     align-self: center;
 }
 
+.no-data {
+    font-size: 24px;
+    text-align: center;
+    position: relative;
+    transform: translate(0, -50%);
+    top: 50%;
+    width: 90%;
+    align-self: center;
+}
+
 .table-tree>tbody>tr:nth-child(1)>th {
     background-color: var(--theia-editor-background);
     position: sticky;


IbrahimFradj (changes requested):
I noticed a few bugs: 

- when I click on a view that is not working (e.g. Event Matching Latency - Latency vs Time), the correct message is displayed but when I open another view that is working, I first see the message that was in the old view.

**To reproduce :**
 
jeff bug 1
_____________________________________________________________________________________________
- When I open the view that is not working (Event Matching Latency - Latency vs Time), close it and reopen it, the message is no longer visible.

**To reproduce :**

jeff bug 2
_____________________________________________________________________________________________
- When I refresh the page with the non-working view already open and open it again, the message is not visible. 

**To reproduce :**

jeff bug 3

Thank you Jeff for your contribution.

jeffGodonou:
> I noticed a few bugs:
> 
>     * when I click on a view that is not working (e.g. Event Matching Latency - Latency vs Time), the correct message is displayed but when I open another view that is working, I first see the message that was in the old view.
> 
> 
> **To reproduce :**
> 
> jeff bug 1
> 
>     * When I open the view that is not working (Event Matching Latency - Latency vs Time), close it and reopen it, the message is no longer visible.
> 
> 
> **To reproduce :**
> 
> jeff bug 2
> 
>     * When I refresh the page with the non-working view already open and open it again, the message is not visible.
> 
> 
> **To reproduce :**
> 
> jeff bug 3
> 
> Thank you Jeff for your contribution.

jeffGodonou:
Thanks to report that. I'll watch that.

ebugden:
I was able to reproduce certain bugs observed by @IbrahimFradj. I think it's possible that these weren't necessarily introduced by this feature.

## Setup
* Browser: Chrome (version 92.0.4515.131)
* Trace: "apt"
* The exceptional behaviour happens **the first time the analysis results are requested** from the server. To re-observe, close the trace in the server (right-click in "Opened Traces" --> "Delete Trace").

## Bugs observed

### 1. On first open, "No data" message shows before "Analysis running"
For views with results to display (ex. CPU Usage, Disk I/O for trace "apt"): 
* The very first time they are opened, the  "Trace analysis complete. No related events found" message shows for a brief moment before the "Analysis running" spinner appears. 
* Apart from the very first open, this does not happen.  

https://user-images.githubusercontent.com/28311615/129622211-860a634d-c4fe-4033-8312-2ed73b013513.mp4

### 2. Some views display "No data" message on first open, but then show results when opened again later

* For the trace "apt", the views "Event Matching Latency", "Futex Contention Analysis", and "IRQ Analysis" all do this. For the trace "201-lttng-userspace-tracing", the "Ust Memory" view appears to do this. 
* I think this is what @IbrahimFradj was observing in the 2nd and 3rd case above. 

This suggests that either:
* The wrong indicator from the server is being used to show the "No data" message and the analysis isn't actually complete.
* The results aren't actually available when the server says the analysis is complete.

https://user-images.githubusercontent.com/28311615/129623660-735a30f1-d254-4820-91f3-175f3f96f6fe.mp4

### 3. Some views don't display any data when their box is checked

For the trace "apt", the view "Event Matching Latency" appears to display nothing in the chart when the box is checked. Having an empty graph even though the chart and checkbox areas appears, suggests that the wrong thing is being used to determine whether or not an analysis found no relevant events.

https://user-images.githubusercontent.com/28311615/129625486-f314d370-d6a2-4453-a1f6-62b8c3af1300.mp4

----

## Can we filter all the finicky XYs until they actually work?
Coming back to the root motivation for these changes #296, I seriously want to filter everything that is displayed as an XY Chart (except maybe CPU Usage and Disk I/O) and only make views available once they are complete (ex. Have y-axis labels), are tested, and can actually be used for analysis. I feel like we're trying to duct-tape views that weren't ready to be available as features in the first place. Since @jeffGodonou and @thefinaljob need XY Charts to be able to test their current features, I think filtering broken XY Charts should be done in September.

----

> @IbrahimFradj: 
> * When I open the view that is not working (Event Matching Latency - Latency vs Time), close it and reopen it, the message is no longer visible.
> * When I refresh the page with the non-working view already open and open it again, the message is not visible.

I think these two issues are the same. Both involve closing the view (either by clicking the x or by refreshing the tab).

ebugden:
@jeffGodonou Was it verified that the events required for the analysis were missing from the trace? Or did we assume that that was why the sever wasn't returning anything?

bhufmann (dismissed):
LGTM. Thanks for the contribution.

IbrahimFradj (changes requested):
With the **ust Memory** view everything works fine but with the **Event Matching Latency - Latency vs time** view the view stays stuck in analysis running and when I close and reopen the view it displays an empty view

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -93,6 +93,10 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
                     xyTree: treeResponse.model.entries,
                     columns
                 });
+            } else {
+                this.setState({
+                    outputStatus: treeResponse.status
+                });
             }
             return treeResponse.status;
```
bhufmann, line 101: if status is FAILED, then the XY chart is not updated and the "Analysis Running" message is shown. It can be reproduced with the apt trace in the tracevizlab when opening the first time the event matching xy chart. The failed case needs to be updated as well.

bhufmann (changes requested):

packages/react-components/src/components/xy-output-component.tsx
ebugden: Thank you for figuring out this bug! Should these additional changes be a separate PR or just a separate commit in this PR (with an updated description)?

bhufmann: @ebugden  I'm ok to track it in a separate issue. This PR alone is already an good improvement and it's good to get it merged. Could you please open a separate issue tracker?

ebugden: For those trying to reproduce the failed status case with the Event Matching Latency analysis: **Event Matching Latency only returns status FAILED the first time the analysis is opened on server (0:26-0:30)**, otherwise it returns the status COMPLETED (0:37-0:40). 

https://user-images.githubusercontent.com/28311615/130796784-624f5c6d-48d8-4cb6-9a9b-4703d439c61f.mp4

(Forgive my debugging via printing, I don't know how to debug using gitpod.)

ebugden:
After some thought, it may be more accurate to show a different message like: 

**"Trace analysis complete. No results: Trace missing required events."**

The message "Trace analysis complete. No related events found." says that no events relevant to the analysis were found. My guess is that this isn't necessarily true. If an analysis requires event types A and B, but only events B are present in the trace, the analysis results would be empty even if some of the relevant events are in the trace, right? @arfio Could you confirm/deny whether this how analyses behave?

arfio:
It can behave like this, yes. It all depends on how the analysis is implemented. Most analysis will look at particular event types, to calculate latencies for example, or just to create a custom flame chart. In some cases, missing some event types is expected, in others it might break the analysis, or just not output anything. But I agree that  "Trace analysis complete. No results: Trace missing required events." is a more accurate statement than "Trace analysis complete. No related events found.". The latter statement can be misleading in some cases.

ebugden:
> **arfio:** I agree that "Trace analysis complete. No results: Trace missing required events." is a more accurate statement than "Trace analysis complete. No related events found.". The latter statement can be misleading in some cases.

Ok! @jeffGodonou could you please:

- [ ] Change the view message to "Trace analysis complete. No results: Trace missing required events."

----
> **arfio:** It can behave like this, yes. It all depends on how the analysis is implemented. Most analysis will look at particular event types, to calculate latencies for example, or just to create a custom flame chart. In some cases, missing some event types is expected, in others it might break the analysis, or just not output anything. 

* Ok, so sometimes certain events are soft requirements. 
* It's interesting to note that sometimes analyses will fail if there are events missing. This could be relevant context for #460 since this means an analysis could fail even if it's not "broken".

ebugden:
I can't do a proper review since Gitpod isn't working right now, but looks good to me! Thanks for the update 🌿

IbrahimFradj (changes requested):
I noticed that the message does not display with **Event Matching Latency - Latency vs. Time** and with **Memory Latency vs. Time**.
I also noticed that for the **Memory Usage** view, sometimes it displays with some traces and sometimes it doesn't display and gets stuck in Analysis running. (ex: with the **apt** trace it works but with **ls-ust** it doesn't)

jeffGodonou:
> I noticed that the message does not display with **Event Matching Latency - Latency vs. Time** and with **Memory Latency vs. Time**.
> I also noticed that for the **Memory Usage** view, sometimes it displays with some traces and sometimes it doesn't display and gets stuck in Analysis running. (ex: with the **apt** trace it works but with **ls-ust** it doesn't)

It appears as a set back. Thank you for reporting this. I'll verify what happen with my last commit.

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -159,6 +163,17 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
             events: [ 'mousedown' ],
         };
         // width={this.props.style.chartWidth}
+        if (this.state.outputStatus === ResponseStatus.COMPLETED && this.state.xyTree.length === 0 ) {
```
ebugden, line 166: > @IbrahimFradj: I noticed that the message does not display with **Event Matching Latency - Latency vs. Time** and with **Memory Latency vs. Time**.
> I also noticed that for the **Memory Usage** view, sometimes it displays with some traces and sometimes it doesn't display and gets stuck in Analysis running. (ex: with the **apt** trace it works but with **ls-ust** it doesn't)

My understanding is this message only displays when then analysis returns status "COMPLETE", but with an empty model. 

Could you check whether you're observing cases where the analysis returns "FAILED"? The case where the analysis returns "FAILED" is still not managed #460.

jeffGodonou: Yes I can look at that

ebugden: To clarify, @bhufmann said earlier that this PR doesn't need to manage the FAILED case. If the problems @IbrahimFradj noticed are with the FAILED case then we can still merge this PR.

image
Source: [Comment thread](https://github.com/theia-ide/theia-trace-extension/pull/447#discussion_r695193161)

bhufmann (approved):
LGTM. Thanks. @jeffGodonou please rebase the patch to latest maser so that we can merge.

IbrahimFradj (approved):
LGTM. @jeffGodonou an you put it in one commit please thank you

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -159,6 +163,17 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
+            return <React.Fragment>
+                {
+                    <p className='no-data' style={{fontSize: 20, marginRight: '5px', marginLeft: '5px', justifyContent:'center', alignItems:'center'}}>
```
ebugden, line 169: Why are there style properties here that contradict or seem to repeat the styles in the no-data CSS class (see below: output-components-style.css line 84)? Why not just have all the style properties in the CSS class?

Examples:
* fontSize: 20 vs. font-size: 24
* justifyContent: center, alignItems: center vs. text-align: center

MatthewKhouzam (approved):
lgtm

Issue #384, opened 2021-06-07
Allow renaming trace viewer tabs


ebugden:
Allowing users to rename tabs lets them give concise and meaningful labels. This makes it easier to distinguish between tabs because, unlike the trace file names (which tend to be long), user written labels are more likely to fit inside narrow tabs without being cut off. 

Examples of useful tab names:
* Crash A (Server 1), Crash A (Server 2)
* Before update, After update
* Baseline, Change A, Change B

Changes made to the tab name shouldn't change the trace file name (or the trace group name).

----
CC: @ssmagula

thefinaljob:
Revised approach to solving the issue:
- Decouple code related to menu items in trace-explorer-opened-traces-widget.tsx by making it into its own component
   - pass in functions from the traces widget to the menu item component to avoid code duplication
   - move functions from trace widget to menu item component whenever possible
- Tie state of the tab to the state of the menu-item so that changes are reflected when the menu item is changed
   - Approach to this is TBD, could use a state store like Redux or callback functions

CC: Arnaud Fiorini

ebugden:
@thefinaljob @ssmagula You plan to have the edit trace button appear in the opened traces section (A) rather than in the trace tab label (B)?

Couldn't we edit directly in the tab and either:
* **When the pencil fades in, it obscures a portion of the title** that was previously visible (instead of always leaving an empty space after the title), or
* **On hover of the tab label text, the cursor changes to the "editable text" cursor**: Ꮖ

My reflex would be to put it in the trace tab first:
* I think that's where people would look first. I don't think people would necessarily make the link that editing the name in the list of traces would affect the tab name.
* One main goal of renaming trace groups is to have a label that is readable within the space of a tab. Editing the tab label directly allows users to use the tab size as a reference for whether it will be readable or not.
* I don't think space would be a big issue. My guess is that tracers would have 1-2 (4 max) trace groups open. Although if you're also developing within the context of tracing you could have source files open as well...

![rename-tab-placement](https://user-images.githubusercontent.com/28311615/126699564-5c8bef82-b17e-4a04-9007-ff71314a066d.png)

ebugden:
**RE: Space for editing button** - This shouldn't be an issue since the tab size doesn't shrink in Theia when there are too many to display in the tab bar. In Theia, tab size is a function of label length. When there are too many tabs to display in the tab bar, a scroll wheel appears.

Example with a ton of tabs open:

[...]

Issue #450, opened 2021-08-16
Clicking on trace file to change tab focus opens duplicate tab


ebugden:
If the trace is already open in a tab, clicking the trace file in "Opened Traces" should just select the existing tab. 

In the video:

* **0:06** - Click to change tab highlight from "apt" trace to "201-lttng-userspace-tracing" trace that is already open in a tab --> Duplicate tab created.
* **0:14** - Click to change tab highlight from "apt" trace to "201-lttng-userspace-tracing" trace that is already open in a tab (and already has a duplicate) --> No additional duplicate created. This only happens once (only one duplicate is ever created).
* **0:18** - Duplicate "apt" tab created.

https://user-images.githubusercontent.com/28311615/129616513-063f849e-14f8-4090-b11f-788e7540d44b.mp4

**Browser:** Chrome (version 92.0.4515.131)

mirsky-work:
I believe it's a duplicate of #364.

Issue #364, opened 2021-05-11
The same trace must be opened only once 


IbrahimFradj:
I notice that the same trace can be opened twice. When we open the trace with TRACE VIEWER and double-click on the same open trace, another trace opens.  
Video from the gitpod demo (master):
**To reproduce (see video)**
![bug](https://user-images.githubusercontent.com/83769103/117889778-17387280-b282-11eb-9c26-fb25c23031b0.gif)

ebugden:
Closing since description is stale (bug now produced with single click) and #450 (duplicate) has a more detailed/updated description.

bhufmann:
I'm reopening this one because it's a different flavor of the bug that #450. #521 fixes #450 but doesn't fix it when following the steps described in this tracker.

bhufmann:
Closing it because #521 actually fixes this issue. #569 was created to track the remaining issue.

PR #451, opened 2021-08-17
Allow editing trace set labels via Opened Traces widget

thefinaljob:
Contributes towards fixing #384 (changing the trace set label also changes the tab label for that trace). 

Known Issues: 
- Default trace tab opened **on launch of trace server** does not have its tab renamed, issue is unrelated to code changes. @arfio is aware of this
- Editing tab could use a bit more padding

Default Behaviour:
- After consulting with @ebugden, both clicking outside the menu item and pressing enter will save the tab name and changes will be reflected in both the menu item and the tab name. @ssmagula
- Max tab length of 50 characters
- If 0 characters are inputed, tab name and menu item name are reverted to their state before they were being edited

Signed-off-by: Nikolai Peram <nikolai_peram@outlook.com>

### DIFF
--- a/packages/base/src/signals/signal-manager.ts
+++ b/packages/base/src/signals/signal-manager.ts
@@ -23,6 +23,7 @@ export declare interface SignalManager {
 }
 
 export const Signals = {
+    EXPERIMENT_CHANGED: 'tab changed',
     TRACE_OPENED: 'trace opened',
     TRACE_CLOSED: 'trace closed',
     EXPERIMENT_OPENED: 'experiment opened',
@@ -43,6 +44,9 @@ export const Signals = {
 };
 
 export class SignalManager extends EventEmitter implements SignalManager {
+    fireExperimentChangedSignal(tabName: string, experimentUUID: string): void {
+        this.emit(Signals.EXPERIMENT_CHANGED, { tabName, experimentUUID });
+    }
     fireTraceOpenedSignal(trace: Trace): void {
         this.emit(Signals.TRACE_OPENED, trace);
     }
--- a/packages/react-components/src/trace-explorer/menu-item-trace.tsx
+++ b/packages/react-components/src/trace-explorer/menu-item-trace.tsx
@@ -1,0 +1,148 @@
+import * as React from 'react';
+import { Trace } from 'tsp-typescript-client';
+import { signalManager, Signals } from '@trace-viewer/base/lib/signals/signal-manager';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import { faPencilAlt } from '@fortawesome/free-solid-svg-icons';
+
+interface MenuItemProps {
+    index: number;
+    experimentName: string;
+    experimentUUID: string;
+    traces: Trace[];
+    onExperimentNameChange: (newExperimentName: string, index: number) => void;
+    menuItemTraceContainerClassName: string;
+    handleClickEvent: (event: React.MouseEvent<HTMLDivElement>, experimentName: string) => void;
+    handleContextMenuEvent: (event: React.MouseEvent<HTMLDivElement>, experimentUUID: string) => void;
+}
+interface MenuItemState {
+    editingTab: boolean;
+    oldExperimentName: string;
+    experimentNameState: string;
+}
+
+export class MenuItemTrace extends React.Component<MenuItemProps, MenuItemState> {
+
+    private wrapper: React.RefObject<HTMLDivElement>;
+
+    constructor(menuItemProps: MenuItemProps) {
+        super(menuItemProps);
+        this.state = {
+            oldExperimentName: this.props.experimentName,
+            experimentNameState: this.props.experimentName,
+            editingTab: false
+        };
+        this.wrapper = React.createRef();
+        this.handleClickOutside = this.handleClickOutside.bind(this);
+    }
+
+    submitNewExperimentName(): void {
+        if (this.state.experimentNameState.length > 0) {
+            this.setState({
+                editingTab: false,
+                oldExperimentName: this.state.experimentNameState
+            });
+            this.props.onExperimentNameChange(this.state.experimentNameState, this.props.index);
+        } else {
+            this.setState({
+                editingTab: false,
+                experimentNameState: this.state.oldExperimentName
+            });
+            const tabName = 'Trace: ' + this.state.oldExperimentName;
+            signalManager().fireExperimentChangedSignal(tabName, this.props.experimentUUID);
+        }
+        document.removeEventListener('click', this.handleClickOutside);
+    }
+
+    handleClickOutside = (event: Event): void => {
+        const node = this.wrapper.current;
+        if ((!node || !node.contains(event.target as Node)) && this.state.editingTab === true) {
+            this.submitNewExperimentName();
+        }
+    };
+
+    protected handleEnterPress(event: React.KeyboardEvent<HTMLInputElement>): void{
+        if (event.key === 'Enter' && this.state.editingTab === true){
+            this.submitNewExperimentName();
+        }
+    }
+
+    protected changeText(event: React.ChangeEvent<HTMLInputElement>): void{
+        let newName = event.target.value.toString();
+        this.setState({
+            experimentNameState : newName
+        });
+        newName = 'Trace: ' + newName;
+        signalManager().fireExperimentChangedSignal(newName,  this.props.experimentUUID);
+    }
+    protected inputTab(): React.ReactNode {
+        if (!this.state.editingTab) {
+            return (
+            <div className='wrapper'>
+                {this.state.experimentNameState}
+                <div className='edit-trace-name' onClick={e => {this.renderEditTraceName(e);}}>
+                    <FontAwesomeIcon icon={faPencilAlt} />
+                </div>
+            </div>
+            );
+        }
+        return (<input name="tab-name" className="theia-input name-input-box"
+            defaultValue = {this.state.experimentNameState}
+            onChange = {e => (this.changeText(e))}
+            onClick = {e => e.stopPropagation()}
+            onKeyPress = {e => (this.handleEnterPress(e))}
+            maxLength = {50}
+        />);
+    }
+    protected renderEditTraceName(event: React.MouseEvent<HTMLDivElement>): void {
+        document.addEventListener('click', this.handleClickOutside);
+        this.setState(() => ({
+            editingTab: true
+        }));
+        event.stopPropagation();
+        event.preventDefault();
+    }
+    protected renderTracesForExperiment = (): React.ReactNode => this.doRenderTracesForExperiment();
+    protected doRenderTracesForExperiment(): React.ReactNode {
+        const tracePaths = this.props.traces;
+        return (
+            <div className='trace-element-path-container'>
+                {tracePaths.map(trace => (
+                    <div className='trace-element-path child-element' id={trace.UUID} key={trace.UUID}>
+                        {` > ${trace.name}`}
+                    </div>
+                ))}
+            </div>
+        );
+    }
+    protected subscribeToExplorerEvents(): void {
+        signalManager().on(Signals.OUTPUT_ADDED, this.changeText);
+    }
+
+    render(): JSX.Element {
+        return (
+            <div className={this.props.menuItemTraceContainerClassName}
+            id={`${this.props.menuItemTraceContainerClassName}-${this.props.index}`}
+            onClick={event => {
+                    this.props.handleClickEvent(event, this.props.experimentUUID);
+                }
+            }
+            onContextMenu={event => { this.props.handleContextMenuEvent(event, this.props.experimentUUID); }}
+            data-id={`${this.props.index}`}
+            ref={this.wrapper}>
+            <div className='trace-element-container'>
+                <div className='trace-element-info' >
+                    <h4 className='trace-element-name'>
+                        {this.inputTab()}
+                    </h4>
+                    { this.renderTracesForExperiment() }
+                </div>
+                {/* <div className='trace-element-options'>
+                    <button className='share-context-button' onClick={this.handleShareButtonClick.bind(this, props.index)}>
+                        <FontAwesomeIcon icon={faShareSquare} />
+                    </button>
+                </div> */}
+            </div>
+        </div>
+        );
+    }
+}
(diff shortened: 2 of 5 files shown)


mirsky-work:
Thanks for contributing this useful feature!
Works nicely, but I do have a couple of comments:
1. An open tab's name doesn't change until it's closed and then reopened. After that it does update while typing. See video below.
2. Initially trace names are prefixed with "Trace:". On the first renaming it's preserved, but in following renaming it's discarded. I think it would be nice to preserve it, do differentiate trace tabs from source code tabs, for instance, or at least have a consistent behavior.
3. I think it would be visually cleaner to align the pen icon to the right, so it'll be at the same "column" for all traces.
4. In addition to the pen icon, I think we should also add a "Rename Trace Tab" entry to the Opened Traces context menu.
5. The text input box, when renaming, could perhaps be thinner by a few pixels, so it won't shadow part of the first line of the experiment's contents.

https://user-images.githubusercontent.com/67920463/129965318-0f4816ba-5639-498f-a7a1-70444a305741.mp4

ssmagula:
Looking good! Below are some notes after screensharing with @thefinaljob :

This feature is valuable because experienced trace visualization users say things like: "It's important I know which is the 'before trace' and which is the 'after [I make a change] trace'..." and: "When I share screenshots with others I hate scrawling all the details on the screenshots like 'Before...' and 'After I did xyz' because it's so hard to write letters with a mouse."

Re: saving on enter or loss of focus: I've seen many users make edits in place then move on—never hitting enter—so saving when focus is lost is a good way to prevent errors. Nice nuance and thumbs up! The ability to revert is also good thinking, and I only wish we had a good visible affordance for "revert."  Maybe the reload/revert icon?

I'd title this commit "Enable user to to edit _trace set labels from within the Trace Viewer list of traces" 

Here's an edit-in-place sketch, showing various states in the process:
image

Everyone reading this is invited to join this whiteboard (Figjam). There you can zoom in, comment, vote, augment, correct, educate. It's free, and you can join via this link: https://www.figma.com/file/5nxXdb198DkTwT24ZyqLkv/TraceViz-Figjam-Edit-in-place?node-id=0%3A1

Key points, in priority order:
2. Indicate focus state with standard cyan border around text entry box  
3. Pencil icon and darker bg appear on hover of label (rationale: imagine a list of many trace sets: do we need to clutter the screen with many pencils?)
3. Text entry box should have enough padding so it doesn't collide/overlap
4. Add the "Edit trace label" item to the contextual menu

ebugden:
@thefinaljob look over the commit message guidelines in the new [How to contribute code](https://github.com/theia-ide/theia-trace-extension#how-to-contribute-code) readme section. Some specific issues:

- [ ]  Title should be in imperative
- [ ] Fixed issue should be linked to the PR so the issue automatically closes when the PR is merged. Issue is linked automatically when you include "Fixes #384" in the commit message (preferred) or you can do this manually via Github UI. 

More thoughts to come! Thanks for the PR 🍰

----

> **ssmagula:** I'd title this commit "Enable user to to edit _trace set labels from within the Trace Viewer list of traces"

I second this! Even though #384 was the motivation for these changes, the main action implemented in this PR is changing the trace set label (the tab changing names is a logical consequence of relabeling the trace set not the main action).

> **mirsky-work:** I think we should also add a "Rename Trace Tab" entry to the Opened Traces context menu.

I think a context menu entry isn't critical at the moment since the action is already possible. However, if an entry is added it should be "Change trace label" or "Relabel trace" (for the same reason as above).

Ideally, renaming tabs would eventually also be possible through the tab's context menu, but @ssmagula made the good point that this should be a Theia contribution not a trace-viewer-extension only thing.

packages/base/src/signals/signal-manager.ts
```diff
@@ -43,6 +44,9 @@ export const Signals = {
 };
 
 export class SignalManager extends EventEmitter implements SignalManager {
+    fireTabChangedSignal(tabName: string, expirementUUID: string): void {
+        this.emit(Signals.TABNAME_CHANGED, {tabName, expirementUUID});
```
arfio, line 48: Should be `{ tabName, experimentUUID });` (typo + space separating the curly braces) and the typo should be corrected in the argument

packages/react-components/src/trace-explorer/menu-item-trace.tsx
```diff
@@ -0,0 +1,147 @@
...
+    experimentUUID: string;
+    traces: Trace[];
+    onTabNameChange: (tabEditingOpen: string, index: number) => void;
+    menuItemTraceContainerClassName: string;
+    handleClickEvent: (event: React.MouseEvent<HTMLDivElement>, experimentName: string) => void;
+    handleContextMenuEvent: (event: React.MouseEvent<HTMLDivElement>, experimentUUID: string) => void;
+}
+interface MenuItemState {
+    editingTab: boolean;
+    oldexpirementName: string;
```
arfio, line 17: Incorrect lower camel case + typo `oldExperimentName`

packages/react-components/src/trace-explorer/menu-item-trace.tsx
```diff
@@ -0,0 +1,147 @@
...
+    expirementNameState: string;
```
arfio, line 18: All `expirement` should be changed in `experiment`

packages/react-components/src/trace-explorer/trace-explorer-opened-traces-widget.tsx
```diff
@@ -235,15 +228,33 @@ export class ReactOpenTracesWidget extends React.Component<ReactOpenTracesWidget
...
         this.setState({ openedExperiments: remoteExperiments, selectedExperimentIndex: selectedIndex !== -1 ? selectedIndex : 0 });
         signalManager().fireOpenedTracesChangedSignal(new OpenedTracesUpdatedSignalPayload(remoteExperiments ? remoteExperiments.length : 0));
     }
 
+    protected handleTabNameUpdate(tabEditingOpen: string, index: number): void {
+        const exp = this.state.openedExperiments[index];
+        exp.name = tabEditingOpen;
+        const experimentOpenNew = this.state.openedExperiments;
+        experimentOpenNew[index] = exp;
+        this.setState({openedExperiments: experimentOpenNew});
```
arfio, line 255: Should have spaces `{ openedExperiments: experimentOpenNew }`

packages/react-components/src/trace-explorer/trace-explorer-opened-traces-widget.tsx
```diff
@@ -235,15 +228,33 @@ export class ReactOpenTracesWidget extends React.Component<ReactOpenTracesWidget
...
+            });
+        });
+
         const selectedIndex = remoteExperiments.findIndex(experiment => this._selectedExperiment &&
             experiment.UUID === this._selectedExperiment.UUID);
         this.setState({ openedExperiments: remoteExperiments, selectedExperimentIndex: selectedIndex !== -1 ? selectedIndex : 0 });
         signalManager().fireOpenedTracesChangedSignal(new OpenedTracesUpdatedSignalPayload(remoteExperiments ? remoteExperiments.length : 0));
     }
 
+    protected handleTabNameUpdate(tabEditingOpen: string, index: number): void {
```
arfio, line 250: All functions or events related to experiment name change should not refer to the tab.
`submitNewTab` -> `submitNewExperimentName`
`handleTabNameUpdate` -> `handleExperimentNameUpdate`
`fireTabChangedSignal` -> `fireExperimentChangedSignal`
`TABNAME_CHANGED` -> `EXPERIMENT_CHANGED`
`onTabNameChange` -> `onExperimentNameChange`
`doHandleTabeNameChange` -> `doHandleExperimentNameChange`
`tabEditingOpen` -> `newExperimentName`

arfio (changes requested):
Thank you for the contribution! I think it is a much needed improvement in terms of usability, where most traces do not have meaningful names.

> 1. An open tab's name doesn't change until it's closed and then reopened. After that it does update while typing. See video below.
This issue is related to #364 and #450. I don't think this pull request makes the issue worse.
Also, if you reopen the trace directly from the files it will focus on the tab that did not change and stay on it when you select the menu item. 
I agree with all the changes requested before, and added some comments that are mainly coding style changes.

bhufmann (changes requested):
Thank you very much for the contribution. Will be a useful feature for the user.

I have some comments about the functionality. 

Firstly, right now the change of the name is done purely on the front-end side. The back-end (trace-server) doesn't know about the changes. When disconnecting and re-connecting the front-end the old name is back. 
To fix that, the name has to be propagated to the trace server so that the change is persisted for re-connection or connection using a different client. A fullstack update (front-end, TSP, and back-end) is needed to support that.

Secondly, we need to be clear that the renaming of experiment name doesn't change any file resources (i.e. the trace files)., and hence using this to keep a "before"-trace to be compared with an "after"-trace for trace comparison purposes won't work. This compare use case has to be addressed differently so that on the trace server each trace file of the experiment is kept. I think a deep copy of the experiment on the server would be needed instead of renaming the existing experiment.

bhufmann:
> > **bhufmann:** Firstly, right now the change of the name is done purely on the front-end side. The back-end (trace-server) doesn't know about the changes. When disconnecting and re-connecting the front-end the old name is back.
> > To fix that, the name has to be propagated to the trace server so that the change is persisted for re-connection or connection using a different client. A fullstack update (front-end, TSP, and back-end) is needed to support that.
> 
> True. I agree that the trace set label should be propagated to the server. **Do these additional changes need to be prepared before this PR is merged?** I'm divided, since I agree that merging just the front end changes is incomplete, but I think it still improves most use cases (re-connection or connection using a different client are less common).
> 
I think that persistence is important. Closing the trace viewer while keeping a trace on the server is quite common. So, when opening the trace viewer again I personally would expect, that the renamed trace name is shown.

> > **bhufmann:** Secondly, we need to be clear that the renaming of experiment name doesn't change any file resources (i.e. the trace files)., ...
> 
> I agree that it's not clear that changing the trace set label doesn't affect the filesystem and that this is confusing. But **I don't think resolving this is within the scope of this PR.** At the moment, trace set generation is all implicit and behind the scenes so the first step would be to to introduce the concept of "trace sets" that have no connection to the filesystem. If necessary, the changes proposed below could be made before merging this PR.
> 
> In the short term (new issues?):
> 
>     * **Better default trace set name:** The difference between the filesystem and the trace set would be more clear if the trace set label wasn't exactly the folder name (when automatically generating a trace set when a folder containing CTF traces), but this can be addressed in a different issue/PR. For example: "Trace Set 0 - _Generated from trace-folder-name_" instead of just "trace-folder-name".
> 
This could be the description and shown as tooltip. Using "Trace Set 0", "Trace Set 1" is not necessary better than using the directory name. The directory name right now works well, for example, when opening a trace set from the LTTng session root directory and the trace set name becomes the session name. Another use case is, when the root directory is a test case name and opening the trace set name become the unit test name. Changing the name to "Trace Set 0" as default, would break these use cases. 

I'm wondering if a dialog box should open with text box after selecting "Open with Trace Viewer". It opens with the default name. If the use wants to keep it, then he only needs to press ok. Otherwise the use can type a new name and then press enter.

>     * **Explicitly show file "Refresh"** button next to trace set (or toast notification) when traces files are overwritten. This clearly communicates that a copy of the trace files was not made.
> 
> 
Some indication is needed and a refresh needs to be done. Right now with the local trace server implementation, the trace server only has a symbolic link to the original trace file in the file system. If the original trace file is overridden, then the trace index, state systems etc (all trace analysis files) are out-dated as well and need to be re-built.

> > ... and hence using this to keep a "before"-trace to be compared with an "after"-trace for trace comparison purposes won't work. This compare use case has to be addressed differently so that on the trace server each trace file of the experiment is kept. I think a deep copy of the experiment on the server would be needed instead of renaming the existing experiment.
> 
> Good point that these changes don't make it easier to manage trace files when doing a before/after comparison (as in, users would still have to manually move/rename the "before" traces before generating the "after" traces). I don't think comparison is within the scope of this issue and I don't think renaming the trace set is the only indicator of "I want to keep this version of these trace files", but more thought definitely needs to go into comparison and whether/when to make copies of the actual trace files (especially since it's likely that a portion of users will trace using scripts that output trace files that always have the same name).
> 
Yes, Trace comparison is not scope of this PR. But it's somehow related when it comes to manage the trace files and I think it was important to mention it here. 

> Thank you for this relevant feedback!
No problem!

MatthewKhouzam (commented):
Reminder, commit messages need to be in the present tense.

thefinaljob:
@arfio has been working on getting the re-render issue for the menu completed. If we can't get it fixed shortly, we might have to live with this and submit another issue stating that the menu needs to be re-tooled so that it works with react components as menu items.

ebugden:
I'm going to make edits to the title and description of this PR to correct style issues.

ebugden:
## Storing label names in the trace server
> **bhufmann:** I think the back-end support can be implemented outside of the scope of this PR. However, it needs to be done shortly after.

I created an issue #474 so the required additional changes can be defined there. However, it's unlikely that they would be made soon. At the moment @arfio is working on finishing this PR when they have free time, but they don't have much free time these days. 

## Clarify that re-labeling the trace doesn't change the trace files
> **bhufmann:** Secondly, we need to be clear that the renaming of experiment name doesn't change any file resources (i.e. the trace files)., 

I agree it's best clarify this before merging this PR. I moved this discussion into its own issue #475

thefinaljob:
Given that reworking the entire menu is taking. a while I suggest we live with the current rendering present (currently the user needs to resize the menu in order to reflect the most recently selected trace. Other than that, all of the requested changes have been implemented

- [x] Commit has been modified per the contribution guidelines
- [x] Typos have been fixed as per @arfio's feedback
- [x] You can easily click between menu items and their changes are reflected
- [x] "Trace: " now always appears 
- [x] @ssmagula's and @mirsky-work's feedback regarding the styling has been implemented

I also made a modification to the selected item colour as having it as blue with the black bar had too much of a contrast between them.

bhufmann (changes requested):
> ## Storing label names in the trace server
> 
> > **bhufmann:** I think the back-end support can be implemented outside of the scope of this PR. However, it needs to be done shortly after.
> 
> I created an issue #474 so the required additional changes can be defined there. However, it's unlikely that they would be made soon. At the moment @arfio is working on finishing this PR when they have free time, but they don't have much free time these days.


As I said above the back-end should be available soon after this PR is ready. Merging without having a proper back-end support I'd like to avoid.

BTW, this PR needs to be rebased to master. As well as it' seems to have some styling issue in light theme:
image

ebugden:
@thefinaljob Thank you for the update!

I agree with @bhufmann on not merging as it is now. I'm uncomfortable with merging a feature feels unfinished when the timeline of when it will be completed is unclear since currently no one is actively working on finalizing this feature. 

At this stage in the project we're working on making the tool feel more stable and polished and adding this feature as it is now would be a step backwards for polish. Allowing users to label traces is convenient, but doesn't add enough user value to justify making the tool feel less polished.

PR #452, opened 2021-08-19
Null check annotation categories

MatthewKhouzam:
Avoid a null pointer access if no annotation categories are provided. (e.g. with the flame chart)

Fixes front-end issue with 438

### DIFF
--- a/packages/react-components/src/trace-explorer/trace-explorer-views-widget.tsx
+++ b/packages/react-components/src/trace-explorer/trace-explorer-views-widget.tsx
@@ -133,7 +133,7 @@ export class ReactAvailableViewsWidget extends React.Component<ReactAvailableVie
             const annotationCategories = await tspClient.fetchAnnotationsCategories(this._selectedExperiment?.UUID, output.id);
             const annotationCategoriesResponse = annotationCategories.getModel();
             if (annotationCategories.isOk() && annotationCategoriesResponse) {
-                const categoriesList = annotationCategoriesResponse.model.annotationCategories;
+                const categoriesList = annotationCategoriesResponse.model ? annotationCategoriesResponse.model.annotationCategories : [];
                 signalManager().fireAnnotationsFetchedSignal(categoriesList);
             }
         }


MatthewKhouzam (commented):
Fix Commit message

packages/react-components/src/trace-explorer/trace-explorer-views-widget.tsx
```diff
@@ -133,7 +133,7 @@ export class ReactAvailableViewsWidget extends React.Component<ReactAvailableVie
             const annotationCategories = await tspClient.fetchAnnotationsCategories(this._selectedExperiment?.UUID, output.id);
             const annotationCategoriesResponse = annotationCategories.getModel();
             if (annotationCategories.isOk() && annotationCategoriesResponse) {
-                const categoriesList = annotationCategoriesResponse.model.annotationCategories;
+                const categoriesList = annotationCategoriesResponse.model?.annotationCategories;
```
PatrickTasse, line 136: It still causes an exception later at reception of the signal.

I think we still need to fire the signal to let the users know there are no categories, either by replacing with empty array or having the signal parameter possibly undefined.

PatrickTasse (changes requested):

packages/react-components/src/trace-explorer/trace-explorer-views-widget.tsx
MatthewKhouzam: Yeah, I was doing that before, ESLint told me to do this instead, Thanks!

MatthewKhouzam: Please check the updated version of the patch.

bhufmann (approved):
LGTM

PatrickTasse (changes requested):

packages/react-components/src/trace-explorer/trace-explorer-views-widget.tsx
```diff
@@ -133,7 +133,7 @@ export class ReactAvailableViewsWidget extends React.Component<ReactAvailableVie
             const annotationCategories = await tspClient.fetchAnnotationsCategories(this._selectedExperiment?.UUID, output.id);
             const annotationCategoriesResponse = annotationCategories.getModel();
             if (annotationCategories.isOk() && annotationCategoriesResponse) {
-                const categoriesList = annotationCategoriesResponse.model.annotationCategories;
+                const categoriesList = annotationCategories ? [] : annotationCategoriesResponse.model.annotationCategories;
```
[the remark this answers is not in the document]
MatthewKhouzam: Got a point!

packages/react-components/src/trace-explorer/trace-explorer-views-widget.tsx
```diff
@@ -133,7 +133,7 @@ export class ReactAvailableViewsWidget extends React.Component<ReactAvailableVie
             const annotationCategories = await tspClient.fetchAnnotationsCategories(this._selectedExperiment?.UUID, output.id);
             const annotationCategoriesResponse = annotationCategories.getModel();
             if (annotationCategories.isOk() && annotationCategoriesResponse) {
-                const categoriesList = annotationCategoriesResponse.model.annotationCategories;
+                const categoriesList = annotationCategories ? annotationCategoriesResponse.model.annotationCategories : [];
```
bhufmann, line 136: line 135 does annotationCategories.getModel(). So, there is already a null pointer problem and the check here is not needed. You need to verify it earlier.

bhufmann (changes requested):

packages/react-components/src/trace-explorer/trace-explorer-views-widget.tsx
PatrickTasse: annotationCategories is a TspClientResponse < GenericResponse < AnnotationCategoriesModel > >
annotationCategoriesResponse is a GenericResponse < AnnotationCategoriesModel >
annotationCategoriesResponse.model is a AnnotationCategoriesModel that is nullable.

It is that last field (annotationCategoriesResponse.model) that needs a null check.

PatrickTasse: github swallows brackets in review comments like if they were HTML tags. sigh.

PatrickTasse (changes requested):

PatrickTasse (approved):

bhufmann (approved):

PR #461, opened 2021-08-27
upgrade all dependencies + Theia 1.17.0

paul-marechal:
Use `lerna@^4.0.0`.
Run `yarn upgrade`.
Fix build errors due to the new version of `tsc`.
Change TS compilation target to `ES2017` to work with Theia 1.17.0.
Remove `@theia/cpp-debug` from the browser example as it seems to be
incompatible with Theia 1.17.0.


theia-extensions/viewer-prototype/src/node/trace-server-service.ts
```diff
@@ -1,57 +1,65 @@
...
+        }
+        // TODO: Since the resolve never actually gets returned, we need a better way to determine if the child process
+        // initiated by the spawn successfully starts the trace server.
+        const server = spawn(path, ['-vmargs', `-Dtraceserver.port=${port}`]);
+        if (server.pid === undefined) {
+            // When pid is undefined this means we're about to get an error:
+            return new Promise<never>((resolve, reject) => server.once('error', reject));
+        }
+        this.server = server;
+        const timeouts: NodeJS.Timeout[] = [];
```
bhufmann, line 25: Is this change required for the upgrade or is it an improvement? I think it should be outside of the scope of this PR of upgrading Theia then. Could you have that as a separate PR after the upgrade?

bhufmann (changes requested):

examples/browser/package.json
```diff
@@ -17,25 +17,24 @@
...
+    "@theia/navigator": "1.17.0",
+    "@theia/process": "1.17.0",
+    "@theia/terminal": "1.17.0",
+    "@theia/editor": "1.17.0",
+    "@theia/markers": "1.17.0",
+    "@theia/monaco": "1.17.0",
+    "@theia/messages": "1.17.0",
+    "@theia/vsx-registry": "1.17.0",
+    "@theia/keymaps": "1.17.0",
+    "@theia/getting-started": "1.17.0",
```
gbodeen, line 33: On these lines and related lines, it may be preferable to keep the package.json entries set to `latest` and instead update only the lockfile entries to 1.17.0.

paul-marechal: Then everytime we run a `yarn` upgrade we get bitten by a Theia upgrade as well... I'm more in favor of explicitly targeting what we plan on using. Which reminds me that I should update this to target 1.17.2 now.

theia-extensions/viewer-prototype/src/node/trace-server-service.ts
paul-marechal: Upgrading TypeScript caused an issue with some of the previous code, and that caused me to drive-by this code as well... I'll provide a dirty fix instead that just tells TypeScript to ignore the previous errors for now I guess.

examples/browser/package.json
paul-marechal: there's also the issue that Theia's versioning can be quite problematic as it follows romantic version rather than semantic. So if we target `latest` but only build against 1.17.2 in this repository because of the lockfile, then the day someone pulls that Trace extension when 1.18.0 is released with breaking changes there's no telling if the extension will keep working...

bhufmann: I'm ok to be more explicit about the Theia version that the theia-trace-extension is developed against. In the end, when we upgrade to new Theia version, we need to update the version numbers and do yarn upgrade. So, not much more work.

bhufmann:
@paul-marechal I was running the browser app on my laptop I get the toast message below:

image

This seems to be introduced with this PR. I did "yarn clean"  and "git clean -xdf" before to make sure that I have clean build. Could you please have look?

vince-fugnitto:
@paul-marechal the `git` builtin likely needs to be updated due to https://github.com/eclipse-theia/theia/pull/8911#issue-546523069
- https://open-vsx.org/api/vscode/git/1.52.1/file/vscode.git-1.52.1.vsix

paul-marechal:
Upgraded all builtin plugins from 1.45.1 to 1.50.0 and specifically the `vscode-builtin-git` to 1.52.1.

examples/browser/package.json
paul-marechal: Especially if you prefer to do an upgrade of everything but Theia first, and then upgrade Theia like you asked me to do for this PR. If you target `latest` then doing `yarn upgrade` will do everything at once.

bhufmann:
> Upgraded all builtin plugins from 1.45.1 to 1.50.0 and specifically the `vscode-builtin-git` to 1.52.1.
> 
> @bhufmann the error should be fixed.

@paul-marechal yes it's fixed. Thanks!

I see the error print-outs in the console which I don't remember seeing before. Any ideas?

root ERROR [hosted-plugin: 325] Promise rejection not handled in one second: Error: Command with id 'extension.js-debug.clearAutoAttachVariables' is not registered. , reason: Error: Command with id 'extension.js-debug.clearAutoAttachVariables' is not registered.
root ERROR [hosted-plugin: 325] With stack trace: Error: Command with id 'extension.js-debug.clearAutoAttachVariables' is not registered.
    at CommandRegistryMainImpl.$executeCommand (file:///home/eedbhu/git/theia-training/theia-trace-extension/examples/electron/lib/vendors-node_modules_theia_plugin-ext_lib_hosted_browser_hosted-plugin_js.bundle.js:1413:19)
    at RPCProtocolImpl.doInvokeHandler (file:///home/eedbhu/git/theia-training/theia-trace-extension/examples/electron/lib/vendors-node_modules_theia_plugin-ext_lib_common_disposable-util_js-node_modules_theia_plugin-cba7ee.bundle.js:1020:23)
    at RPCProtocolImpl.invokeHandler (file:///home/eedbhu/git/theia-training/theia-trace-extension/examples/electron/lib/vendors-node_modules_theia_plugin-ext_lib_common_disposable-util_js-node_modules_theia_plugin-cba7ee.bundle.js:1005:41)
    at RPCProtocolImpl.receiveRequest (file:///home/eedbhu/git/theia-training/theia-trace-extension/examples/electron/lib/vendors-node_modules_theia_plugin-ext_lib_common_disposable-util_js-node_modules_theia_plugin-cba7ee.bundle.js:969:33)
    at RPCProtocolImpl.receiveOneMessage (file:///home/eedbhu/git/theia-training/theia-trace-extension/examples/electron/lib/vendors-node_modules_theia_plugin-ext_lib_common_disposable-util_js-node_modules_theia_plugin-cba7ee.bundle.js:933:26)
    at file:///home/eedbhu/git/theia-training/theia-trace-extension/examples/electron/lib/vendors-node_modules_theia_plugin-ext_lib_common_disposable-util_js-node_modules_theia_plugin-cba7ee.bundle.js:857:48
    at file:///home/eedbhu/git/theia-training/theia-trace-extension/examples/electron/lib/bundle.js:56184:69
    at CallbackList.invoke (file:///home/eedbhu/git/theia-training/theia-trace-extension/examples/electron/lib/bundle.js:56190:26)
    at Emitter.fire (file:///home/eedbhu/git/theia-training/theia-trace-extension/examples/electron/lib/bundle.js:56302:29)
    at file:///home/eedbhu/git/theia-training/theia-trace-extension/examples/electron/lib/vendors-node_modules_theia_plugin-ext_lib_common_disposable-util_js-node_modules_theia_plugin-cba7ee.bundle.js:1063:37

bhufmann:
@paul-marechal I just merged #462 which upgrades lerna to 4.0.0 as well. Please rebase your PR to pick up these changes. 

BTW, any news about the error above https://github.com/theia-ide/theia-trace-extension/pull/461#issuecomment-915540535?

paul-marechal:
@bhufmann I just came back the rabbit hole of why the error is displayed, and my conclusion is that it is not important.

`vscode-builtin-debug-auto-launch@1.50.0` tries to call a command id `extension.js-debug.clearAutoAttachVariables` which should be contributed by `js-debug@1.50.0`, but the latter extension is not yet activated so the command is not yet contributed. This doesn't look like a race condition either, so not sure what should happen in that case.

@vince-fugnitto Is there a special semantic to invoking a command like `extension.X.Y` where we should force activate an extension? That's the only thing that comes to mind... Should Theia do anything special here?

paul-marechal:
@bhufmann I rebased the PR.

bhufmann:
> @bhufmann I just came back the rabbit hole of why the error is displayed, and my conclusion is that it is not important.
> 
I don't see any side effects for the theia-trace-extension itself. Only running the example application (browser and Electron) has that issue. I'm ok to merge the pull request.

bhufmann:
> `vscode-builtin-debug-auto-launch@1.50.0` tries to call a command id `extension.js-debug.clearAutoAttachVariables` which should be contributed by `js-debug@1.50.0`, but the latter extension is not yet activated so the command is not yet contributed. This doesn't look like a race condition either, so not sure what should happen in that case.
> 
> @vince-fugnitto Is there a special semantic to invoking a command like `extension.<ext-name>.<command-id>` where we should force activate an extension? That's the only thing that comes to mind... Should Theia do anything special here?

@vince-fugnitto and @paul-marechal could please follow-up this to see if there is an issue with debug?

bhufmann (approved):
LGTM. Thanks for the upgrade.

MatthewKhouzam (approved):
Ran it here, it works. Thanks!

bhufmann:
@paul-marechal are you providing another PR for the trace server start procedure that I've asked to remove from this PR? Please let me know.

PR #462, opened 2021-08-27
Add support for publishing packages and theia-trace-extension to NPM

### LINKED ISSUE #429: Release to npm
First of all, thank you for contributing this great extension!!!

Do you have plans to publish it to npm? This would make consuming it much easier...

bhufmann:
- Use lerna 4.0.0
- Add publishing scripts in package.json
- Update related information in relevant package.json files
- Add github action to publish next version using --canary lerna option
- Add README.md files for each published package

The publish commands will publish the following packages to NPM:
- traceviewer-base
- traceviewer-react-components
- theia-traceviewer

Fixes #429


packages/base/README.md
```diff
@@ -0,0 +1,9 @@
+## Description
+
+The Trace Viewer base package contains trace management utilities for managing traces Trace Server applications implementing the Trace Server protocol. While being initially used within the Theia Trace Viewer extension, its code base is independent to any Theia APIs and hence can be integrated in other web applications.
```
mirsky-work, line 3: Something is missing in that sentence:

> managing traces <exported by?> Trace Server applications

packages/base/LICENSE
```diff
@@ -0,0 +1,19 @@
+Copyright (c) 2019 Ericsson and others
```
mirsky-work, line 1: Nitpicking: shouldn't be 2019-2021?

bhufmann: It was copied from the root directory. I'll update all the Lisense files.

examples/browser/package.json
```diff
@@ -32,7 +32,7 @@
     "@theia/vsx-registry": "latest",
     "@theia/keymaps": "latest",
     "@theia/getting-started": "latest",
-    "theia-trace-viewer": "0.0.0"
+    "@trace-viewer/theia-trace-extension": "0.0.1"
```
marcdumais-work, line 35: I do not think this scope this will work, to publish under the `theia` npm user

"Each npm user/organization has their own scope, and only you can add packages in your scope."
https://docs.npmjs.com/cli/v7/using-npm/scope

So, I think you can either create your own npm "trace-viewer" user/scope and publish under that user, use the `@theia` scope or use no scope and publish under the `theia` npm user.

marcdumais-work: (assuming this is a package you intend to publish - I did not look in-depth)

bhufmann: > (assuming this is a package you intend to publish - I did not look in-depth)

Thanks Marc for the information. We'd like to publish the 3 following packages:

-     \@trace-viewer/base
-     \@trace-viewer/react-components
-     \@trace-viewer/theia-trace-extension

We used \@trace-viewer to indicate that they below together. But I don't actually want to create that scope. So I guess I remove the scope an rename the packages "base" and "react-components" to something clearer.

marcdumais-work: Yes, that would work. Even smt like: 
`traceviewer-base`
`traceviewer-react-components`
`traceviewer-trace-extension` (is it really an extension? if a Theia extension, you could use e.g. `@theia/trace-viewer-extension` - it's allowed to mix and match, packages with and without a scope)

bhufmann: Your suggestions are good proposals. However, removing the scope from @trace-viewer/base and @trace-viewer/react-components many source code files need to be changed to use new names. Not clear yet the way forward I should take. I'll need some more thinking.

bhufmann: I think the following name suitable:

`traceviewer-base`
`traceviewer-react-components`
`theia-traceviewer`

marcdumais-work: > theia-traceviewer

is this a pure JS library or a Theia extension?  If the later, I would go with `@theia/traceviewer`.

MatthewKhouzam (approved):
Works on my machine! :)

examples/browser/package.json
marcdumais-work: Your name will work too, so at the end of the day, it's your decision.

Rodrigoplp-work (approved):

PR #471, opened 2021-09-07
Add Marker Rows

muddana-satish:


MatthewKhouzam (commented):
please fix the lint errors, then I'll try it out.

packages/react-components/src/components/timegraph-output-component.tsx
```diff
@@ -384,7 +500,105 @@ export class TimegraphOutputComponent extends AbstractTreeOutputComponent<Timegr
...
+    }
+
     private getStateStyle(state: TimelineChart.TimeGraphState) {
+
+        if (state.data && state.data.annotation) {
+            const style = this.getAnnotationStyle(state.data.annotation);
+            return {
+                color: style ? style.color : 0x000000,
+                opacity: style ? style.opacity : 1.0,
+                height: 10,
```
PatrickTasse, line 596: I think the height of the state should be computed using its row height. It would allow for bigger and easier to read labels.

packages/react-components/src/components/timegraph-output-component.tsx
```diff
@@ -384,7 +500,105 @@ export class TimegraphOutputComponent extends AbstractTreeOutputComponent<Timegr
...
+                borderWidth: 0,
+                borderColor: 0x000000
```
PatrickTasse, line 598: The border fields are optional, they could just be omitted.

packages/react-components/src/components/timegraph-output-component.tsx
```diff
@@ -384,7 +500,105 @@ export class TimegraphOutputComponent extends AbstractTreeOutputComponent<Timegr
...
+        };
+    }
+
     private getStateStyle(state: TimelineChart.TimeGraphState) {
+
+        if (state.data && state.data.annotation) {
+            const style = this.getAnnotationStyle(state.data.annotation);
+            return {
+                color: style ? style.color : 0x000000,
+                opacity: style ? style.opacity : 1.0,
```
PatrickTasse, line 595: In Trace Compass we don't use the opacity in the marker axis (only in the time graph itself), so they are fully opaque. If we want to achieve the same, you can just omit the opacity.

packages/react-components/src/components/timegraph-output-component.tsx
```diff
@@ -384,7 +500,105 @@ export class TimegraphOutputComponent extends AbstractTreeOutputComponent<Timegr
...
+                label: event.label,
+                data: {
+                    style: event.data?.style,
+                    annotation: event
+                }
+            };
+            states.push(state);
+        });
+
+        const defaultRow = {
```
PatrickTasse, line 530: I wish the default row would be smaller in height, but I don't think it's possible right now in timeline-chart...

packages/react-components/src/components/timegraph-output-component.tsx
```diff
@@ -216,20 +278,49 @@ export class TimegraphOutputComponent extends AbstractTreeOutputComponent<Timegr
...
+                    className="timegraph-tree"
+                />
+            </div>
+            <div ref={this.markerTreeRef} onScroll={_ev => { this.synchronizeTreeScroll(); }}
+                className='output-component-tree'
+                style={{ height: this.getMarkersLayerHeight() }}>
+                <EntryTree
+                    collapsedNodes={this.state.collpasedAnnotationNodes}
+                    showFilter={false}
+                    entries={this.state.annotationEntries}
```
PatrickTasse, line 314: I'm not sure where it happens, but it seems to me like the close button and category label are centered horizontally in the tree. I think they should be left- or right- aligned.

PatrickTasse (changes requested):

packages/react-components/src/components/timegraph-output-component.tsx
```diff
@@ -384,7 +500,105 @@ export class TimegraphOutputComponent extends AbstractTreeOutputComponent<Timegr
         };
     }
 
+    private updateMarkersData(rangeEvents: TimelineChart.TimeGraphAnnotation[], newRange: TimelineChart.TimeGraphRange, newResolution: number) {
+        const annotationEntries: Entry[] = [];
+        const markers: Map<string, TimelineChart.TimeGraphState[]> = new Map();
+        const rows: TimelineChart.TimeGraphRowModel[] = [];
+        const filteredEvents = rangeEvents.filter(event => this.annotationMarkers?.includes(event.category));
```
PatrickTasse, line 507: We might want to also filter the events that do not overlap with the current view range. The model is fetched for a wider range than the viewable range.

However this would need to be done when the view range changes, not when fetching the model... A row could appear and disappear as you browse in your cached data's range without ever updating the model.

packages/react-components/src/components/timegraph-output-component.tsx
PatrickTasse: Actually, when you have contiguous markers of the same color, it might be hard (impossible?) to see where the bounds are, because the label is drawn in the center of the state. To be checked. A possible solution would be to use a border on the states to make their bounds very obvious.

packages/react-components/src/components/timegraph-output-component.tsx
PatrickTasse: This might actually be required because for marker sets, sometimes alternate markers (e.g. odd markers) have the same common marker color as the even marker, but with 0.0 opacity. That would make them invisible in the marker row. To be checked...

PatrickTasse (changes requested):

PR #472, opened 2021-09-07
Filter XY chart type

### LINKED ISSUE #296: Hide "analyses" that do not provide value to the user

ebugden:
Currently, XY charts are missing y-axis labels and units so it is
impossible for users to learn how to interpret the data shown in these
charts using the tool. Without a sense of scale, XY charts have little
analysis use and mainly distract users from views that are ready to be
used (ex. Timegraph views, Events table). Filter out XY chart views to
correct this.

For developers who need to test or demonstrate new features for XY
charts, the condition that filters XY charts
(value.type !== 'TREE_TIME_XY') can be temporarily removed to allow
testing XY chart features.

Fixes #296

### DIFF
--- a/packages/react-components/src/trace-explorer/trace-explorer-views-widget.tsx
+++ b/packages/react-components/src/trace-explorer/trace-explorer-views-widget.tsx
@@ -171,7 +171,7 @@ export class ReactAvailableViewsWidget extends React.Component<ReactAvailableVie
         if (descriptors && descriptors.length) {
             outputDescriptors.push(...descriptors);
         }
-        return outputDescriptors.filter(value => value.type !== 'DATA_TREE');
+        return outputDescriptors.filter(value => value.type !== 'DATA_TREE' && value.type !== 'TREE_TIME_XY');
     }
 }
 


ebugden:
I almost submitted this as a draft, but I thought these changes were too much of a good idea ☀️ Let me know what you think!

It's a tradeoff of showing a less accurate picture of the amount of work that has been completed on the tool in exchange for cleaning things up in preparation for users.

CC: @ssmagula

packages/react-components/src/trace-explorer/trace-explorer-views-widget.tsx
```diff
@@ -171,7 +171,7 @@ export class ReactAvailableViewsWidget extends React.Component<ReactAvailableVie
         if (descriptors && descriptors.length) {
             outputDescriptors.push(...descriptors);
         }
-        return outputDescriptors.filter(value => value.type !== 'DATA_TREE');
+        return outputDescriptors.filter(value => value.type !== 'DATA_TREE' && value.type !== 'TREE_TIME_XY');
     }
```
bhufmann, line 175: I'm against removing the XY charts as a whole. Event if they have short-comings they have already value, e.g. CPU usage view or histogram. Besides the charts are being improved right now regarding navigation. I suggest to fix the most pressing short-comings (e.g. y-labels, tooltips) instead of hiding them.

bhufmann (changes requested):
see comment https://github.com/theia-ide/theia-trace-extension/pull/472/files#r704653017

packages/react-components/src/trace-explorer/trace-explorer-views-widget.tsx
ebugden: What is a concrete example of the value they offer without having any sense of scale? I gave this some thought before proposing the patch and didn't have any luck:
* Since the data is displayed in a line you can see the shape of the data, which in theory could allow you to use this as a reference for where to explore more in detail using a timegraph chart, but without scale there's no way of knowing whether spikes in the graph are just noise. 
* I could imagine an experienced trace compass user might know whether the shape of a graph looks typical or not in specific situations, but this feels like a stretch.

bhufmann: @ebugden I suggest that the pressing short-comings are fixed instead of removing the whole support of such graphs.

ebugden:
Closing because this PR is no longer relevant. The linked issue #296 has been resolved by several other PRs (for details, read as of this [comment](https://github.com/theia-ide/theia-trace-extension/issues/296#issuecomment-932650175)).

PR #476, opened 2021-09-10
Make tooltips always on.

### LINKED ISSUE #296: Hide "analyses" that do not provide value to the user

MatthewKhouzam:
Partially fixes #296

### DIFF
--- a/packages/react-components/src/components/xy-output-component.tsx
+++ b/packages/react-components/src/components/xy-output-component.tsx
@@ -145,6 +145,10 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
             },
             maintainAspectRatio: false,
             legend: { display: false },
+            tooltips: {
+                intersect: false,
+                mode: 'index'
+                },
             layout: {
                 padding: {
                     left: 0,


bhufmann (changes requested):
I doesn't work for me. I have to click into the graph to get the tooltip. Moreover, when selecting many series, the placement of the tooltip window is mostly outside, below the graph and not visible. Only the top of the window is visible.

ebugden:
Could you remove #296 from "Linked issues" PR? I don't want the issue to close if this PR is merged.

----

I would say unless having the **tooltip visible on hover** and using the **human-readable labels** (ex. "sudo" rather than "apt:2723") are 2 second fixes, my reflex would be to not put more dev effort into adding tooltips. And even with those fixes I'm not sure adding this without more polish is a good idea. If there's more time to invest in XY charts, the priority is having a y-axis.

## PR Goals
The goal with this PR was to spend minimal dev effort and quickly give users at way to:
* **Read absolute values** (to compensate for XY charts not having a y-axis at the moment)
* **Associate lines with labels** (since there's no legend at the moment)

## Thoughts
### Not great for reading absolute values 
* As a scale reference, tooltips aren't as good as a y-axis: 
   * No units in the tooltip
   * Tooltips only show the value of one point of the chart at a time and they require interacting with the chart to see the scale. Users would have to remember the scale of the chart and need to check it again every time they zoom. For example, glancing back and forth between several points in different charts to compare doesn't really work with tooltips. 
* Difficult to use: It's hard to select sudden peaks/dips (requires extremely accurate mouse mouvement) without zooming and you can't zoom directly on the XY charts right now (have have a timeline chart open next to it and zoom with that). Having the tooltip visible on hover would help with this (but not fix completely).

### Doesn't work as a legend
* As is it doesn't help with this since the process labels are just numbers. Using the human-readable labels would fix this. 
image

### General

* Makes the tool feel more messy: Adds another element that needs to be polished/adjusted eventually
   * Tooltips on other charts don't move when the selected time changes. You have to move all the tooltips individually.
   * Style is different from the timegraph tooltip ( having two different tooltip styles looks unfinished)

Tooltips not moving with selected time:

https://user-images.githubusercontent.com/28311615/133470002-88336ac6-deee-4028-aeaa-901ca76ee06c.mp4

PR #477, opened 2021-09-13
Fix token name for publishing to NPM

bhufmann:


marcdumais-work (approved):
LGTM!

PR #482, opened 2021-09-15
Only run publish-next build job for theia-ide/theia-trace-extension

bhufmann:
This check added prevents the publish job to be run for forks of this
project.

### DIFF
--- a/.github/workflows/publish-next.yml
+++ b/.github/workflows/publish-next.yml
@@ -6,6 +6,7 @@ on:
 #    types: [created]
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


MatthewKhouzam (approved):
LGTM, good catch. I'm kinda nervous if we don't merge this.

Rodrigoplp-work (approved):

bhufmann:
> LGTM, good catch. I'm kinda nervous if we don't merge this.

As long as the fork doesn't have write permission to  the NPM package, it's doesn't matter. Also, actions could be disabled in the forks. Nevertheless it's a good practice to only allow the main repository to publish.

PR #484, opened 2021-09-15
improve the trace server spawning logic

paul-marechal:
It is often easy to write JS code that works in general cases, but miss corner cases... Especially with the process API which I believe was designed to be the most painful thing to use.

Some quirks include:
- `pid` is `undefined` iff Node couldn't actually spawn the process,
- `childProcess.exitCode` or `childProcess.signalCode` are set when the process exits,
- `childProcess.pid` can outlive the process, so issuing a kill on it after the process actually exited may end up killing another process that ends up inheriting the same `pid` some time later.

There were some code oddities like ``spawn(`${path}`, ...)`` which most likely was some kind of workaround around the fact that `path` can be `undefined`... But in such a case it would try to do `spawn('undefined', ...)` which doesn't make any sense. Until someone figures out what it means to pass `undefined` to this method, we'll throw an error.

I cannot guaranty that this new implementation is 100% foolproof, but it's definitely more robust.

### DIFF
--- a/theia-extensions/viewer-prototype/src/node/trace-server-service.ts
+++ b/theia-extensions/viewer-prototype/src/node/trace-server-service.ts
@@ -1,60 +1,91 @@
-import { spawn } from 'child_process';
+import { spawn, ChildProcess } from 'child_process';
 import { injectable } from 'inversify';
 import { PortBusy, TraceServerConfigService } from '../common/trace-server-config';
 import treeKill = require('tree-kill');
 
+const SUCCESS = 'success';
+
 @injectable()
 export class TraceServerServiceImpl implements TraceServerConfigService {
-    private processId: number;
 
-    startTraceServer(path: string | undefined, port: number | undefined): Promise<string> {
-        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-        const server = spawn(path!, ['-vmargs', `-Dtraceserver.port=${port}`]);
-        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-        this.processId = server.pid!;
-        const timeouts: NodeJS.Timeout[] = [];
+    private server?: ChildProcess;
+    private port?: number;
 
-        return new Promise<string>((resolve, reject) => {
-            // TODO: Since the resolve never actually gets returned, we need a better way to determine if the child process
-            // initiated by the spawn successfully starts the trace server
-
-            // If the server doesn't error or exit within 2 seconds, we consider that a success.
-            // That doesn't mean that the server has really started. On the frontend, the `TraceServerConnectionStatusService`
-            // will ping the port until it receives a response, which is the official measure of success.
-
-            timeouts.push(setTimeout(() => resolve('success'), 2000));
-
-            // If the server exits or errors before it outputs, consider it a failure.
-            server.on('error', err => {
-                reject(err);
+    async startTraceServer(path?: string, port?: number): Promise<string> {
+        if (path === undefined) {
+            throw new Error('no Trace Server path specified');
+        } else if (port === undefined) {
+            throw new Error('no Trace Server port specified');
+        } else if (this.isServerRunning(this.server)) {
+            if (this.port === port) {
+                return SUCCESS;
+            } else {
+                throw new Error('the Trace Server is already running on a different port');
+            }
+        }
+        const server = spawn(path, ['-vmargs', `-Dtraceserver.port=${port}`]);
+        if (server.pid === undefined) {
+            // When pid is undefined it usually means we're about to get an error.
+            return new Promise<never>((_, reject) => server.once('error', reject));
+        }
+        this.server = server;
+        this.port = port;
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        let timeout: any;
+        try {
+            await new Promise<void>((resolve, reject) => {
+                // If the server doesn't error or exit within 2 seconds, we consider that a success.
+                // That doesn't mean that the server has really started. On the frontend, the `TraceServerConnectionStatusService`
+                // will ping the port until it receives a response, which is the official measure of success.
+                timeout = setTimeout(resolve, 2000);
+                server.once('exit', (code, _signal) => reject(PortBusy(code)));
+                server.once('error', reject);
             });
-            server.on('exit', code => {
-                reject(PortBusy(code));
-            });
-            // If no response recieved from the trace server in 10 seconds, reject with an error (for internal use)
-            timeouts.push(setTimeout(() => reject('Waited 10 seconds but nothing happened and hence exiting'), 10000));
-
-        }).finally(() => {
+        } catch (error) {
+            this.server = undefined;
+            this.port = undefined;
+            throw error;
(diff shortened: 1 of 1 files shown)


theia-extensions/viewer-prototype/src/node/trace-server-service.ts
```diff
@@ -1,60 +1,65 @@
...
-            server.on('error', err => {
-                reject(err);
-            });
-            server.on('exit', code => {
-                reject(PortBusy(code));
+    async startTraceServer(path?: string, port?: number): Promise<string> {
+        if (path === undefined) {
+            throw new Error('no trace server path specified');
+        } else if (this.isServerRunning(this.server)) {
+            throw new Error('trace server is already running');
```
paul-marechal, line 15: Should this function return `success` if the server is already running?

bhufmann (dismissed):
I verified the updates and my comments were addressed. Looks good to me. Thanks for the updates.

theia-extensions/viewer-prototype/src/node/trace-server-service.ts
```diff
@@ -1,60 +1,88 @@
...
+        this.port = port;
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        let timeout: any;
+        try {
+            await new Promise<void>((resolve, reject) => {
+                // If the server doesn't error or exit within 2 seconds, we consider that a success.
+                // That doesn't mean that the server has really started. On the frontend, the `TraceServerConnectionStatusService`
+                // will ping the port until it receives a response, which is the official measure of success.
+                timeout = setTimeout(resolve, 2000);
+                // If the server exits or errors before it outputs, consider it a failure.
```
colin-grant-work, line 41: This comment doesn't quite apply, since we're not waiting for output, we're just waiting on a timeout.

theia-extensions/viewer-prototype/src/node/trace-server-service.ts
```diff
@@ -1,60 +1,88 @@
...
+        } finally {
             server.removeAllListeners();
-            timeouts.forEach(timeout => clearTimeout(timeout));
-        });
-
+            // The process exited so we can clean our internal state.
+            server.once('exit', () => {
+                this.server = undefined;
+                this.port = undefined;
+            });
```
colin-grant-work, line 51: Is this guaranteed to run even if the server has already exited (e.g. if we arrive here because we rejected the Promise due to an exited event)? I don't think it is, and since we have checks for whether the server is running that don't require us to remove these fields, and this listener isn't guaranteed to remove them, I don't this listener is necessary? Or, if we want this code to run when the server exits, it should be put in the exit handler defined in the Promise callback.

theia-extensions/viewer-prototype/src/node/trace-server-service.ts
```diff
@@ -1,60 +1,88 @@
...
-                    return;
+                } else {
+                    resolve();
                 }
-                resolve('success');
             });
         });
+        const timeout = new Promise((_, reject) => {
+            setTimeout(reject, 1000, new Error('the Trace Server did not exit'));
+        });
```
colin-grant-work, line 77: The Promise logic here is a little odd:
 - Create a Promise that checks for exit
 - `await` a Promise that actually does the exiting. If this succeeds, aren't we guaranteed to have resolved the first Promise?
 - Create a new Promise with a timeout - we only get here if we successfully `await`ed `treeKill`, so we've probably already exited?
 - Race promises 1 & 3

Some comments might be useful to understand why it's done this way, if this is the right way to do it.

colin-grant-work (commented):
I know this comes late - sorry!

theia-extensions/viewer-prototype/src/node/trace-server-service.ts
paul-marechal: It's a weird one, the comment comes from the previous code but the way it seem to work is that this API is supposed to spawn the trace server, and the service telling if the server is healthy is located somewhere else and is listening to the output.

This is because this method only cares about spawning the process, but the process then can take its sweet time before actually running the server logic (e.g. initialization time past process creation).

So I kept the same logic as before:

- Spawn the server process
- Wait 2 seconds
- If the process exits for any reason during those 2s then something went wrong

theia-extensions/viewer-prototype/src/node/trace-server-service.ts
paul-marechal: It is to use `server.on('exit', ...)` as source of truth: As you said `treeKill` should kill the process, so `exit` should be fired. Since we relay on that `exit` event to clear the state of the class, we want to make sure that it is correctly invoked.

theia-extensions/viewer-prototype/src/node/trace-server-service.ts
paul-marechal: Good catch, this was bad logic. This listener must only be added if the server process "successfully" starts.

theia-extensions/viewer-prototype/src/node/trace-server-service.ts
colin-grant-work: You're right that the comment is from the old code, but it was actually invalid even in the old code. When that code was being drafted, there was a time when it listened for output on the server's stdio, and that's what this comment is referring to. Before that code was merged, the output listener was removed here (invalidating the comment), and a two-part startup process was implemented on the frontend that (1) asked for the server to be started, and then (2) waited for a separate service that pings the server to declare success. It looks like the logic has been changed around a little, though, so now if the `startServer` promise resolves, the frontend tries to execute the code that requires it immediately.

https://github.com/theia-ide/theia-trace-extension/blob/64a63a19c75596bfe7f7099a6790f05f2b6e1956/theia-extensions/viewer-prototype/src/browser/trace-viewer/trace-viewer-contribution.ts#L74-L96

https://github.com/theia-ide/theia-trace-extension/blob/64a63a19c75596bfe7f7099a6790f05f2b6e1956/theia-extensions/viewer-prototype/src/browser/trace-explorer/trace-explorer-sub-widgets/trace-explorer-placeholder-widget.tsx#L82-L104

paul-marechal: TIL thanks for explaining how it works now, will update the comment. I'll leave it up to the Trace people to deal with that!

theia-extensions/viewer-prototype/src/node/trace-server-service.ts
colin-grant-work: It looks like you've relocated this listener after the `finally`, but I think that leaves the same problem as before: if the `exit` event fires while we're `await`ing the Promise, then this listener will be applied too late.

colin-grant-work: Touche ⚔️ . I missed the re`throw` in the `catch` block 👍.

colin-grant-work (approved):
The logic looks good to me.

bhufmann (approved):
Thanks @colin-grant-work for valuable review. I re-tested it and code looks good to me. Thanks @paul-marechal for the contribution.

PR #486, opened 2021-09-16
Refresh output views on selecting experiments

### LINKED ISSUE #457: Opening a view that does not exist in the trace
The ls-ust trace has no resource status view but I was still able to open it in this trace 

image

**To reproduce:**

1. open a trace that has the **resource status** view **(example: apt)** and open another trace that does not have this view **(ls-ust)**.
2. go to the **apt** trace and open **resource status**  
3. open the **ls-ust** trace  and for a few moments you can select **resource status view**  and select it. 

view

muddana-satish:
fixes #457


IbrahimFradj (changes requested):
Now we can't open a view that doesn't exist in another trace. thanks for this contribution. but i noticed that when i press a view, all the views disappear for a little while and when i change the trace all the views disappear too but for a longer time. 

**you can see it here** 
bugsatish

I think that the views should not disappear when we select a view

muddana-satish:
> I think that the views should not disappear when we select a view

That's because the api is fetching the results for the newly selected experiment. Do you mean we should show some kind of loading icon?

IbrahimFradj:
I thought that the available views should not be reloaded when I select a view, but that they could be reloaded if I select another trace. 

> 
> 
> > I think that the views should not disappear when we select a view
> 
> That's because the api is fetching the results for the newly selected experiment. Do you mean we should show some kind of loading icon?

I thought that the available views should not be reloaded when I select a view, but that they could be reloaded if I select another trace.

mirsky-work:
I agree with Ibrahim's comment - can we reload the analyses list only when selecting an experiment, and not an analysis?

In addition, leaving the list of analyses visible when a trace tab is closed is confusing - it should be cleared when a tab is closed.

pr486-refresh-analyses

bhufmann (dismissed):
Looks good to me. Thanks!

bhufmann:
> I agree with Ibrahim's comment - can we reload the analyses list only when selecting an experiment, and not an analysis?
> 
> In addition, leaving the list of analyses visible when a trace tab is closed is confusing - it should be cleared when a tab is closed.
> 
> pr486-refresh-analyses

I opened a new issue tracker to follow-up with this discussion (see #489).

bhufmann:
> > I agree with Ibrahim's comment - can we reload the analyses list only when selecting an experiment, and not an analysis?
> > In addition, leaving the list of analyses visible when a trace tab is closed is confusing - it should be cleared when a tab is closed.
> > pr486-refresh-analyses
> 
> I opened a new issue tracker to follow-up with this discussion (see #489).

Sorry, I misunderstood the comment of @IbrahimFradj. After reading it again I now understand what he meant. I also agree that the list of available should not be reloaded when a view is clicked on. Looking at the source code, this happens because after clicking a view, the method `doHandleOutputClicked` does call `signalManager().fireExperimentSelectedSignal(this._selectedExperiment);` which on the other hand calls `doExperimentSelected` that @muddana-satish just changed.

I'm removing my approval and ask that this is addressed in the next update of this PR.

packages/react-components/src/trace-explorer/trace-explorer-views-widget.tsx
```diff
@@ -149,6 +149,7 @@ export class ReactAvailableViewsWidget extends React.Component<ReactAvailableVie
 
     protected doExperimentSelected(experiment: Experiment | undefined): void {
         this._selectedExperiment = experiment;
+        this.setState({availableOutputDescriptors: []});
```
bhufmann, line 152: This method is also called after clicking a view by the method `doHandleOutputClicked` through the  call `signalManager().fireExperimentSelectedSignal(this._selectedExperiment);` This chain of events causes the list to be cleared and refreshed when opening a view. Could you please make sure that the list is not refreshed in that case?

bhufmann (changes requested):

PR #487, opened 2021-09-17
Partial Gitpod fix: Add keytar (new Theia dependency)

ebugden:
This PR is only a partial fix because Gitpod breaks in two places right now:
- [x]  The build fails because keytar is missing. Fixed by this PR.
- [ ]  The frontend fails to connect to the server #488. @bhufmann will propose a solution to this by fixing #493

Details:
* Both issues (build fail, server connection fail) aren't present before the Theia version was upgraded.
* To observe the fontend not being able to connect to the server on Gitpod, you need to open this PR's branch (the extension doesn't build in Gitpod on the most recent version of master).

Commit a3a4002 "upgrade to Theia 1.17.0" in PR theia-ide#461 upgraded the version
of Theia used by the trace viewer extension. The new version of Theia
now requires keytar. This new dependency is missing in the gitpod setup
so the extension build fails in gitpod.

Update gitpod setup to install keytar (dependency) and fix gitpod
build.

### DIFF
--- a/.gitpod.dockerfile
+++ b/.gitpod.dockerfile
@@ -11,6 +11,8 @@ RUN sudo apt-get update \
     && sudo apt-get install -y libgtk-3-0 libnss3 libasound2 libgbm1 \
     # native-keymap
     && sudo apt-get install -y libx11-dev libxkbfile-dev \
+    # keytar
+    && sudo apt-get install -y libsecret-1-dev \
     && sudo rm -rf /var/lib/apt/lists/*
 
 ENV NODE_VERSION="12.14.1"
--- a/.gitpod.dockerfile
+++ b/.gitpod.dockerfile
@@ -13,6 +13,7 @@ RUN sudo apt-get update \
     && sudo apt-get install -y libx11-dev libxkbfile-dev \
     # keytar
     && sudo apt-get install -y libsecret-1-dev \
+    # Clear package info (to free space)
     && sudo rm -rf /var/lib/apt/lists/*
 
 ENV NODE_VERSION="12.14.1"


paul-marechal (approved):

IbrahimFradj (approved):

IbrahimFradj:
Would it be better to put it in 1 commit ? thanks for your contribution

ebugden:
> Would it be better to put it in 1 commit ? thanks for your contribution

The clarification isn't part of installing the package which is why I separated it. I'm in the habit of separating refactor changes from functional changes. Technically the cleanup commit could probably be a different PR, but we'll see what people say.

bhufmann:
I was testing this PR and debugging the issue of #493 and I think I understand what happens. After a small hack, I was able to run the trace server successfully in GitPod with your this PR. So, I think this PR is fine and can be merged. I'll provide a proper fix for #493 (not just a hack). @ebugden please update this patch so it's not a draft and ready for review.

bhufmann (approved):

arfio:
> I was testing this PR and debugging the issue of #493 and I think I understand what happens. After a small hack, I was able to run the trace server successfully in GitPod with your this PR. So, I think this PR is fine and can be merged. I'll provide a proper fix for #493 (not just a hack). @ebugden please update this patch so it's not a draft and ready for review.

Is it related to this piece of code by any chance ?:
```Typescript
async getTraceServerUrl(): Promise<string> {
    if (!this._traceServerUrl) {
        const traceServerUrl = await this.environments.getValue('TRACE_SERVER_URL');
        this._traceServerUrl = traceServerUrl ? this.parseUrl(traceServerUrl.value || TRACE_SERVER_DEFAULT_URL) : TRACE_SERVER_DEFAULT_URL;
    }
    return this._traceServerUrl.replace(/{}/g, this.port ? this.port : TRACE_SERVER_DEFAULT_PORT);
}
```
I changed this part of the code which seems to not have the "correct" logic. It loads the TRACE_SERVER_URL only if the default is not defined. In my mind, it should always replace the traceServerUrl if the TRACE_SERVER_URL is defined. So I just removed the if condition.

bhufmann:
@ebugden, @arfio  I uploaded a fix to be able to connect to the GitPod server #499. I temporarily added the keytar fix as well so that you can try it.

bhufmann:
I'll merge this one and then rebase #499

PR #498, opened 2021-09-22
Add the shift-click selection interval in XY chart

### LINKED ISSUE #439: The shift-click selection interval in XY charts does not work.
When i want to enlarge or reduce my selection with shift click it doesn't work in xy charts.

**Here is how it should work :**

shift click

IbrahimFradj:
We can select interval with shift-click.

**Here is the result :**

shiftclick

Fixes #439.

### DIFF
--- a/packages/react-components/src/components/xy-output-component.tsx
+++ b/packages/react-components/src/components/xy-output-component.tsx
@@ -44,24 +44,10 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
     private plugin = {
         afterDraw: (chartInstance: Chart, _easing: Chart.Easing, _options?: any) => { this.afterChartDraw(chartInstance); }
     };
-    private updateSelection = (event: MouseEvent) => {
-        const scale = this.props.viewRange.getEnd() - this.props.viewRange.getstart();
-        if (this.mouseIsDown && this.props.unitController.selectionRange && !this.isRightClick) {
-            const xStartPos = this.props.unitController.selectionRange.start;
-            let end = xStartPos + ((event.screenX - this.posPixelSelect) / this.lineChartRef.current.chartInstance.width) * scale;
-            end = Math.min(Math.max(end, 0), this.props.unitController.absoluteRange);
-            this.props.unitController.selectionRange = {
-                start: xStartPos,
-                end: end
-            };
-        }
-    };
 
     private endSelection = () => {
         if (this.isRightClick) {
            this.applySelectionZoom();
-        } else {
-            document.removeEventListener('mousemove', this.updateSelection);
         }
         this.mouseIsDown = false;
         document.removeEventListener('mouseup', this.endSelection);
@@ -342,11 +328,18 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
             this.startPositionMouseRightClick = startTime;
         } else {
             this.isRightClick = false;
-            this.props.unitController.selectionRange = {
-                start: startTime,
-                end: startTime
-            };
-            document.addEventListener('mousemove', this.updateSelection);
+            if (event.shiftKey && this.props.unitController.selectionRange) {
+                this.props.unitController.selectionRange = {
+                    start: this.props.unitController.selectionRange.start,
+                    end: startTime
+                };
+            } else {
+                this.props.unitController.selectionRange = {
+                    start: startTime,
+                    end: startTime
+                };
+            }
+            this.onMouseMove(event);
         }
         document.addEventListener('mouseup', this.endSelection);
     }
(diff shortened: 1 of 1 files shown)


ebugden:
Is this a feature in Trace Compass that current TC users expect to have and enjoy? 

If not, could you explain:
* An example where it's much more convenient to adjust an existing selected area instead of re-selecting the adjusted smaller/larger area?
* Why implementing adjusting an existing area with shift-click is the most practical/intuitive? (ex. Instead of being able to click on the bars showing the start/end of the selected area to drag them smaller/larger)

IbrahimFradj:
> 
> 
> I definitely should have asked the following questions before assigning the associated issue to "High priority". I apologize for not asking these earlier!
> 
> Is this a feature in Trace Compass that current TC users expect to have and enjoy?
> 
> If not, could you explain:
> 
>     * An example where it's much more convenient to adjust an existing selected area instead of re-selecting the adjusted smaller/larger area?
> 
>     * Why implementing adjusting an existing area with shift-click is the most practical/intuitive? (ex. Instead of being able to click on the bars showing the start/end of the selected area to drag them smaller/larger)
> 
> 
> Thanks for working on the XY chart navigation! ☀️

In fact, this functionality exists in Trace compass and in timeline chart on theia-trace-extension. my goal is that xy chart has the same functionality as timeline chart.
This can be useful when for example I find a place where I want to make a selection without changing the beginning of the selection

bhufmann:
> I definitely should have asked the following questions before assigning the associated issue to "High priority". I apologize for not asking these earlier!
> 
> Is this a feature in Trace Compass that current TC users expect to have and enjoy?
> 
> If not, could you explain:
> 
>     * An example where it's much more convenient to adjust an existing selected area instead of re-selecting the adjusted smaller/larger area?
> 
>     * Why implementing adjusting an existing area with shift-click is the most practical/intuitive? (ex. Instead of being able to click on the bars showing the start/end of the selected area to drag them smaller/larger)
> 
> 
> Thanks for working on the XY chart navigation! sunny

The advantage of this capability is that users can keep one side of selection (start or end) constant and then extend the selection without changing the starting point to measure time from a given timestamp on.

ebugden:
@IbrahimFradj Can you rebase on master so that you get the recent Gitpod fixes? I'd like to play around with this feature to see if there's a quick way we could make it discoverable (ex. changing the cursor to a ⟷ when they press shift while hovering on a graph).

Thank you for the extra context IbrahimFradj and bhufmann. I don't know the practical analysis applications of measuring times with a fixed reference, but I trust your experience.

IbrahimFradj:
> 
> 
> @IbrahimFradj Can you rebase on master so that you get the recent Gitpod fixes? I'd like to play around with this feature to see if there's a quick way we could make it discoverable (ex. changing the cursor to a ⟷ when they press shif

IbrahimFradj:
> 
> 
> @IbrahimFradj Can you rebase on master so that you get the recent Gitpod fixes? I'd like to play around with this feature to see if there's a quick way we could make it discoverable (ex. changing the cursor to a ⟷ when they press shift while hovering on a graph).
> 
> Thank you for the extra context **IbrahimFradj** and **bhufmann**. I don't know the practical analysis applications of measuring times with a fixed reference, but I trust your experience.

@ebugden  I rebased on master, gitpod works.

PatrickTasse:
> Why implementing adjusting an existing area with shift-click is the most practical/intuitive? (ex. Instead of being able to click on the bars showing the start/end of the selected area to drag them smaller/larger)

This could make it impossible to replace a narrow range selection with a new single selection in one click (without needing to click somewhere far away first).

Shift to extend selection also applies when using keyboard left/right navigation. That is really useful to make a selection to the exact time bounds of a state or group of states.

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -45,14 +45,11 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
     private updateSelection = (event: MouseEvent) => {
-        const scale = this.props.viewRange.getEnd() - this.props.viewRange.getstart();
         if (this.mouseIsDown && this.props.unitController.selectionRange && !this.isRightClick) {
             const xStartPos = this.props.unitController.selectionRange.start;
-            let end = xStartPos + ((event.screenX - this.posPixelSelect) / this.lineChartRef.current.chartInstance.width) * scale;
-            end = Math.min(Math.max(end, 0), this.props.unitController.absoluteRange);
             this.props.unitController.selectionRange = {
                 start: xStartPos,
-                end: end
+                end: this.getTimeX(event.offsetX)
```
PatrickTasse, line 52: If you drag selection outside of the bounds of the XY chart then you start receiving event.offsetX in another component's coordinates and the end time is wrong.

PatrickTasse (changes requested):

PatrickTasse (approved):

ebugden:
> I rebased on master, gitpod works.

Looks good to me! Thanks for the update. Consistent navigation features across chart types is important to avoid user confusion.

bhufmann (approved):
LGTM. Thanks!

PR #499, opened 2021-09-23
Fix handling TRACE_SERVER_URL environment variable

### LINKED ISSUE #493: GitPod not able to connect to GitPod trace server but connects to localhost server
When starting the theia-traceviewer in GitPod, the TRACE_SERVER_URL environment variable is set so that it is able to connect to the trace server that is started in GitPod. It seems that this environment variable is not taking into consideration anymore. This work before upgrading to Theia 1.17.2. Something changed in the start-up of the extension that makes it fail.

bhufmann:
Fix handling TRACE_SERVER_URL environment variable

- Make sure that switch over to the TRACE_SERVER_URL updates relevant  widgets
- Adding preference listener was moved to allow reading the environment variable TRACE_SERVER_URL before resetting the connection

Fixes #493

Note: This change doesn't change the behaviour, that the default URL is used to initialize the all the views. In order to fix that, the start-up procedure of the theia-trace-extension has to be overhauled.

### DIFF
--- a/packages/react-components/src/trace-explorer/trace-explorer-opened-traces-widget.tsx
+++ b/packages/react-components/src/trace-explorer/trace-explorer-opened-traces-widget.tsx
@@ -47,6 +47,11 @@ export class ReactOpenTracesWidget extends React.Component<ReactOpenTracesWidget
         this._experimentManager = this.props.tspClientProvider.getExperimentManager();
         this.props.tspClientProvider.addTspClientChangeListener(() => {
             this._experimentManager = this.props.tspClientProvider.getExperimentManager();
+            // new tsp-client connection... re-initialize the this widget and trac explorer
+            this.setState({ openedExperiments: [], selectedExperimentIndex: 0 });
+            this._selectedExperiment = undefined;
+            signalManager().fireOpenedTracesChangedSignal(new OpenedTracesUpdatedSignalPayload(0));
+            this.initialize();
         });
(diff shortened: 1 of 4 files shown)


ebugden:
I haven't taken the time to understand the changes, but I can confirm that Gitpod works for me using these changes!

bhufmann:
> I haven't taken the time to understand the changes, but I can confirm that Gitpod works for me using these changes!
> 
> > **bhufmann:** This change doesn't change the behaviour, that the default URL is used to initialize the all the views. In order to fix that, the start-up procedure of the theia-trace-extension has to be overhauled.
> 
> Is it important that this is eventually corrected? Should we create a low-priority issue for these additional changes?

Yes, it's important that it will be corrected. That's why I put this information there. I have planned to add a new tracker once this is merged. Updating the start-up procedure is an involving task, but necessary to improve the overall quality of the code base.

bhufmann:
@paul-marechal just uploaded #502 that fixes this problem as well, while fixing some other race conditions and while refactoring the code. I'll review it and I try to get #502 merged instead.

Please note, the extension start-up procedure is out of scope of #502.

paul-marechal:
@bhufmann note that #502 fixes the issue where a `TspClient` is first initialized with the default URL. The temporary workaround for it is the `LazyTspClient` that defers async calls to its methods until the `baseUrl` is resolved. This should do until something better is done elsewhere in the code.

PR #502, opened 2021-09-23
refactor+fix trace server url initialization

paul-marechal:
Both `TraceServerUrlProviderImpl` and `TraceViewerEnvironment` are
riddled with concurrency issues. A lot of events happen asynchronously
but the logic doesn't take this into account, leaving the application
vulnerable to race conditions when trying to fetch the trace server url.

Refactor by removing `TraceViewerEnvironment` and moving its
functionality directly into `TraceServerUrlProviderImpl`, add logic to
prevent race conditions.

Unfortunately I was forced to introduce a `LazyTspClient` which is
implemented as a proxy to a `TspClient` and defers all async method
calls until `baseUrl` is asynchronously resolved. This only works because
all `TspClient` methods are asynchronous.

The proxy is a temporary measure which should buy you time until
you guys rework the frontend initialization sequence to make sure
everyone expects and handles the asynchronous steps.

Fixes https://github.com/theia-ide/theia-trace-extension/issues/493
Closes https://github.com/theia-ide/theia-trace-extension/pull/499


paul-marechal:
Some notes about the use of the `LazyTspClient`: classes seem to want to synchronously get a reference to a `TspClient`, but all work done through it is asynchronous. All places that uses the `TspClient` deal with asynchronous calls. This makes me think that it should be possible to instead store a promise to a reference of a `TspClient` and first await on this before using its async methods. We're dealing with async stuff anyway when it gets time to use it.

packages/base/src/lazy-tsp-client.ts
```diff
@@ -0,0 +1,33 @@
...
+    // The `LazyTspClient` will just delay each call to its methods by
+    // first awaiting for the asynchronous `baseUrl` resolution which
+    // is used to then asynchronously get a valid `TspClient`.
+    const tspClientPromise = url.then(baseUrl => new TspClient(baseUrl));
+    // eslint-disable-next-line no-null/no-null
+    return new Proxy(Object.create(null), {
+        get(target, property, _receiver) {
+            let method = target[property];
+            if (!method) {
+                target[property] = method = async (...args: any[]) => {
```
bhufmann, line 25: I had to learn how the proxy works and I think I understand how it works. Basically, with your code each method call is intercepted here and forwarded to the corresponding TspClient object's method while making sure that baseUrl is set properly first. Thanks for adding the comment above that helps to understand.

One thing though, the code will fail to compile the day the TspClient has one synchronous method. Do you see any issues with that? Or do we just make sure that all new methods in the TspClient is async until the LazyTspClient is not needed anymore? Please let me know what you think.

bhufmann (commented):
It looks good to me and it works locally and in GitPod. Thanks for improvement! @paul-marechal, could you please answer my question.

packages/base/src/lazy-tsp-client.ts
paul-marechal: Yes the idea was to delay the method calls on the `LazyTspClient` until the time it is possible to instantiate a `TspClient`, which we do once the promise to an URL resolves.

On the other hand, with the way I wrote things here compilation won't fail even if you add synchronous methods... It will most likely fail at runtime. But ultimately what this means is that this code cannot handle synchronous calls, not as it is right now.

paul-marechal: I type a lot don't I.

> Or do we just make sure that all new methods in the TspClient is async until the LazyTspClient is not needed anymore?

It sucks but that's one way to go. The best would be to get rid of the `LazyTspClient` before the time you need to add synchronous methods to the `TspClient` ;)

Otherwise on a per-case basis it might be possible to try and intercept the calls to the synchronous APIs and try to provide a somewhat safe value/function, but that starts to become smelly.

packages/base/src/lazy-tsp-client.ts
```diff
@@ -0,0 +1,55 @@
...
+export type LazyTspClient = {
+    [K in keyof TspClient]: TspClient[K] extends (...args: infer A) => (infer R | Promise<infer R>)
+        ? (...args: A) => Promise<R>
+        : never // discard field
+};
+
+export const LazyTspClientFactory = Symbol('LazyTspClientFactory');
+export type LazyTspClientFactory = (url: Promise<string>) => LazyTspClient;
+
+export function LazyTspClientFactoryImpl(url: Promise<string>): IsTspClientCompatible<LazyTspClient> {
```
paul-marechal, line 36: I just added typings to ensure the compilation fails as soon as the `LazyTspClient` is no longer compatible with the `TspClient` type. This will most likely happen once a synchronous method or a property is added to `TspClient`.

packages/base/src/lazy-tsp-client.ts
```diff
@@ -0,0 +1,55 @@
...
+
+/**
+ * Hack!
+ * When using actual classes as a type, TypeScript cares about private fields.
+ * In our case we only care about public APIs, not internal fields.
+ * This type only extracts public APIs out of `TspClient`.
+ */
+export type ITspClient = {
+    [K in keyof TspClient]: TspClient[K]
+};
```
paul-marechal, line 13: Having `TspClient` be a class with private properties makes it a bit annoying to mock.

packages/base/src/lazy-tsp-client.ts
bhufmann: I tried it with yarn link "tsp-typescript-client" and a synchronous method, and it fail at compile time. That's better than runtime.

packages/base/src/lazy-tsp-client.ts
paul-marechal: See [illustration](https://www.typescriptlang.org/play?#code/MYGwhgzhAEAqEAcDCICWBTAdgF2gbwFgAoaU6YAe0wmwCcBXYbC2gCmLM+gVtQDcw2dNABGkdAFVaIAFzQavTAHMOZAJT4AvqtKQAnpmDQAtumwALCgBNWauQAVaFY6gjoAPHwqorAPi3E2kTEAGb0htioVNDYYADW6PDIaFjYrKAYOHJJKJnYdtBePgHBRAD0ZdBIYJgA5LjokKggeiYUwHFwiLmpoujAYPRu0BQhMebCPPyCwiEYIFYAdMvElNS4xu1x2d0pONAAvPg60PqGJmaWNhqEJFykaxAUIOiLIBRKrLWbHehWtWoTkEgsQKtAAEL0XCobC1GCYCi4NAJaAAd2EEEs9AW0HQAA8EP1cISKAgXjEKOQwCAQNBUDhKRZXNxeAIhNA5ugFgBCVZUGjkPa4I6YdCorrJPJfMAygGrIWLMRuKQgADc0DB6FoTloxCAA).

packages/base/src/lazy-tsp-client.ts
paul-marechal: Oh that's surprising, cool.

paul-marechal: Found it. It caused an error when forcing the type `as TspClient`. I moved types around so that the error happens in the factory instead of at call sites.

paul-marechal:
PR is good for re-review/merging again!

bhufmann (approved):
I tested it locally (with default URL) and in GitPod where a custom URL via the TRACE_SERVER_URL environment variable is used. Both work well. Thanks for the contribution.

PatrickTasse (approved):

PatrickTasse (approved):

PatrickTasse (dismissed):

paul-marechal:
Ok I am ashamed of continuously removing your approvals but this wasn't just nitpicking.

The value returned by `getTraceServerUrlPromise()` used to be set once, and any subsequent call would only ever return the initial value. This is now fixed.

Thanks for bearing with me!

bhufmann (changes requested):
now I found something :-) Could you please update it?

theia-extensions/viewer-prototype/src/browser/trace-server-url-provider-frontend-impl.ts
```diff
@@ -1,66 +1,126 @@
...
-     * Add a listener for trace server url changes
-     * @param listener The listener function to be called when the url is
-     * changed
-     */
-    addTraceServerUrlChangedListener(listener: (url: string) => void): void {
-        this._listeners.push(listener);
+    protected normalizeUrl(url: string): string {
+        url = url.toLowerCase();
+        // Add missing http protocol.
+        if (!url.endsWith('http://') || !url.endsWith('https://')) {
```
bhufmann, line 107: This if statement needs to be like this:
if (!url.startsWith('http://') && !url.startsWith('https://'))

bhufmann (changes requested):
The normalizeUrl() method change in one of the previous commits causes that it won't work in GitPod anymore. See comment and suggestion below.

ebugden:
Thank you all for working on this! 🌳

theia-extensions/viewer-prototype/src/browser/trace-server-url-provider-frontend-impl.ts
paul-marechal: I always get this wrong, good catch.

bhufmann: Even with the original code it would have failed now because GitPod uses https now

bhufmann (approved):

PatrickTasse (approved):

PR #504, opened 2021-09-27
Add zoom shortcut +/- for TIMEGRAPH and XY CHART

IbrahimFradj:
We can now zoom with +/- or =/_ in timegraph and xy chart.

### DIFF
--- a/packages/react-components/src/components/trace-context-component.tsx
+++ b/packages/react-components/src/components/trace-context-component.tsx
@@ -273,10 +273,27 @@ export class TraceContextComponent extends React.Component<TraceContextProps, Tr
     }
 
     render(): JSX.Element {
-        return <div className='trace-context-container' ref={this.traceContextContainer}>
+        return <div className='trace-context-container'
+                onKeyDown={event => this.onKeyDown(event)}
+                ref={this.traceContextContainer}>
             <TooltipComponent ref={this.tooltipComponent} />
             {this.props.outputs.length ? this.renderOutputs() : this.renderPlaceHolder()}
         </div>;
+    }
+
+    private onKeyDown(key: React.KeyboardEvent) {
+        switch (key.key) {
+            case '+':
+            case '=': {
+                this.zoomButton(true);
+                break;
+            }
+            case '-':
+            case '_': {
+                this.zoomButton(false);
+                break;
+            }
+        }
     }
 
     private renderOutputs() {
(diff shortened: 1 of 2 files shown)


packages/react-components/src/components/trace-context-component.tsx
```diff
@@ -273,12 +273,29 @@ export class TraceContextComponent extends React.Component<TraceContextProps, Tr
     }
 
     render(): JSX.Element {
-        return <div className='trace-context-container' ref={this.traceContextContainer}>
+        return <div className='trace-context-container' tabIndex = {0}
```
PatrickTasse, line 276: This 'tabIndex' causes a blue border to be drawn when the trace tab area has focus (including extra blank space below opened outputs).

The border is not very nice because it's partially hidden by the time axis when an output is opened.

Further, when no output is opened, the blank area can still be selected, and you can zoom in and out (e.g. change the view area) even if no time graph or xy chart is visible. I don't know if this is desirable behavior. Note that it's the same when click the +/-/home toolbar buttons with no output opened.

Without this tabIndex, the blue border appears only around the time graph or xy chart that is selected, and the +/- keys only change the view range when such an element is selected. It seems to me like this might be the better behavior.

PatrickTasse (changes requested):

packages/react-components/src/components/trace-context-component.tsx
IbrahimFradj: I thought of putting this in the trace context because we don't need a view to be selected since zooming is done for all views, zooming is not linked to the mouse, just to the range. would it be better if I made shortcuts in the respective classes?

PatrickTasse: It appears that the trace context receives the key events from any child component that has keyboard focus. So it's fine that it handles the common shortcuts for every type of output component, as long as the shortcut applies to every child component that is allowed to have keyboard focus.

We might have an issue if one of these components wants to override or disable the shortcut, or wants to accept keys e.g. to type a string. Then it will have to figure out how to prevent the event from propagating further.

PatrickTasse (approved):

marco-miller (changes requested):
Typo in commit message: `timegraphe` should read `timegraph`.

PatrickTasse (approved):

bhufmann (approved):
Thanks. Looks good to me.

bhufmann:
> Typo in commit message: `timegraphe` should read `timegraph`.

This was addressed in the last commit.

PR #508, opened 2021-09-30
Remove scatter data providers (latency scatter charts)

bhufmann:
All latency scatter XY charts have in common that each series don't
have a common x axis in comparison to the Histogram which have multiple
series where each series have the same x values the same number of
y-values.

The theia-traceviewer currently doesn't have support of drawing such
scatter charts properly and should be hidden.

The data provider descriptor, however, doesn't provide a way to
determine if it's a xy-chart with common x-Axis or not, and hence it's
not possible to generically hide such graphs.

This PR hides such graphs based on the data provider ID, which contain
the string "scatter" in it the Trace Compass case all with
org.eclipse.tracecompass.internal.analysis.timing.core.segmentstore.scatter.dataprovider.

Even if this is Trace Compass server specific, it will improve
the UX for such users. This stop gap implementation will be removed
once the front-end properly supports such graphs or there is a generic
way to filter-out unwanted data provider.

Contributes to fixing of #296

### DIFF
--- a/packages/react-components/src/trace-explorer/trace-explorer-views-widget.tsx
+++ b/packages/react-components/src/trace-explorer/trace-explorer-views-widget.tsx
@@ -164,7 +164,17 @@ export class ReactAvailableViewsWidget extends React.Component<ReactAvailableVie
         if (descriptors && descriptors.length) {
             outputDescriptors.push(...descriptors);
         }
-        return outputDescriptors.filter(value => value.type !== 'DATA_TREE');
+        /**
+         * Remove outputs of type DATA_TREE since the view is not supported in theia-traceviewer
+         */
+        let outputs = outputDescriptors.filter(value => (value.type !== 'DATA_TREE'));
+        /*
+         * Remove certain scatter specific outputs since they are not supported in the theia-traceviewer
+         * There is no generic way to identify such outputs. This is not ideal because other server
+         * implementations are not covered.
+         */
+        outputs = outputs.filter(value => !value.id.includes('scatter'));
+        return outputs;
     }
 }
 


ebugden (dismissed):
Suggested:
* Add "Closes #472" to the commit message
* Create issue that describes changes necessary for generic filtering (if you post the information, I can write the issue)

Thank you for this change! I am 🌈 SO EXCITED 🤸 about this PR! It's too bad we have to use a server-specific hack for now, but this is such an important cleanup step on the journey towards a tool that feels ready to be a reliable work horse. It's easily in the top 3 tool improvements this year (up there with automatic server management)!

bhufmann:
Just an update. I'm currently hesitant to merge this one. I did some investigations and understand more about what is going wrong. #399 would provide a way to show "Latency vs Time" views as scatter chart which would solve the issue. 

Patch for bug https://bugs.eclipse.org/bugs/show_bug.cgi?id=576647, will fix a problem in the Trace Compass server.

So, we are improving and hopefully we have better support for such views instead of removing them. Let's see how things evolve in the next couple of weeks.

ebugden:
>  I'm currently hesitant to merge this one.

Why can't we remove them for now and then revert the changes when things are fixed in a couple weeks? These non-functional views are arguably **the most visible issue** with the tool right now and this PR is ready to go.

The user survey was just sent this morning so in the next couple weeks we're hoping to have a lot of eyes on the tool. I strongly believe it would be best to merge this to improve the situation now because these views are: 
* **Unreadable:** These views cannot be used when the data is visualized as lines since it's discrete not continuous data
* **Unstable:** Sometimes they spin infinitely unless the window is refreshed

## User survey risk

Out of the 5 people who have responded to the user survey so far, we've already received 2 comments relevant to this PR:
* One person pointed out that some views seem to be unavailable (spin infinitely), but can still be selected by users
* One person commented on the general lack of stability  

If we leave obvious and easily fixable issues we're already aware of in the tool:
* We reduce the value of the survey by distracting the users (who could notice issues that we aren't aware of)
* We risk damaging the reputation of the tool by giving a bad impression to users

This seems like especially unnecessary risk since this PR is ready to go (apart from a second review).

ebugden:
@MatthewKhouzam @mirsky-work @ssmagula Could you weigh in on this? I would argue that this is the most important PR on the project right now.

mirsky-work:
> @MatthewKhouzam @mirsky-work @ssmagula Could you weigh in on this? I would argue that this is the most important PR on the project right now.

No strong opinion, but I tend to agree with Bernd. It seems things are moving in the right direction, so maybe we can avoid this trace-compass-server specific hack. I think it's reasonable to have some unstable features at this point, as we haven't officially released an MVP.

ssmagula:
Ideally, even the earliest iterations of a new product will provide usefulness to the user and won't leave users with the impression that the tool is perhaps a bit unreliable or unpredictable. The reasoning is something like: 
- When a tool gives users the impression of unreliability or unpredictability (even if it's _merely_ the impression!), users might have a harder time trusting the tool's authority overall
- Early adopters might have an outsized effect on a product's reputation 
- Early users can provide important feedback, so it's valuable to have repeat customers

That's my point of view in an ideal case. In this specific case, I'll defer to Bernd on whether a temporary patch or what is best right now.

ebugden:
Thank you for taking the time to look at this PR! 🌼

MatthewKhouzam (approved):
I'm OK either way

packages/react-components/src/trace-explorer/trace-explorer-views-widget.tsx
```diff
@@ -171,7 +171,17 @@ export class ReactAvailableViewsWidget extends React.Component<ReactAvailableVie
...
+        /**
+         * Remove outputs of type DATA_TREE since the view is not supported in theia-traceviewer
+         */
+        let outputs = outputDescriptors.filter(value => (value.type !== 'DATA_TREE'));
+        /*
+         * Remove certain Trace Compass specific outputs since they are not supported in the theia-traceviewer
+         * There is no generic way to identify such outputs. This is not ideal because other server
+         * implementations are not covered.
+         */
+        outputs = outputs.filter(value => !value.id.startsWith('org.eclipse.tracecompass.internal.analysis.timing.core.segmentstore.scatter.dataprovider'));
```
MatthewKhouzam, line 183: contains "scatter" ?

bhufmann: sure we can do that as well.

MatthewKhouzam (approved):

PatrickTasse (approved):

PR #512, opened 2021-10-01
Rename signals for zoom-in, zoom-out & reset zoom and related methods

bhufmann:
Initially the zooming capability was only supported for time graphs and it made somewhat sense to reference 'timegraph' in the names. However, after supporting it globally it's better to use generic names.


IbrahimFradj (approved):
Looks good to me. Thank you.

PatrickTasse (approved):

PR #514, opened 2021-10-05
Fix endless loop when opening certain XY Charts.

### LINKED ISSUE #513: Deadlock when opening certain XY Charts
Using the Trace Compass Server, the trace `apt` from the tutorial traces (see gitPod) and when opening the `SystemCall Analysis - Latency vs Time` the application ends in a deadlock and the application is not responsive anymore, hence it has to be "killed".

bhufmann:
Don't allow children nodes with same ID as parent when creating the
XY tree.

Fixes #513

### DIFF
--- a/packages/react-components/src/components/utils/filtrer-tree/utils.tsx
+++ b/packages/react-components/src/components/utils/filtrer-tree/utils.tsx
@@ -31,7 +31,9 @@ export const listToTree = (list: Entry[], headers: ColumnHeader[]): TreeNode[] =
         if ((entry.parentId !== undefined) && (entry.parentId !== -1)) {
             const parent: TreeNode = lookup[entry.parentId];
             if (parent) {
-                parent.children.push(node);
+                if (parent.id !== node.id) {
+                    parent.children.push(node);
+                }
             } else {
                 // no parent available, treat is as root node
                 node.isRoot = true;


packages/react-components/src/components/utils/filtrer-tree/utils.tsx
```diff
@@ -31,7 +31,9 @@ export const listToTree = (list: Entry[], headers: ColumnHeader[]): TreeNode[] =
         if ((entry.parentId !== undefined) && (entry.parentId !== -1)) {
             const parent: TreeNode = lookup[entry.parentId];
             if (parent) {
-                parent.children.push(node);
+                if (parent.id !== node.id) {
```
PatrickTasse, line 34: You may or may not want to consider indirect circular references as well...

packages/react-components/src/components/utils/filtrer-tree/utils.tsx
```diff
@@ -31,7 +31,9 @@ export const listToTree = (list: Entry[], headers: ColumnHeader[]): TreeNode[] =
         if ((entry.parentId !== undefined) && (entry.parentId !== -1)) {
             const parent: TreeNode = lookup[entry.parentId];
             if (parent) {
```
PatrickTasse, line 33: Should the condition be included here? In the case of the circular reference should the node be considered a root or not? Otherwise does it mean the node just disappears from the tree?

PR #518, opened 2021-10-08
Let timeline-chart create gap states

PatrickTasse:
Do not add gap states to the row model, but include all states including
null states (states without a style). Add the optional gap style for a
row to the row model. This will allow the timeline-chart to create its
own gaps dynamically as needed for its current view range.


PatrickTasse:
> I noticed something else on GitPod for bug_hunt experiment and the flame chart. I wonder if the end state line is drawn by a gap?

That's a problem in the new code in timeline-chart, it draws a gap before every blank state even if the end time of previous state and start time of blank state is the same. That will need to be fixed in timeline-chart.

PR #521, opened 2021-10-14
Don't execute open command if TraceViewerWidget is already open

### LINKED ISSUE #450: Clicking on trace file to change tab focus opens duplicate tab

bhufmann:
This avoids duplicate TraceViewerWidget tabs when 1) open a trace from the file explorer and 2) switch between traces in the open traces widget in the trace explorer. For 1) the widget is created with an URI (file path) and for 2) the widget is created with an existing traceUUID. Internally in Theia these causes 2 widgets to be opened because it uses the URI and the TraceViewerWidgetOptions (incl. UUID) to create a key for an internal map. For both cases the key is different.

Avoiding executing the open command makes sure that the existing widget is used, no matter which way it was opened.

Fixes #450

### DIFF
--- a/theia-extensions/viewer-prototype/src/browser/trace-explorer/trace-explorer-sub-widgets/trace-explorer-opened-traces-widget.tsx
+++ b/theia-extensions/viewer-prototype/src/browser/trace-explorer/trace-explorer-sub-widgets/trace-explorer-opened-traces-widget.tsx
@@ -1,5 +1,5 @@
 import { inject, injectable, postConstruct } from 'inversify';
-import { ReactWidget, Widget } from '@theia/core/lib/browser';
+import { ReactWidget, Widget, WidgetManager } from '@theia/core/lib/browser';
 import * as React from 'react';
 import { Experiment } from 'tsp-typescript-client/lib/models/experiment';
 import { ExperimentManager } from 'traceviewer-base/lib/experiment-manager';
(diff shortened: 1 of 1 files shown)


bhufmann:
> > Still happening. Apparently the issue only exists after opening a trace from the "Explorer" (see attached video). Traces opened from the "Trace Viewer" don't cause problems.
> 
> Thanks for testing this. So, my change only fixes it for some cases. But with your steps, my patch doesn't help. Knowing the root cause I understand why it still happens with your flow. Unfortunately, at the moment I don't know how to fix it. I'll investigate a bit further.

It's tricky to fix when opening from a file explorer because the widget is opened before the experiment is posted to the server. The code to detect and post to the server is done the the widget class and is triggered after the widget is constructed. The UUID then is created on the server it's not possible to know beforehand to prevent the open with Trace Viewer command being executed. To fix that it needs more investigation and thinking. 

I wonder if we could merge this PR first which fixes the other occurrence of this bug and have a separate issue tracker to fix the remaining bug? @Rodrigoplp-work please let me know what you think.

PR #523, opened 2021-10-15
Display the tooltip for xy

IbrahimFradj:
The tooltip is displayed when you move in the view.

**Here is the result :**

tooltip
 Continuation of the RP #476

However, there are issues:
- If I select several processes the tooltip gets bigger and you can't see it in the view anymore
**Example :**

tooltip

-  If we manage to put the right position for the tooltip, there will be another problem with the size of the tooltip that will hide the graphic


packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -173,7 +180,7 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
                 yAxes: [{ display: false }]
             },
             animation: { duration: 0 },
-            events: [ 'mousedown' ],
+            events: [ 'mousemove', 'mouseout' ],
```
bhufmann, line 183: if you remove mousemove here, the widget with all the line series are not redrawn for every move (remove also the vertical line from afterChartDraw()), then the tooltip creation is very fast, because the chart is not redrawn, only the tooltip is rendered. 

This could be a compromise for now till we have better performance solution for drawing the vertical line. I'm not sure if I'd prefer to have the vertical line at all, because it's seems confusing to me when doing the time selection (left mouse drag) or right mouse zoom. 

Alternatively, draw the tooltip only on hover after a little delay, like the timeline-chart.

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -224,6 +231,13 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
     private afterChartDraw(chart: Chart) {
         const ctx = chart.ctx;
         if (ctx) {
+            if (this.isMouseMove) {
+                const startPixel = this.positionXMove;
+                const endPixel = this.positionXMove;
+                ctx.strokeStyle = '#9f9f9f';
+                ctx.fillStyle = '#9f9f9f';
+                this.drawSelection(chart, startPixel, endPixel);
+            }
```
bhufmann, line 240: as mentioned above, this can be removed.

packages/react-components/src/components/tooltip-xy-component.tsx
```diff
@@ -0,0 +1,62 @@
...
+                </ul>
+                {this.state.tooltipData?.zeros > 0 &&
+                    <p style={{marginBottom: 0}}>{this.state.tooltipData?.zeros} other{this.state.tooltipData?.zeros > 1 ? 's' : ''}: 0</p>
+                }
+            </div>
+        );
+    }
+
+    setElement<T>(tooltipData: T): void {
+        this.setState({ tooltipData });
```
bhufmann, line 60: what will the display look like if there a large number of non-zero lines? Will the tooltip window just grow?
We could limit the size and create a scrollbar (or more button). But this would require that the user is able to click into the tooltip and use the scrollbar.

bhufmann (changes requested):

packages/react-components/src/components/xy-output-component.tsx
Rodrigoplp-work: Good way to solve the problem. Done.

packages/react-components/src/components/tooltip-xy-component.tsx
Rodrigoplp-work: Done.

packages/react-components/src/components/xy-output-component.tsx

packages/react-components/src/components/tooltip-xy-component.tsx
```diff
@@ -0,0 +1,101 @@
...
+                    if (this.state.tooltipData !== tooltipData || this.state.tooltipData === undefined) {
+                        this.setState({ tooltipData: { opacity: 0 }});
+                    }
+                }, 500);
+            }
+        }
+
+        // Delay show
+        setTimeout(() => {
+            this.setState({ tooltipData });
```
PatrickTasse, line 89: I'm not sure how it happened, but when hovering around a left point I saw sometimes the tooltip info of a right point.

After a while, it just stopped showing any point (or perhaps always the first point which was y=0?).

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -426,24 +433,90 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
+                };
+                points.push(point);
+            }
+            else {
+                zeros += 1;
+            }
+        });
+
+        const tooltipData = {
+            title: this.state.xyData.labels[startPixel],
```
PatrickTasse, line 458: Unfortunately we cannot use the labels from the data model, as these are currently rounded-out integers.

I think we need to compute the timestamp bigint from the mouse position. And the timestamp should be formatted using the numberTransalator.

PatrickTasse (changes requested):

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -426,24 +433,90 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
+            dataPoints: points,
+            top: window.pageYOffset + y,
+            left: window.pageXOffset + x - 200,
```
PatrickTasse, line 461: The tooltip position is part of the data, and the tooltip is delayed by 500ms. So what happens is that the user's mouse movements are reproduced after a delay. It makes a very odd behavior. For example, draw a figure 8 with the mouse, the tooltip will also move in a figure 8 but after the delay, even if the mouse has settled on a new position.

I thought of a possible solution, to have two modes:

A) In this mode, the tooltip is updated and shown immediately as the user moves the mouse. Every time the mouse moves, a 500ms debounce is called. When the debounce expires (mouse has stopped moving), we go to mode B.

B) In this mode, we have not moved the mouse for 500 ms and are showing the tooltip. When the user moves the mouse, the tooltip remains where it is and does not update, but a 500ms debounce is called. When the debounce expires, or if the user mouses over the tooltip, we go back to mode A and start updating the tooltip again.

In other words, if the users stops moving for 500ms, then the tooltip will initially not move for 500ms when the user resumes moving.

packages/react-components/src/components/tooltip-xy-component.tsx
Rodrigoplp-work: Solved by changing the way the timeout is used to display the tooltip.

packages/react-components/src/components/xy-output-component.tsx

packages/react-components/src/components/xy-output-component.tsx
Rodrigoplp-work: Good suggestions. In B, either using a timer or a debouncer, there would always be a case where the tooltip would behave badly, like staying in one point for longer than needed, and then jumping to another far away point before returning to A.
The solution was to stop the tooltip from following the mouse. Now it shows up after the mouse stops, like the tooltip of the resources status chart.

PatrickTasse (changes requested):

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -426,24 +433,97 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
+            timeLabel = this.props.unitController.numberTranslator(timeForX);
+        }
+        const chartWidth = this.lineChartRef.current.chartInstance.width;
+        const points: any = [];
+        let zeros = 0;
+        this.state.xyData.datasets.forEach((d: any) => {
+            // In case there are less data points than pixels in the chart,
+            // calculate nearest value.
+            const relativeX= (startPixel * chartWidth) / d.data.length;
+            const adjustedX = d.data[relativeX];
```
PatrickTasse, line 451: Rename the variable, this is a yValue.

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -426,24 +433,97 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
         this.isMouseLeave = true;
         this.positionXMove = Math.max(0, Math.min(event.nativeEvent.offsetX, this.lineChartRef.current.chartInstance.width));
         this.forceUpdate();
         if (this.mouseIsDown && !this.isRightClick) {
             this.updateSelection();
         }
+
+        const x = event.nativeEvent.x;
+        const y = event.nativeEvent.y;
+        if (!this.inTooltipBox(x, y)) {
```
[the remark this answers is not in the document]
Rodrigoplp-work: Indeed. With the new debouncer and timer, this is not necessary anymore. Removed.

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -426,24 +433,97 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
         }
     }
 
+    private tooltip(x: number, y: number): void {
+        this.isComputing = true;
```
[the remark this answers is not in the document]
Rodrigoplp-work: This was helping to reduce the load on the browser, but became irrelevant after I stopped showing the tooltip on realtime.
Removed.

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -426,24 +433,97 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
+        const startPixel = this.positionXMove;
```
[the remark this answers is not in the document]
Rodrigoplp-work: Changed to "xPos".

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -426,24 +433,97 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
+        const timeForX = this.getTimeForX(startPixel);
+        let timeLabel: string  | undefined = startPixel.toString();
```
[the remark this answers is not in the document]

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -426,24 +433,97 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
+        if (this.props.unitController.numberTranslator) {
+            timeLabel = this.props.unitController.numberTranslator(timeForX);
+        }
+        const chartWidth = this.lineChartRef.current.chartInstance.width;
+        const points: any = [];
+        let zeros = 0;
+        this.state.xyData.datasets.forEach((d: any) => {
+            // In case there are less data points than pixels in the chart,
+            // calculate nearest value.
+            const relativeX= (startPixel * chartWidth) / d.data.length;
```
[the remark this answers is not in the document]
Rodrigoplp-work: Good catch. Done.

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -426,24 +433,97 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
+            const adjustedX = d.data[relativeX];
+            const rounded = isNaN(adjustedX) ? 0 : (BIMath.round(Number(adjustedX) * 100)/BIMath.round(100));
```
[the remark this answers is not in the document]
Rodrigoplp-work: The rounding is necessary to round to show the values with 2 decimal cases.
The isNaN check is necessary because without it, sometimes a small tooltip without any text appears a the top left corner of the window. I believe it happens if the mouse moves out of to chart too fast. The isNaN check avoids that.
I prefer to leave this line as it is.

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -426,24 +433,97 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
+                zeros += 1;
+            }
+        });
+        // Sort in decreasing order
+        points.sort((a: any, b: any) => Number(b.value) - Number(a.value));
+
+        const tooltipData = {
+            title: timeLabel,
+            dataPoints: points,
+            top: window.pageYOffset + y + 15,
```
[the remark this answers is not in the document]
Rodrigoplp-work: Indeed. Fixed.

packages/react-components/src/components/xy-output-component.tsx

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -426,24 +432,93 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
+        const chartWidth = this.lineChartRef.current.chartInstance.width;
+        const arraySize = this.state.xyData.labels.length;
+        const index = Math.max(Math.round((xPos / chartWidth) * (arraySize - 1)), 0);
+        const points: any = [];
+        let zeros = 0;
+        this.state.xyData.datasets.forEach((d: any) => {
+            // In case there are less data points than pixels in the chart,
+            // calculate nearest value.
+            const adjustedY = d.data[index];
+            const rounded = isNaN(adjustedY) ? 0 : (BIMath.round(Number(adjustedY) * 100)/BIMath.round(100));
```
PatrickTasse, line 451: BIMath returns bigint integers, so it loses all fractional part after the division. I think the intention was to keep two decimal places? You could use Number.toFixed(2), perhaps conditionally if the number has a fractional part (can be checked with % 1 != 0 perhaps ?).

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -426,24 +432,93 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
+        }
+        const chartWidth = this.lineChartRef.current.chartInstance.width;
+        const arraySize = this.state.xyData.labels.length;
+        const index = Math.max(Math.round((xPos / chartWidth) * (arraySize - 1)), 0);
+        const points: any = [];
+        let zeros = 0;
+        this.state.xyData.datasets.forEach((d: any) => {
+            // In case there are less data points than pixels in the chart,
+            // calculate nearest value.
+            const adjustedY = d.data[index];
```
PatrickTasse, line 450: Why adjusted? I think it is the raw y value?

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -426,24 +432,93 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
+                const point: any = {
+                    label: d.label,
+                    color: d.borderColor,
+                    background: d.backgroundColor,
+                    value: rounded.toString(),
+                };
+                points.push(point);
+            }
+            else {
+                zeros += 1;
```
PatrickTasse, line 462: It gives odd results, like if there are only two values and they are both zero, it shows only '2 others: 0'. Or if there are two values and one is zero, '1 other: 0'.

Also it doesn't protect against large tooltip if there are many non-zero values.

I would suggest the following algorithm: Compile all the points and sort them. Then only if there are more points than a certain threshold (e.g. 10+1) *** , show only the top 10, and show '(num - 10) others'.

We may also keep special handling for zeros, e.g.: with 69 data points, 32 non-zero and 37 zero:

<top 10>
22 others: non-zero
37 others: 0

*** threshold is 10+1 because if there are exactly 11 points, it's better to show them all than to show top 10 and '1 other' since it takes the same space anyways...

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -426,24 +432,93 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
+            }
+        });
+        // Sort in decreasing order
+        points.sort((a: any, b: any) => Number(b.value) - Number(a.value));
+
+        // Adjust tooltip position if mouse is close to the bottom of the window
+        let topPos = undefined;
+        let bottomPos = undefined;
+        if (y > window.innerHeight - 270) {
```
PatrickTasse, line 471: This only works if the tooltip height is guaranteed to be less than 270, is it the case?

Otherwise maybe safer to use innerheight / 2 as a condition? Although, that wastes half the space in every case. So it wouldn't work if the available height is less than twice the tooltip size. That is worst. never mind...

I don't know if there's a way to know the exact height of the tooltip before computing the position? Otherwise calculate an approximation based on the number of rows of text it will contain?

In any case, I would also suggest to invert logic so that if there is not enough height, topPos is used instead of bottomPos, as I think it's better to clip the bottom than to clip the top?

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -426,24 +432,93 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
+        else {
+            topPos = window.pageYOffset + y + 15;
+        }
+
+        const tooltipData = {
+            title: timeLabel,
+            dataPoints: points,
+            top: topPos,
+            bottom: bottomPos,
+            left: window.pageXOffset + x - 200,
```
PatrickTasse, line 483: Are we guaranteed that the tooltip width will not exceed 200? I'm worried that the series name is unconstrained?

PatrickTasse (changes requested):

packages/react-components/src/components/xy-output-component.tsx
PatrickTasse: Sorry, I didn't notice that you can have a scrollbar to show all non-zero data points. So I think it would be good enough to just have a threshold for a number of series where, below than number, we always include all data points regardless of the yValue.

packages/react-components/src/components/xy-output-component.tsx
Rodrigoplp-work: True. I used BIMath when TypeScript complained about assigning values at some point, but that requirement is no longer there when I added a new interface in tooltip-xy-component.
But I prefer to use `(Number(adjustedY) * 100) / 100` instead of `toFixed(2)` just because the former strips decimal numbers when they are 0 (it shows `100` , not `100.00`).

packages/react-components/src/components/xy-output-component.tsx
Rodrigoplp-work: It was "adjusted" because `index` is proportional to the ratio of of data per chart width.
Changed to "yValue".

packages/react-components/src/components/xy-output-component.tsx
Rodrigoplp-work: Good suggestion. Done.

packages/react-components/src/components/xy-output-component.tsx
Rodrigoplp-work: We have an upper limit to the size of the tooltip (`max-height = 250px` in tooltip-xy-component), so using 250 here is safe and more precise than inneheight / 2.

packages/react-components/src/components/xy-output-component.tsx
Rodrigoplp-work: Yes, the height of the tooltip is constrained to 250px, so this is safe.

packages/react-components/src/components/tooltip-xy-component.tsx
```diff
@@ -0,0 +1,123 @@
...
+                                margin: 'auto 0',
+                                border: 'solid thin',
+                                borderColor: point.color,
+                                backgroundColor: point.background
+                                }}
+                            ></div>
+                            <span style={{marginLeft: '5px'}}>{point.label} {point.value}</span>
+                        </li>
+                    )}
+                </ul>
```
PatrickTasse, line 95: I'm not sure what is the cause but odd behavior: Initially, the tooltip can go beyond the bounds of the xy-output-component (for example, over the right-side Outline widget). But after some window manipulation (for example, collapsing the left-side Trace Viewer widget) then the tooltip can no longer go beyond the xy-output-component bounds, it gets clipped to where the Outline widget starts.

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -426,24 +432,96 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
+        const arraySize = this.state.xyData.labels.length;
+        const index = Math.max(Math.round((xPos / chartWidth) * (arraySize - 1)), 0);
+        const points: any = [];
+        let zeros = 0;
+
+        this.state.xyData.datasets.forEach((d: any) => {
+            // In case there are less data points than pixels in the chart,
+            // calculate nearest value.
+            const yValue = d.data[index];
+            const rounded = isNaN(yValue) ? 0 : (Math.round(Number(yValue) * 100)/Math.round(100));
```
PatrickTasse, line 452: Space before and after /, and I think you don't need to round 100 ;)

PatrickTasse (changes requested):

packages/react-components/src/components/xy-output-component.tsx
Rodrigoplp-work: Right. The solution is to use "right" instead of "left". Now longer texts will make the tooltip grow to the left, where we have way more space.
Fixed.

packages/react-components/src/components/tooltip-xy-component.tsx
```diff
@@ -0,0 +1,123 @@
...
+                ref={this.divRef}
+                onMouseEnter={() => {
+                    this.setState({ inTooltip: true });
+                    if (this.timerId) {
+                        clearTimeout(this.timerId);
+                        this.timerId = undefined;
+                    }
+                }}
+                onMouseLeave={() => {
+                    this.setState({ tooltipData: { opacity: 0, transition: '0s' }, inTooltip: false });
```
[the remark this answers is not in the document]
Rodrigoplp-work: Using "right" instead of "left" to position the tooltip now. It will no longer go beyond the right bound of the xy-output-component.

packages/react-components/src/components/xy-output-component.tsx
Rodrigoplp-work: Indeed. Fixed.

packages/react-components/src/components/tooltip-xy-component.tsx
Rodrigoplp-work: Changed the "position" property of the tooltip from "fixed" to "absolute", which resolves the issue.

packages/react-components/src/components/xy-output-component.tsx
PatrickTasse: I can reproduce with the maximum number of items in the tooltip and just at the threshold of top/bottom condition a tooltip below the pointer that is clipped for last 1 or 2 text lines.

packages/react-components/src/components/tooltip-xy-component.tsx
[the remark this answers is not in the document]
PatrickTasse: Unfortunately that just moves the problem to the left bound :(

With a long trace name the tooltip can get clipped on the left bound.

I wonder if it's possible to have the tooltip to the left of the cursor but have its position snapped to the trace viewer left bound (if it exceeds)?

It seems it's very hard to have proper tooltip placement unless you know its exact rendered size...

PatrickTasse (changes requested):

packages/react-components/src/components/tooltip-xy-component.tsx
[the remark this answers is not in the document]
Rodrigoplp-work: Added a limit to the width of the tooltip. If text is longer, a scroll bar appears.
Also, if the mouse is close to the left border of the chart, the tooltip now appears on the right side of the mouse.
Now it is always possible to read the text, regardless of how long it is or the location of the tooltip.

packages/react-components/src/components/xy-output-component.tsx
Rodrigoplp-work: Increased the threshold. Now the tooltip is never clipped.

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -426,24 +432,108 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
+        }
+        else {
+            topPos = window.pageYOffset + y - 40;
+        }
+
+        // Adjust tooltip position if mouse is too close to the left edge of the chart
+        let leftPos = undefined;
+        let rightPos = undefined;
+        const xLocation = chartWidth - xPos + 40;
+        if (xLocation > chartWidth * 0.8) {
```
PatrickTasse, line 485: This condition doesn't seem to be foolproof, it seems it should not depend on width ratio. If the chart is not wide, the tooltip is truncated on the left bound.

Actually, if the tooltip width is larger than half the chart width, then neither option will prevent clipping.

It seems to me the solution has to be, that if we know (with toolitip size) that it will exceed the bounds then we should snap to that edge (leftBound=0 or rightBound=0).

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -426,24 +432,108 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
+            leftPos = x - 280;
```
PatrickTasse, line 486: If I understand correctly, leftPos should be a value relative to the left bound (0=left edge). But this value is using x which is display-relative coordinates. I see the tooltip very far to the right of the cursor.

PatrickTasse (changes requested):

packages/react-components/src/components/xy-output-component.tsx
Rodrigoplp-work: You are right. xPos is a better value here. It still depends on the width of the chart, but it will give a much more consistent result. Changed.

PatrickTasse (changes requested):

packages/react-components/src/components/xy-output-component.tsx
Rodrigoplp-work: The exact width of the tooltip is determined dynamically. Building a solution around calculating it before rendering might be tricky.
As the 80% threshold works well, I suggest leaving it like that for now. In case it proves not to be reliable, we attempt a different solution.

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -426,24 +432,108 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
+        else {
+            topPos = window.pageYOffset + y - 40;
+        }
+
+        // Adjust tooltip position if mouse is too close to the left edge of the chart
+        let leftPos = undefined;
+        let rightPos = undefined;
+        const xLocation = chartWidth - xPos + 40;
+        if (xLocation > chartWidth * 0.8) {
+            leftPos = xPos + 200;
```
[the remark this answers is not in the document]
Rodrigoplp-work: Good observation. I made the relative left position of the parent component of the chart available as a props, so that `leftPos` can now be inferred by taking into consideration all these variables you listed. It is indeed a much more precise solution.
Also replaced the arbitrary numbers with a private constant inside tooltip-xy-component.

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -467,24 +473,108 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
+
+        // Adjust tooltip position if mouse is too close to the left edge of the chart
+        let leftPos = undefined;
+        let rightPos = undefined;
+        const xLocation = chartWidth - xPos;
+        if (xLocation > chartWidth * 0.8) {
+            leftPos = x - this.props.style.componentLeft;
+        }
+        else {
+            rightPos = xLocation;
```
PatrickTasse, line 530: There's actually a 17 pixel right margin in the output component to the right edge of the chart. So the rightPos is 17 pixels closer to the mouse position than the leftPos would be. To be the same distance we would have to also consider the componentRight.

However, because of the width of the mouse cursor icon itself (~15 pixels) the end result is actually very nice, because the distance from the closest edge of the cursor icon to the tooltip is about the same on both sides!

So it's fine as long as that right margin does not change.

PatrickTasse (approved):

bhufmann (approved):
Great improvement. Thanks for the contribution.

Rodrigoplp-work (approved):

Issue #277, opened 2021-02-23
Rename Opened Traces widget and commands


PatrickTasse:
With the latest changes, the 'Opened Traces' widget in the Trace Viewer view will show the list of every trace that has ever been opened in the trace server (using 'Open With > Trace Viewer' menu in Explorer or 'Open Trace' dialog in Trace Viewer) and that has not been deleted. When a trace tab is closed, the trace remains opened in the trace-server and still appears in the 'Opened Traces' widget. But after a trace-server restart, the trace may be closed in the trace-server and it will still appear in the client.

Furthermore, each trace in the 'Opened Traces' widget has a context menu with three commands:
- Open Trace
- Close Trace
- Delete Trace

The Open Trace command opens the trace tab and if necessary, opens the trace in the trace-server
The Close Trace command closes the trace tab but has no effect on the trace-server
The Delete Trace command removes the trace from the trace-server workspace (links, supplementary files) but does not delete the actual trace files.

This issue is to consider whether the widget title 'Opened Traces' and the three command names are still appropriate and have an intuitive meaning for the users.

ebugden:
This is somewhat tangential, but **why not just have "Remove Trace / Delete Trace"** in the context menu? 
 * **"Close Trace"** (Hide Trace) is covered by closing the open tab (arguably this could be included in context menu to avoid searching for the tab to close).
 * **"Open Trace"** (View Trace) is covered by clicking on the trace on the list.

----

### Widget name thoughts
* **"Traces"** : My favourite. Simple and the ambiguity allows for flexibility without lying to the user (ex. Traces that aren't being viewed could be faded but still in the list).
* **"Opened traces"** : The traces in the list aren't necessarily opened.
* **"Available traces"** : There could be available traces that aren't on this list.
* **"Recently opened"** : Not necessarily. Just previously opened.

An option we can consider when renaming is white lying to the user. Well, misrepresenting reality in a way that makes their life less confusing, but isn't technically accurate. Ex. Using the same label for two different operations that feel similar for the user, but actually have different implementations (like label overloading instead of function overloading)

----

> **PatrickTasse:** 'Opened Traces' widget in the Trace Viewer view will show the list of every trace that has ever been opened in the trace server and that has not been deleted.

Listing all previously opened traces makes me nervous, but I'm not sure why. Could someone point me to where the rationale behind this decision was explained?

* Should there at least be an expiry on how long the traces sit there if they're not explicitly deleted?

[...]

PR #541, opened 2021-10-26
Add ability to remove trace sets from Trace Explorer

### LINKED ISSUE #336: Improve discoverability of removing traces from Opened Traces
For example, close box 'X' could appear on hover (right aligned). Currently, the "Remove" operation is in the right-click context menu which is less discoverable [#277 (comment)](https://github.com/theia-ide/theia-trace-extension/issues/277#issuecomment-802157186).

image

----
Issue source: @ssmagula's user tests + brainstorming with TC and tool team members

hriday-panchasara:
Fixes #336 
Remove button appears on hover
Screen Shot 2021-10-26 at 4 18 01 PM

When window is made smaller the remove button would still appear
Screen Shot 2021-10-26 at 4 18 29 PM

Clicking on remove button deletes the trace, does the same as right-click->'Delete Trace'

### DIFF
--- a/packages/react-components/src/trace-explorer/trace-explorer-opened-traces-widget.tsx
+++ b/packages/react-components/src/trace-explorer/trace-explorer-opened-traces-widget.tsx
@@ -8,6 +8,7 @@ import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import { faCopy } from '@fortawesome/free-solid-svg-icons';
 import { OpenedTracesUpdatedSignalPayload } from 'traceviewer-base/lib/signals/opened-traces-updated-signal-payload';
 import { ITspClientProvider } from 'traceviewer-base/lib/tsp-client-provider';
+import { faTimes } from '@fortawesome/free-solid-svg-icons';
 
 export interface ReactOpenTracesWidgetProps {
     id: string,
@@ -165,6 +166,13 @@ export class ReactOpenTracesWidget extends React.Component<ReactOpenTracesWidget
                     <h4 className='trace-element-name'>{traceName}</h4>
                     {this.renderTracesForExperiment(props.index)}
                 </div>
+                <div className='remove-trace-button-container' title='Remove traces from Trace Viewer'>
+                    <button data-tip data-for="removeTip" className='remove-trace-button'
+                        onClick={event => {this.handleOnExperimentDeleted(event,traceUUID);}}
+                    >
+                        <FontAwesomeIcon icon={faTimes} />
+                    </button>
+                </div>
                 {/* <div className='trace-element-options'>
                     <button className='share-context-button' onClick={this.handleShareButtonClick.bind(this, props.index)}>
                         <FontAwesomeIcon icon={faShareSquare} />
@@ -172,6 +180,13 @@ export class ReactOpenTracesWidget extends React.Component<ReactOpenTracesWidget
                 </div> */}
             </div>
         </div>;
+    }
+
+    protected doHandleOnExperimentDeleted(e: React.MouseEvent<HTMLButtonElement>, traceUUID: string): void {
+        this._experimentManager.closeExperiment(traceUUID);
+        signalManager().fireCloseTraceViewerTabSignal(traceUUID);
+        e.preventDefault();
+        e.stopPropagation();
     }
 
     protected renderTracesForExperiment(index: number): React.ReactNode {
@@ -258,6 +273,7 @@ export class ReactOpenTracesWidget extends React.Component<ReactOpenTracesWidget
     protected handleOnExperimentSelected = (e: React.MouseEvent<HTMLDivElement>): void => this.doHandleOnExperimentSelected(e);
     protected handleContextMenuEvent = (e: React.MouseEvent<HTMLDivElement>, traceUUID: string): void => this.doHandleContextMenuEvent(e, traceUUID);
     protected handleClickEvent = (e: React.MouseEvent<HTMLDivElement>, traceUUID: string): void => this.dohandleClickEvent(e, traceUUID);
+    protected handleOnExperimentDeleted = (e: React.MouseEvent<HTMLButtonElement>, traceUUID: string): void => this.doHandleOnExperimentDeleted(e, traceUUID);
 
     protected doHandleOnExperimentSelected(e: React.MouseEvent<HTMLDivElement>): void {
         const index = Number(e.currentTarget.getAttribute('data-id'));
--- a/packages/react-components/style/trace-explorer.css
+++ b/packages/react-components/style/trace-explorer.css
@@ -63,6 +63,44 @@
     grid-template-rows: auto;
 } */
 
+.remove-trace-button-container {
+    margin-right: 2px;
+    display: flex;
+    justify-content: center;
+    align-items: flex-start;
+}
+
+.remove-trace-button-container.theia-mod-selected {
+    background-color: var(--theia-selection-background);
+}
+
+.remove-trace-button-container :hover {
+    background-color: var(--theia-scrollbarSlider-background);
+    cursor: pointer; 
+}
+
+.remove-trace-button {
+    background: none;
+    border: none;
+    visibility: hidden;
+    margin-top: 2px;
+    color: var(--theia-ui-font-color0)
+}
+
+.remove-trace-button :hover {
+    background: none;
+}
+
+.trace-element-container {
+    display: flex;
+    flex-direction: row;
+    justify-content: space-between;
+}
+
+.trace-element-container:hover .remove-trace-button {
+    visibility: visible;
+}
+
 /* Remove focus outline on grid when selected */
 .ReactVirtualized__Grid {
     outline: none;


hriday-panchasara:
Tooltip added to remove button:
Screen Shot 2021-11-01 at 11 32 53 AM

bhufmann (dismissed):
LGTM. Thanks for the contribution.

ebugden (changes requested):
Requested change:
- [x] Button should be aligned with the trace group name (top right of the box). Otherwise when there are multiple files in the group it looks like you're deleting a particular file instead of all of them

image


----

🌈 Thank you for this improvement! These are great design choices:
- Having the button appear on hover keeps the widget visually clean and then draws attention to the button at a time the user might need it.
- @mirsky-work's suggestion to highlight the button when you hover directly over it clearly communicates that the user can interact with this button.

ebugden:
> > Wondering what tooltip should we add to clarify that the buttons deletes an experiment and not the trace files. Perhaps something like "Delete Trace Set"? @ebugden, any thoughts?
> 
> Do we need to have a warning dialog box asking the user to confirm it?

Personally, I'm comfortable merging without a confirmation dialogue and waiting for user feedback on this. 
* I don't think the average user (my guess of an average user: ~1 GB traces, 1-2 traces opened together) would be bothered that there's no confirmation 
* But I think it's probable that users dealing with massive traces and complex analyses would want a confirmation box (ex. Because they don't want to accidentally lose all the analysis configuration work they've done, or because it takes a long time to complete analyses again).

hriday-panchasara:
Remove button is aligned with trace group name:
Screen Shot 2021-11-03 at 10 30 41 AM

Added warning dialog box to confirm action:
remove-trace-button-gif

ssmagula:
Is it correct that clicking this close-box merely removes the traces from the UI of the viewer? The user could add them again, if it were removed by mistake? If that's correct, then maybe the correct verb is "Remove traces from Trace Viewer"? I'd lean away from confirming a non-destructive action, but it's a bit of a judgement call.

hriday-panchasara:
Screen Shot 2021-11-03 at 11 51 42 PM

bhufmann:
> Is it correct that clicking this close-box merely removes the traces from the UI of the viewer? The user could add them again, if it were removed by mistake? If that's correct, then maybe the correct verb is "Remove traces from Trace Viewer"? I'd lean away from confirming a non-destructive action, but it's a bit of a judgement call.

What it does it removes the traces from the trace viewer UI and from the trace server back-end. But, the actual traces files are not deleted from the local file system.

hriday-panchasara:
> Requested change:
> 
> * [ ]  Button should be aligned with the trace group name (top right of the box). Otherwise when there are multiple files in the group it looks like you're deleting a particular file instead of all of them
> 
> image
> 
> 🌈 Thank you for this improvement! These are great design choices:
> 
> * Having the button appear on hover keeps the widget visually clean and then draws attention to the button at a time the user might need it.
> * @mirsky-work's suggestion to highlight the button when you hover directly over it clearly communicates that the user can interact with this button.

I've implemented the requested changes as well as added a warning dialog box to confirm delete action

ebugden (changes requested):
Requested changes:
- [x] The tooltip label ("Delete trace from trace viewer") and confirmation messages ("Remove traces from trace viewer") should be the same (since they refer to the same action). 

Thanks for the update!

ebugden:
Something related to this PR is that the right-click context menu currently lists this same "Remove Trace from Trace Viewer" action as "Delete Trace". For consistency. This context menu should be changed to match the tooltip and confirmation "Remove Trace", but that has to do with #277 and would be a separate PR (I added a comment in that issue).

Context menu label for the same action as this PR: "Delete Trace"
image

----

> > Is it correct that clicking this close-box merely removes the traces from the UI of the viewer? The user could add them again, if it were removed by mistake? If that's correct, then maybe the correct verb is "Remove traces from Trace Viewer"? I'd lean away from confirming a non-destructive action, but it's a bit of a judgement call.
> 
> What it does it removes the traces from the trace viewer UI and from the trace server back-end. But, the actual traces files are not deleted from the local file system.

@ssmagula Removing the traces from the viewer is not entirely without consequence because you lose the analysis layout/configuration and also I believe all the analyses and indexing would need to be re-run which for large traces could be time consuming.

hriday-panchasara:
> Requested changes:
> 
> * [ ]  The tooltip label ("Delete trace from trace viewer") and confirmation messages ("Remove traces from trace viewer") should be the same (since they refer to the same action).
> 
> Thanks for the update!

Screen Shot 2021-11-10 at 12 45 08 PM

ebugden (approved):
Thanks for the update! Looks good to me 🚀

ebugden:
Before merging, please squash (combine) some of your commits. This PR doesn't need to be 8 commits.

ebugden (commented):
I'll let someone else take a look at the code, but the visual result and the commit looks good to me! Thanks for this improvement @hriday-panchasara ☀️

ssmagula:
Thanks, all for clarifying. I grasp the main issues, now. 

Short term: Let's change the labeling in the UI
from "Opened Traces" to » "Trace Sets"

This will enable the UI to use "Remove Trace Set" (on hover of close-box, and on right-click) to refer to the action of getting rid of this element ("trace group" seems equally clear, and we need to formalize our nomenclature, see below). The word "delete" implies "destruction that is non-recoverable." When a user gets rid of a trace set, the user will lose the indexing and any config, but this can be _redone_ so in my opionion, that's still a recoverable action. My gut is that confirming a recoverable action is not needed. But I'll defer to those who better know the pain of losing a giant trace set.

This discussion is a great example of why crystal clear nomenclature is _vital_. We (the product development team) use the word "Traces" to refer to both "a single trace data file" _and_ "a group or set of trace data files, which have already been indexed, and may already be configured by the user" 

**How might we even further clarify these nouns to the user?** If we are finding it hard to refer to these core nouns in a non-ambiguous way, how will users feel in the product? 
- Recommended: 
- A single trace file should have an icon and label that communicates "a single trace data file"
- A group of "trace data files" should similarly have a label and an icon that communicate "multiple trace data files" Maybe an acronym representing the trace format is also useful info to help disambiguate this set vs. that set?
- If people agree, I can write the user story for this

ebugden:
Thank you for weighing in on this topic @ssmagula!

> **ssmagula:** Short term: **Let's change the labeling in the UI from "Opened Traces" to » "Trace Sets"**
>
> This will enable the UI to use "Remove Trace Set" (on hover of close-box, and on right-click) to refer to the action of getting rid of this element ("trace group" seems equally clear, and we need to formalize our nomenclature, see below). 

I can see how changing the widget label from "Opened Traces" to "Trace Sets" would clarify some things, but this change should be proposed and debated in a separate issue and it is out of the scope of this PR in my opinion. **Can you create a separate issue for this and we can discuss there?**

In the mean time I changed this PR's title to more accurately reflect the changes that are proposed here. (We can use the language "trace set" in the commit log even if it's not yet in the UI because on the implementation side of things it's relevant to make the distinction between a "trace" and a "trace set"/"trace group" (formerly "experiment"))

----

> **ssmagula:** The word "delete" implies "destruction that is non-recoverable." When a user gets rid of a trace set, the user will lose the indexing and any config, but this can be _redone_ so in my opionion, that's still a recoverable action. My gut is that confirming a recoverable action is not needed. But I'll defer to those who better know the pain of losing a giant trace set.

This PR already lave this action as "Removing a trace". I agree that "delete" shouldn't be used to describe this action (unrelated to this PR: this needs to be changed in the right-click context menu, see my [comment](https://github.com/theia-ide/theia-trace-extension/issues/277#issuecomment-964266704) on #277)

----

> **ssmagula:** This discussion is a great example of why **crystal clear nomenclature is _vital_.** We (the product development team) use the word "Traces" to refer to both "a single trace data file" _and_ "a group or set of trace data files, which have already been indexed, and may already be configured by the user"
> 
> **How might we even further clarify these nouns to the user?** If we are finding it hard to refer to these core nouns in a non-ambiguous way, how will users feel in the product?
> 
> * Recommended:
> * A single trace file should have an icon and label that communicates "a single trace data file"
> * A group of "trace data files" should similarly have a label and an icon that communicate "multiple trace data files" Maybe an acronym representing the trace format is also useful info to help disambiguate this file vs. that file?
> * If people agree, I can write the user story for this

**I agree that we should have a discussion about nomenclature, but again in a separate issue. Can you make a separate issue?**

One of the reasons this nomenclature discussion isn't totally obvious is that some trace formats (ex. CTF) are already a collection of several trace files (although you're never really looking at them individually?).

mirsky-work:
Thanks everyone for the weighing in. I agree that _Remove_ is a better term than _Delete_, and the confirmation would probably be more annoying than useful, as it's a recoverable operation.

> * A single trace file should have an icon and label that communicates "a single trace data file"
> * A group of "trace data files" should similarly have a label and an icon that communicate "multiple trace data files" Maybe an acronym representing the trace format is also useful info to help disambiguate this file vs. that file?

@ssmagula, the distinction is already communicated today. Trace sets list the containing traces:
image

ebugden:
> **mirsky-work:** I agree that [...] the confirmation would probably be more annoying than useful, as it's a recoverable operation.

Ok let's remove it then since several folks seem to agree. Not having a confirmation was my first reflex. @hriday-panchasara could you make this additional change?
- [ ] Remove confirmation box

----

> > * A single trace file should have an icon and label that communicates "a single trace data file"
> > * A group of "trace data files" should similarly have a label and an icon that communicate "multiple trace data files" Maybe an acronym representing the trace format is also useful info to help disambiguate this file vs. that file?
> 
> @ssmagula, the distinction is already communicated today. Trace sets list the containing traces: image

I made an issue #561 to continue the discussion about this. Left some thoughts there.

bhufmann:
Just a quick note about the confirmation message. To avoid to show it all the time (and be annoying) there could be check-box `Don't show again` and the user's selection is then persisted. With this the users get informed about what happens and can choose to not see the warning again.

ebugden:
My preference would be to merge this without a confirmation box. Then, if users mention that they're frustrated because they're accidentally deleting complex analysis setups we can add a confirmation box (maybe that shows only when the traces are large i.e. have a large number of events)

> **bhufmann:** To avoid to show it all the time (and be annoying) there could be a check-box `Don't show again` and the user's selection is then persisted. With this the users get informed about what is going to happen and can choose to see or to not see the warning again.

I think having a "Don't show again" checkbox makes more sense in a situation that involves risk (ex. "Are you sure you want to automatically connect to this network in the future"), but it this situation (removing a trace set) we think that in most cases there isn't significant risk associated with it (because for most trace sets it takes maybe a minute max to redo the all the analyses the user had open).

hriday-panchasara:
> > **mirsky-work:** I agree that [...] the confirmation would probably be more annoying than useful, as it's a recoverable operation.
> 
> Ok let's remove it then since several folks seem to agree. Not having a confirmation was my first reflex. @hriday-panchasara could you make this additional change?
> 
> * [ ]  Remove confirmation box
> 
> > > * A single trace file should have an icon and label that communicates "a single trace data file"
> > > * A group of "trace data files" should similarly have a label and an icon that communicate "multiple trace data files" Maybe an acronym representing the trace format is also useful info to help disambiguate this file vs. that file?
> > 
> > 
> > @ssmagula, the distinction is already communicated today. Trace sets list the containing traces: image
> 
> I made an issue #561 to continue the discussion about this. Left some thoughts there.

I've removed the confirmation box

ebugden:
The behaviour looks good to me! Thanks for the update.

bhufmann (approved):
It looks good to me. Thanks!

bhufmann:
> My preference would be to merge this without a confirmation box. Then, if users mention that they're frustrated because they're accidentally deleting complex analysis setups we can add a confirmation box (maybe that shows only when the traces are large i.e. have a large number of events)
> 
> > **bhufmann:** To avoid to show it all the time (and be annoying) there could be a check-box `Don't show again` and the user's selection is then persisted. With this the users get informed about what is going to happen and can choose to see or to not see the warning again.
> 
> I think having a "Don't show again" checkbox makes more sense in a situation that involves risk (ex. "Are you sure you want to automatically connect to this network in the future"), but it this situation (removing a trace set) we think that in most cases there isn't significant risk associated with it (because for most trace sets it takes maybe a minute max to redo the all the analyses the user had open).

Removing a trace from the trace viewer will remove all analysis results and it can be a long procedure to rebuild (incl. upload the trace to the server in the future remote case). This depends on the size of the trace. I'll open a separate issue tracker for the confirmation dialog box because it's out of scope of this PR.

PatrickTasse (approved):
The button works but we will need to fix (later?) the inconsistency between the button tooltip and the equivalent item in the context menu. Also not sure about plural in "Remove traces" since the button applies to only one "Opened trace".

bhufmann:
I'll going to merge it as because it brings in an UX improvement. The inconsistent labels can be fixed as part of fix for #277.

PR #550, opened 2021-11-01
Add support for regular expression matching in Events table search

### LINKED ISSUE #410: Support Regular expression matching in Events Table search
Currently , the events table search only supports substring matching.

hriday-panchasara:
fixes #410

Example 1 - Search for substring (all matches for "sc"):
Screen Shot 2021-11-01 at 10 32 58 AM

Example 2 - Search for regular expression (strings that start with "sc"):
Screen Shot 2021-11-01 at 10 33 11 AM

### DIFF
--- a/packages/react-components/src/components/table-output-component.tsx
+++ b/packages/react-components/src/components/table-output-component.tsx
@@ -490,7 +490,7 @@ export class TableOutputComponent extends AbstractOutputComponent<TableOutputPro
             this.gridApi.forEachNode(rowNode => {
                 let isMatched = true;
                 this.filterModel.forEach((value, key) => {
-                    if (!rowNode.data[key].includes(value)) {
+                    if ((rowNode.data[key].search(new RegExp(value)) === -1)) {
                         isMatched = false;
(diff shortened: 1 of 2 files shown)


bhufmann (approved):

IbrahimFradj (changes requested):
I think I found a bug 
when I do a search in a column and go to the end of the line, I don't see the text I wrote in the search. 
the problem that this can cause is that if we don't find what we want with our search, we can't find where we changed it 

**To reproduce it see the GIF :**

tableSearch

bhufmann:
> This works well with regex expressions.
> 
> I want to make sure that we understand the impact on user: any string entered is a regex.
> 
> So the user has to make sure to escape any special regex characters: For example:
> 
> [0 is a regex pattern error that will match nothing, user has to enter \[0 0.5 in Timestamp column will not only match that second/tenth combination, it will match 075 in nanoseconds, user has to enter 0\.5
> 
> If that is fine and understood, I am OK.

Yes, that's fine.

PR #554, opened 2021-11-02
Move findTraces() to a backend service

PatrickTasse:

### DIFF
--- a/theia-extensions/viewer-prototype/src/browser/trace-viewer/trace-viewer.tsx
+++ b/theia-extensions/viewer-prototype/src/browser/trace-viewer/trace-viewer.tsx
@@ -1,5 +1,4 @@
 import { DisposableCollection, MessageService, Path } from '@theia/core';
-import { FileSystem, FileStat } from '@theia/filesystem/lib/common/filesystem';
 import { ApplicationShell, Message, StatusBar } from '@theia/core/lib/browser';
 import { ReactWidget } from '@theia/core/lib/browser/widgets/react-widget';
 import { inject, injectable, postConstruct } from 'inversify';
@@ -12,7 +11,6 @@ import { TraceManager } from 'traceviewer-base/lib/trace-manager';
 import { ExperimentManager } from 'traceviewer-base/lib/experiment-manager';
 import { TraceContextComponent } from 'traceviewer-react-components/lib/components/trace-context-component';
 import { Experiment } from 'tsp-typescript-client/lib/models/experiment';
-import URI from '@theia/core/lib/common/uri';
 import { TheiaMessageManager } from '../theia-message-manager';
 import { ThemeService } from '@theia/core/lib/browser/theming';
 import { signalManager, Signals } from 'traceviewer-base/lib/signals/signal-manager';
@@ -20,6 +18,8 @@ import { OutputAddedSignalPayload } from 'traceviewer-base/lib/signals/output-ad
 import { TraceExplorerWidget } from '../trace-explorer/trace-explorer-widget';
 import { TraceExplorerContribution } from '../trace-explorer/trace-explorer-contribution';
 import { MarkerSet } from 'tsp-typescript-client/lib/models/markerset';
+import { BackendFileService } from '../../common/backend-file-service';
+import { CancellationTokenSource } from '@theia/core';
 
 export const TraceViewerWidgetOptions = Symbol('TraceViewerWidgetOptions');
 export interface TraceViewerWidgetOptions {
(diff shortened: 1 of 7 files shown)


theia-extensions/viewer-prototype/src/common/backend-service.ts
```diff
@@ -0,0 +1,8 @@
+import { CancellationToken } from '@theia/core';
+
+export const backendServicePath = '/services/backend-service';
+export const BackendService = Symbol('BackendService');
+
+export interface BackendService {
```
bhufmann, line 6: Should the name be a bit clearer? It feels that name `BackendService` is too general

theia-extensions/viewer-prototype/src/node/backend-service-impl.ts
```diff
@@ -0,0 +1,49 @@
+import { injectable } from 'inversify';
+import fs = require('fs');
+import { Dirent } from 'fs';
+import { Path } from '@theia/core/lib/common/path';
+import { CancellationToken } from '@theia/core';
+import { BackendService } from '../common/backend-service';
+
+@injectable()
+export class BackendServiceImpl implements BackendService {
```
bhufmann, line 9: Same comment about name.

theia-extensions/viewer-prototype/src/browser/trace-viewer/trace-viewer.tsx
```diff
@@ -136,36 +136,22 @@ export class TraceViewerWidget extends ReactWidget {
...
-                    if (fileStat) {
-                        if (fileStat.isDirectory) {
-                            // Find recursivly CTF traces
-                            await this.findTraces(fileStat, tracesArray, isCancelled);
-                        } else {
-                            // Open single trace file
-                            tracesArray.push(this.uri);
-                        }
-                    }
+                    const tracesArray = await this.backendService.findTraces(this.uri.toString(), cancellation.token);
```
bhufmann, line 154: I pressed cancel while having a breakpoint in the backend service to test the cancellation. In the backend it stopped, but it returned an empty `tracesArray`. So, the check below is then true and I get the error message, `No valid traces found in selected directory ...`. When the array length is 0 and the cancellation was requested I think the error message should not be shown.

bhufmann (changes requested):
Thanks for the contribution

theia-extensions/viewer-prototype/src/common/backend-service.ts
```diff
@@ -0,0 +1,8 @@
+import { CancellationToken } from '@theia/core';
+
+export const backendServicePath = '/services/backend-service';
```
bhufmann, line 3: Can there be a name conflict with other extensions? Make it needs to be more specific.

theia-extensions/viewer-prototype/src/common/backend-service.ts
PatrickTasse: Changed to BackendFileService

theia-extensions/viewer-prototype/src/browser/trace-viewer/trace-viewer.tsx
PatrickTasse: Moved cancellation check before array size check.

theia-extensions/viewer-prototype/src/common/backend-service.ts
PatrickTasse: Changed prefix to /services/theia-trace-extension/<service name>, for the existing service also.

bhufmann (dismissed):
Looks good to me. Thanks!

theia-extensions/viewer-prototype/src/node/backend-file-service-impl.ts
```diff
@@ -0,0 +1,49 @@
...
+        } else if (stats.isFile()) {
+            traces.push(path);
+        }
+        if (cancellationToken.isCancellationRequested) {
+            return [];
+        }
+        return traces;
+    }
+
+    protected async findCtfTraces(path: string, traces: string[], cancellationToken: CancellationToken): Promise<void> {
```
MatthewKhouzam, line 25: Can we have a findDirTrace... uftrace won't work with this and findTraces.

MatthewKhouzam (changes requested):
I can remove the "request change" if you think uftrace is not a priority.

theia-extensions/viewer-prototype/src/node/backend-file-service-impl.ts
PatrickTasse: I renamed it to deepFindTraces() and added Javadoc to specify that it currently only finds CTF traces.

MatthewKhouzam (approved):
In the words of the kool aid man...
Oh Yeah!

bhufmann (approved):
Thanks!

PR #564, opened 2021-11-17
Ability to double-click states in Timegraph views

hriday-panchasara:
Support easier navigation in resources table by double-clicking a state to zoom into it
- Depends on https://github.com/theia-ide/timeline-chart/pull/169
- Width of element occupies the viewRange when double-clicked

Example:
Screen Recording 2021-11-17 at 3 43 02 PM

### DIFF
--- a/packages/react-components/src/components/timegraph-output-component.tsx
+++ b/packages/react-components/src/components/timegraph-output-component.tsx
@@ -133,6 +133,18 @@ export class TimegraphOutputComponent extends AbstractTreeOutputComponent<Timegr
             },
             mouseout: () => {
                 this.props.tooltipComponent?.setElement(undefined);
+            },
+            click: (el, ev, clickCount) => {
+                if (clickCount === 2) {
+                    const start = el.model.range.start;
+                    const end = el.model.range.end;
+                    if (start !== end) {
+                        this.props.unitController.viewRange = {
+                            start,
+                            end
+                        };
+                    }
+                }
             }
         });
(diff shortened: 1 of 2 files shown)


bhufmann (dismissed):
Looks good to me. Thanks for the update.

mirsky-work (approved):
Thank you for this useful contribution! LGTM.

packages/react-components/src/components/timegraph-output-component.tsx
```diff
@@ -133,6 +133,16 @@ export class TimegraphOutputComponent extends AbstractTreeOutputComponent<Timegr
             },
             mouseout: () => {
                 this.props.tooltipComponent?.setElement(undefined);
+            },
+            click: (el, ev, clickCount) => {
+                if (clickCount === 2) {
+                    const start = el.model.range.start;
+                    const end = el.model.range.end;
+                    this.props.unitController.viewRange = {
```
PatrickTasse, line 141: Some time chart elements have zero duration range (start == end) like annotations. We should not attempt to set the viewRange in that case. Although internally, the unit controller will not update the view range but it will still broadcast a viewRangeChanged event with the old view range.

PatrickTasse (changes requested):
The commit message is misleading, this is not only for Resources table (view?), this change adds the ability for all timegraph views.

PatrickTasse (approved):

bhufmann (approved):

PR #567, opened 2021-11-18
Add y-axis to XY charts

### LINKED ISSUE #421: XY Chart: Add y-axis

Rodrigoplp-work:
The y-axis is displayed on the left of the chart, not obstructing the
data, and with the same background color of the chart.

Charts that do not have a y-axis, like the Resources Status chart, have
that space occupied by the component that exists on the left, like the
list of processes.

This way, all charts and the top horizontal rule remain vertically aligned.

Also dynamically changes the numbers on the axis, so that 1,000 appears
as 1K, 50,000,000 as 50M and so on.

Fixes #421.

Screenshot 2021-11-18 at 16 13 54


MatthewKhouzam (dismissed):
I think we should move the scaler to a util soon, it's gonna be used alot.
image

ebugden:
Requested changes:

- [ ] Reference the fixed issue in the commit message (#421)
- [ ] Link the fixed issue in the PR

----

🌈 Thank you @Rodrigoplp-work! This elegant scale will be a _huge_ functional improvement! I messed around with the scale and it looks like it behaves properly, but not giving explicit approval because I didn't read the code. Design decisions I like:
* Using only tick marks and labels (rather than also having a vertical line): Satisfies the functional requirements (showing scale) and minimizes visual noise (distraction) 
* Using 'K' to show thousands instead of showing more digits: Again gives sufficient scale reference while minimizing visual distraction
  * Will it show 'M' for millions? 
* Label font and styling fits with the rest of the tool (so not distracting). 

Now that we have scale, the next step for this theme would be units!

Rodrigoplp-work:
Good catch, @IbrahimFradj. Added a bit more space for the axis, and reduced the font to 11px.
Also improved the commit and the PR message with @ebugden notes.
Thank you both.

packages/react-components/src/components/abstract-tree-output-component.tsx
```diff
@@ -4,14 +4,22 @@ import { ResponseStatus } from 'tsp-typescript-client/lib/models/response/respon
 
 export abstract class AbstractTreeOutputComponent<P extends AbstractOutputProps, S extends AbstractOutputState> extends AbstractOutputComponent<P, S> {
     renderMainArea(): React.ReactNode {
-        const treeWidth = this.props.widthWPBugWorkaround - this.getHandleWidth() - this.props.style.chartWidth;
+        // Make tree thiner when chart has a y-axis
+        const yAxisBuffer = this.props.outputDescriptor.type === 'TREE_TIME_XY' ? this.props.style.yAxisWidth: 0;
+        const treeWidth = this.props.widthWPBugWorkaround - this.getHandleWidth() - this.props.style.chartWidth - yAxisBuffer;
```
PatrickTasse, line 9: Wouldn't it be possible to make it so that only xy-output-component has a y-axis? (and any other component that has a y-axis).

Because now the other components need to implement renderYAxis() when it is not relevant to them. In the future, what possible other part(s) of other components will they need to 'not' have?

Perhaps this main area should be split in renderLeft/renderRight instead of renderTree/renderChart, where the default left is a tree, the default right is a chart, and xy-chart can override renderLeft to add a y-axis to its left part. Then all others don't need to know about it.

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -151,6 +156,39 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
+            .domain([0, Math.max(allMax, 1)])
+            .range([chartHeight - margin.bottom, margin.top]);
+        const yTransform = `translate(${margin.left}, 0)`;
+        // Abbreviate large numbers
+        const scaleYLabel = (d: number) => (
+            d >= 1000000000000 ? d / 1000000000000 + 'G' :
+            d >= 1000000000 ? d / 1000000000  + 'B':
+            d >= 1000000 ? d / 1000000 + 'M' :
+            d >= 1000 ? d / 1000 + 'K':
+            d
```
PatrickTasse, line 178: Should support small numbers also? milli micro nano?

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -151,6 +156,39 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
+        );
+        if (allMax > 0) {
+            select(this.yAxisRef.current).call(axisLeft(yScale).tickSizeOuter(0).ticks(4)).call(g => g.select('.domain').remove());
```
PatrickTasse, line 181: Below a certain height we might want to have only 2 ticks? Otherwise the tick labels overlap.

Also, we use ticks(4) when we want 5 ticks. Not cool. Not your fault :(

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -151,6 +156,39 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
+        const chartHeight = parseInt(this.props.style.height.toString());
+        const margin = { top: 15, right: 0, bottom: 5, left: this.props.style.yAxisWidth };
+        let allMax = 0;
+        this.state.xyData?.datasets?.forEach((dSet: any) => {
+            const rowMax = Math.max(...dSet.data);
+            allMax = Math.max(allMax, rowMax);
+        });
+        const yScale = scaleLinear()
+            .domain([0, Math.max(allMax, 1)])
+            .range([chartHeight - margin.bottom, margin.top]);
```
PatrickTasse, line 170: It appears that when all values are the same e.g. allMax, then Chart.js will draw a line at the middle height. So the yScale has to be different.

Also, when all values are non-zero, then max is at top and min is at bottom, but the yScale is [0, max]. It should be [min, max].

PatrickTasse (changes requested):

packages/react-components/src/components/xy-output-component.tsx
ebugden: True! I didn't think of this at all!

ebugden: Do we have a concrete example of when numbers would be very small? The latency scatter plots?

bhufmann: By the way,  the TSP currently provides a descriptions of the axis (x or y), with a data type, unit etc. With this, we can decide on how to format the axis (like it's done in Eclipse Trace Compass). However, the Trace Compass server doesn't serialize these parameters at the moment. 

So, the formatting of the y axis, including other case than the one currently provided in this PR, can be done in a separate task, that includes back-end and front-end updates. 

https://github.com/theia-ide/tsp-typescript-client/blob/master/tsp-typescript-client/src/models/xy.ts

packages/react-components/src/components/abstract-tree-output-component.tsx
Rodrigoplp-work: That still implicates changing the tree to be different, depending on whether there is a y-axis or not.
I tried a few alternatives, but it was hard to keep the vertical alignment of all charts.
I found the actual implementation was the easiest way to achieve the desired effect. As it satisfies the requirements of the ticket, I suggest it to be accepted.
We could open another where we can plan to experiment further options.

packages/react-components/src/components/trace-context-component.tsx
```diff
@@ -63,6 +63,7 @@ export class TraceContextComponent extends React.Component<TraceContextProps, Tr
     private readonly DEFAULT_COMPONENT_HEIGHT: number = 10;
     private readonly DEFAULT_COMPONENT_ROWHEIGHT: number = 20;
     private readonly SCROLLBAR_PADDING: number = 12;
+    private readonly Y_AXIS_WIDTH: number = 40;
```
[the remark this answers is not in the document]
Rodrigoplp-work: It is very hard to predict the exact width of the axis because D3 calculates the values based on the data. Sometimes it uses a decimal, in which case the point makes the width larger.
I run tests and 40 is a good compromise that will show the axis without taking excessive real state.

packages/react-components/src/components/xy-output-component.tsx
Rodrigoplp-work: In D3, the value of "ticks" is not guaranteed. The scale makes an effort to achieve that number of divisions, taking into consideration other limitations, like avoiding overlapping (see https://observablehq.com/@d3/d3-ticks).
It is possible to achieve an exact division of the scale with a custom implementation, but "ticks" provides a fairly good solution already.
In case you find we really need a custom solution, I suggest it to be done in a separate ticket.

packages/react-components/src/components/trace-context-component.tsx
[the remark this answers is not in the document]
PatrickTasse: OK, we will see later if it needs to be updated. I guess I am used to older, superior SWT technology ;) where the components automatically take their preferred size based on their content ;)

packages/react-components/src/components/xy-output-component.tsx
PatrickTasse: We can add support for small numbers later if necessary.

packages/react-components/src/components/xy-output-component.tsx
PatrickTasse: OK!

packages/react-components/src/components/xy-output-component.tsx
PatrickTasse: Works well now!

PatrickTasse (approved):

bhufmann (approved):
Great improvement... Thanks a lot!

MatthewKhouzam (approved):

ebugden:
Thank you @Rodrigoplp-work for this PR! In my opinion it is one of the most significant usability improvements this year!

PR #570, opened 2021-11-19
Fix duplicate tab when opening from file explorer that already is open

### LINKED ISSUE #569: Opening trace from file explorer opens duplicate tab
Opening trace from file explorer opens duplicate tab. Apparently the issue only exists after opening a trace from the "Explorer" (see attached video). Traces opened from the "Trace Viewer" don't cause problems.

https://user-images.githubusercontent.com/88502808/137372032-9d0d4319-66aa-4193-9c37-0c86375237c9.mp4

_Originally posted by @Rodrigoplp-work in https://github.com/theia-ide/theia-trace-extension/issues/521#issuecomment-943596734_

bhufmann:
Fixes #569

Opening from the file explorer a trace uses the file URI to open the
the trace viewer widget. Then it scans the file system for traces,
posts the traces and experiment to the trace server. Only at that
point the experiment UUID is known. Theia keeps a map from
with a key using URI+TraceWidgetOpenOptions to widget. When opening the
first time the open options are undefined and the key is only the URI.

Closing the trace that map will cleared. When opening the trace from
the trace explorer, the TraceWidgetOpenOption will be populated with
the experiment UUID, but the URI is not known anymore.

When the user now opens the same trace from same file location, the URI
is used but the there is no way to know the corresponding UUID until
the trace(s) and experiment is created on the server.

The fix provided checks uses this UUID to match the UUID of any widgets
that are already open. If an widget is already open it will close the
new widget since it's a duplicate.

The solution proposed will fix the issue, but a follow-up patch needs
to be worked that avoids opening the tab first and closing it when
a duplicate is detected.

### DIFF
--- a/theia-extensions/viewer-prototype/src/browser/trace-viewer/trace-viewer.tsx
+++ b/theia-extensions/viewer-prototype/src/browser/trace-viewer/trace-viewer.tsx
@@ -1,5 +1,5 @@
 import { DisposableCollection, MessageService, Path } from '@theia/core';
-import { ApplicationShell, Message, StatusBar } from '@theia/core/lib/browser';
+import { ApplicationShell, Message, StatusBar, WidgetManager } from '@theia/core/lib/browser';
 import { ReactWidget } from '@theia/core/lib/browser/widgets/react-widget';
 import { inject, injectable, postConstruct } from 'inversify';
 import * as React from 'react';
@@ -71,6 +71,7 @@ export class TraceViewerWidget extends ReactWidget {
     @inject(TheiaMessageManager) protected readonly _signalHandler: TheiaMessageManager;
     @inject(MessageService) protected readonly messageService: MessageService;
     @inject(TraceExplorerContribution) protected readonly traceExplorerContribution: TraceExplorerContribution;
+    @inject(WidgetManager) protected readonly widgetManager!: WidgetManager;
 
     @postConstruct()
     async init(): Promise<void> {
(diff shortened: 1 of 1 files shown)


theia-extensions/viewer-prototype/src/browser/trace-viewer/trace-viewer.tsx
```diff
@@ -204,14 +205,21 @@ export class TraceViewerWidget extends ReactWidget {
...
-                            this.id = experiment.UUID;
-
-                            if (this.isVisible) {
-                                signalManager().fireTraceViewerTabActivatedSignal(experiment);
+                            const widgets = this.widgetManager.getWidgets(TraceViewerWidget.ID);
+                            const widget = widgets.find(w => w.id === experiment.UUID);
+                            if (widget) {
+                                // Close widget if it had been opened previously.
+                                cancellation.cancel();
+                                this.dispose();
```
PatrickTasse, line 213: It should make sure that other widget is brought to front.

PatrickTasse (changes requested):

theia-extensions/viewer-prototype/src/browser/trace-viewer/trace-viewer.tsx
bhufmann: I'll update the patch to make sure it will.

PatrickTasse (approved):

Rodrigoplp-work (approved):

Issue #460, opened 2021-08-25
Inform user when analysis fails


ebugden:
Currently, if the analysis fails (i.e. treeResponse.status is FAILED) then the analysis's view panel never updates and shows "Analysis running" forever even though the server has responded. 

* **When the analysis status returned is FAILED, update the view panel to display a message (ex. "Analysis failed.")** 
* This needs to happen for all chart types (not just XY charts)
* If possible, add detail to the message to give the user some indication of why the analysis failed. Is the analysis broken? Was processing interrupted? The user's first reflex is to think that it's their fault the analysis failed (ex. Their trace files are the wrong format), so if it's not the user's fault this should be clearly communicated. 

https://user-images.githubusercontent.com/28311615/130871089-d9f0e8a3-05ac-40cb-a4ce-f2bf601a74ee.mp4

# To reproduce
This infinite spinning can be observed **the first time** the Event Matching Latency analysis is opened with the tracevizlab's apt trace.

* **Trace:** apt (from the tracevizlabs, can use this project's gitpod demo to open the apt trace)
* **Analysis:** Event Matching Latency

⚠️ **To re-observe, close the trace in the trace server** (right-click --> "Delete Trace" in the Opened Traces menu). **Event Matching Latency only returns status FAILED the first time** the analysis is opened on server (0:26 - 0:30), otherwise it returns the status COMPLETED (0:37 - 0:40).

https://user-images.githubusercontent.com/28311615/130796784-624f5c6d-48d8-4cb6-9a9b-4703d439c61f.mp4

# Implementation
This fix needs to be applied to all chart types (all output components). Ideally adding this message wouldn't require modifying each output component separately.
* PR #447 can be used as an example for which areas of the code to change. However this approach requires modifying each output component separately (code duplication --> more of a hassle to change) which isn't ideal. 
  * The CSS styling also seems to be messy in this PR (unnecessary contradiction/duplication with style tags) so that could be cleaned up as well.
   * Add the code in the [renderChart( ) function in the xy-output-component](https://github.com/theia-ide/theia-trace-extension/blob/f27170908b76bbbb1bfb6c37b3f84eb074a17886/packages/react-components/src/components/xy-output-component.tsx#L139).

## Pseudocode 
```typescript
renderChart {
   // ...
   if (this.state.outputStatus === ResponseStatus.FAILED) {
      // Show "analysis failed" message.
   }
   // ...
}
```

jeffGodonou:
[...]

PR #574, opened 2021-11-23
XY Charts, Timegraph, Table: Display message when analysis fails

ebugden:
# Contents
* **Commits 1-4:** Code and styling cleanup for "no results" message added by #447 (with my limited HTML knowledge)
* **Commits 5-7:** For the XY, Timegraph and Table output components, add message for when the analysis fails:
  * Output status is always set: When the TSP response is bad, and in all cases when the TSP response is good
  * "Analysis failed" message displays when the status is FAILED

Because there's a lot of cleanup and changes in several files I recommend looking at the commits separately.

# Main change

Currently, if the analysis fails (i.e. outputStatus is FAILED) then the
analysis's view panel never updates and shows "Analysis running" forever
even though the server has responded.

Show a message to tell the user the analysis has failed.

Contributes towards fixing #460

# Testing
The code was tested by setting the message to display when the analysis returned the status 'RUNNING' instead of when the analysis returned 'FAILED' and the message displayed correctly.

image

### DIFF
--- a/packages/react-components/src/components/timegraph-output-component.tsx
+++ b/packages/react-components/src/components/timegraph-output-component.tsx
@@ -186,11 +186,18 @@ export class TimegraphOutputComponent extends AbstractTreeOutputComponent<Timegr
                 this.setState({
                     outputStatus: treeResponse.status,
                     timegraphTree: treeResponse.model.entries,
-                    columns,
+                    columns
                 }, this.updateTotalHeight);
+            } else {
+                this.setState({
+                    outputStatus: treeResponse.status
+                });
             }
             return treeResponse.status;
         }
+        this.setState({
+            outputStatus: ResponseStatus.FAILED
+        });
         return ResponseStatus.FAILED;
     }
 
--- a/packages/react-components/src/components/xy-output-component.tsx
+++ b/packages/react-components/src/components/xy-output-component.tsx
@@ -122,6 +122,9 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
             }
             return treeResponse.status;
         }
+        this.setState({
+            outputStatus: ResponseStatus.FAILED
+        });
         return ResponseStatus.FAILED;
     }
 
--- a/packages/react-components/src/components/xy-output-component.tsx
+++ b/packages/react-components/src/components/xy-output-component.tsx
@@ -286,14 +286,14 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
     }
 
     renderChart(): React.ReactNode {
-        if (this.state.outputStatus === ResponseStatus.COMPLETED && this.state.xyTree.length === 0 ) {
+        // width={this.props.style.chartWidth}
+        if (this.state.outputStatus === ResponseStatus.COMPLETED && this.state.xyTree.length === 0) {
             return <React.Fragment>
-                {
-                    <p className='no-data' style={{fontSize: 20, marginRight: '5px', marginLeft: '5px', justifyContent:'center', alignItems:'center'}}>
-                        Trace analysis completed.
-                        No results: Trace missing required events.
-                    </p>
-                }
+                <div className='no-data'>
+                    Trace analysis completed.
+                    <br />
+                    No results: Trace missing required events.
+                </div>
             </React.Fragment>;
         }
         if (this.state.outputStatus === ResponseStatus.COMPLETED && this.state.xyData?.datasets?.length===0 ) {
--- a/packages/react-components/style/output-components-style.css
+++ b/packages/react-components/style/output-components-style.css
@@ -97,6 +97,14 @@ canvas {
     align-self: center;
 }
 
+.analysis-failed-message {
+    font-size: 20px;
+    margin-top: -20px; /* So text is vertically aligned relative to its center rather than its top */
+    transform: translate(0, 50%);
+    margin-left: auto;
+    margin-right: auto;
+}
+
 .table-tree {
     display: grid;
     border-collapse: collapse;
--- a/packages/react-components/src/components/abstract-output-component.tsx
+++ b/packages/react-components/src/components/abstract-output-component.tsx
@@ -109,4 +109,12 @@ export abstract class AbstractOutputComponent<P extends AbstractOutputProps, S e
     }
 
     abstract renderMainArea(): React.ReactNode;
+
+    protected analysisFailedMessage(): React.ReactFragment {
+        return <React.Fragment>
+            <div className='analysis-failed-message'>
+                Trace analysis failed.
+            </div>
+        </React.Fragment>;
+    }
 }
--- a/packages/react-components/src/components/abstract-tree-output-component.tsx
+++ b/packages/react-components/src/components/abstract-tree-output-component.tsx
@@ -4,6 +4,10 @@ import { ResponseStatus } from 'tsp-typescript-client/lib/models/response/respon
 
 export abstract class AbstractTreeOutputComponent<P extends AbstractOutputProps, S extends AbstractOutputState> extends AbstractOutputComponent<P, S> {
     renderMainArea(): React.ReactNode {
+        if (this.state.outputStatus === ResponseStatus.FAILED) {
+            return this.analysisFailedMessage();
+        }
+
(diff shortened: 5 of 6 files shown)


packages/react-components/style/output-components-style.css
```diff
@@ -83,8 +83,8 @@ canvas {
     align-self: center;
 }
 
-.no-data {
-    font-size: 24px;
+.important-message {
+    font-size: 20px;
```
bhufmann, line 87: Any reason for that change? I didn't find it too big before.

bhufmann (changes requested):
You could have a protected method in the AbstractTreeOutputComponent base class to create a fragment and it's called from all the output components to create the error fragment used to in the render method.

I noticed that the case where fetchTree() returns FAILED when called by the base class in waitAnalysisCompletion() is not handled by waitAnalysisCompletion(). This cause that the spinning wheel is shown forever. In this case also the error message should be displayed.

ebugden:
Required changes:

- [ ] In the AbstractTreeOutputComponent base class: Add protected method  to create a fragment
- [ ] In each output component: Call protected method from the render method to create the error fragment 
- [ ] Set state to FAILED when fetchTree() returns FAILED (last line of the method fetchTree()) so that the failed error message also displays in this case

Thank you for the guidance @bhufmann and for catching a bug in advance!

packages/react-components/style/output-components-style.css
ebugden: Good point. The font size change at line 87 should have been committed in this [earlier commit](https://github.com/theia-ide/theia-trace-extension/pull/574/commits/59d064484cb37a0c8e555d9fa44bd6bea44a8fa0) where I removed the hardcoded style tags to just use the CSS class. In theory there should be no change to how the the text is displayed before and after this PR because one of the style tags was setting the font size to 20. My understanding is that hardcoded style tags override CSS class.

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -226,7 +226,7 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
         if (this.state.outputStatus === ResponseStatus.COMPLETED && this.state.xyTree.length === 0 ) {
             return <React.Fragment>
                 {
-                    <p className='no-data' style={{fontSize: 20, marginRight: '5px', marginLeft: '5px', justifyContent:'center', alignItems:'center'}}>
```
ebugden, line 229: @bhufmann This is where the font size was originally set to 20. Now it's specified in the class instead. The change to the class should have been committed in this commit.

bhufmann:
>   - xy-output-component
>   - timegraph-output-component
>   -  table-output-component?

Yes, these are the ones.

bhufmann (changes requested):
> Set state to FAILED when fetchTree() returns FAILED (last line of the method fetchTree()) so that the failed error message also displays in this case

This bullet is not fixed yet. It should be simple by adding the following lines before returning FAILED:

```
        this.setState({
            outputStatus: ResponseStatus.FAILED
        });
```

ebugden:
> > Set state to FAILED when fetchTree() returns FAILED (last line of the method fetchTree()) so that the failed error message also displays in this case
> 
> This bullet is not fixed yet. It should be simple by adding the following lines before returning FAILED:
> 
> ```
>         this.setState({
>             outputStatus: ResponseStatus.FAILED
>         });
> ```

Yeah I'm committing my progress to not lose it, but it's still WIP. I'll let you know when It's ready to review!

ebugden:
## Updates
For the XY and Timegraph output components:
* Output status is always set: When the TSP response is bad, and in all cases when the TSP response is good
* "Analysis failed" message displays when the status is FAILED

## TODO
I'm still figuring out how to do this for the Table output component things since: 

- [ ]  Table output component inherits directly from AbstractOutputComponent rather than AbstractTreeOutputComponent like XY and Timegraph (I'll try adding the protected message method to AbstractOutputComponent instead)
- [ ] Table output component doesn't appear to store the Response Status right now (Need to add something to store it)

packages/react-components/src/components/xy-output-component.tsx
ebugden: Update: I ended up removing the size adjustment in the CSS class.

packages/react-components/style/output-components-style.css
```diff
@@ -93,6 +93,14 @@ canvas {
     align-self: center;
 }
 
+.analysis-failed-message {
+    font-size: 20px;
+    margin-top: -20px; /* So text is vertically aligned relative to its center rather than its top */
+    transform: translate(0, 50%);
+    margin-left: auto;
+    margin-right: auto;
+}
```
ebugden, line 102: CSS recommendations welcome! I have no knowledge of best practices when it comes to CSS.

ebugden:
## Updates
For the XY, Timegraph and Table output components:
* Output status is always set: When the TSP response is bad, and in all cases when the TSP response is good
* "Analysis failed" message displays when the status is FAILED

## TODO

- [ ] Update PR title and description

Rodrigoplp-work (changes requested):
Good work!
Could you join the "cleanup" commits into a single one? This would make your PR more concise and to the point, allowing an easier review and future reference.
Thanks for the contribution!

bhufmann (changes requested):

ebugden:
Updates: 
* Rebase on master
* Squash cleanup commits

Now the Gantt charts stay empty for some reason (they load as expected on master) so I need to look into that.

image

----

@bhufmann Thank you for the review! I haven't got the chance to look at your comments yet.

packages/react-components/src/components/timegraph-output-component.tsx
```diff
@@ -176,6 +176,9 @@ export class TimegraphOutputComponent extends AbstractTreeOutputComponent<Timegr
         const tspClientResponse = await this.props.tspClient.fetchTimeGraphTree(this.props.traceId, this.props.outputDescriptor.id, parameters);
         const treeResponse = tspClientResponse.getModel();
         if (tspClientResponse.isOk() && treeResponse) {
+            this.setState({
```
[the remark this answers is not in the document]
Rodrigoplp-work: Also, this is the reason why the charts are not being displayed.
Suggestion:
```js
if (tspClientResponse.isOk() && treeResponse) {
  if (treeResponse.model) {
    const headers = treeResponse.model.headers;
    const columns = [];
    if (headers && headers.length > 0) {
      headers.forEach(header => {
        columns.push({ title: header.name, sortable: true, tooltip: header.tooltip });
      });
    } else {
      columns.push({ title: 'Name', sortable: true });
    }
    this.setState({
      outputStatus: treeResponse.status,
      timegraphTree: treeResponse.model.entries,
      columns,
    }, this.updateTotalHeight);
  } else {
    this.setState({
      outputStatus: treeResponse.status
    });
  }
  return treeResponse.status;
}
this.setState({
  outputStatus: ResponseStatus.FAILED
});
return ResponseStatus.FAILED;
```

packages/react-components/src/components/timegraph-output-component.tsx
```diff
@@ -189,11 +189,18 @@ export class TimegraphOutputComponent extends AbstractTreeOutputComponent<Timegr
                 this.setState({
                     outputStatus: treeResponse.status,
                     timegraphTree: treeResponse.model.entries,
-                    columns,
+                    columns
```
ebugden, line 192: Removed this comma because I assumed it was a typo and things still worked when I did my usual tests.

ebugden:
Update:
* Applied review comments from @bhufmann and @Rodrigoplp-work. Thank you for your reviews!

Still need to rebase on master again and resolve conflicts.

packages/react-components/src/components/timegraph-output-component.tsx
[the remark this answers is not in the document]
ebugden: Thank you for flagging that this was what caused the charts to not load! It would have taken me a while to figure this out.

bhufmann (approved):
LGTM. Thanks for the updates.

Rodrigoplp-work (approved):
Well done! Thank you for the contribution.

PR #576, opened 2021-11-24
Unit tests events table

hriday-panchasara:
Created some unit tests for user-defined functions and components used in the events table. Was unable to create comprehensive tests for TableOutputComponent since its entirely built upon the AG-Grid library, and snapshot testing does not show the rendered HTML of AG-Grid. Another way to test TableOutputComponent could be to stub api calls made by the component to fetch table data and check if the internal state of the component (table columns & lines) updates. I was unable to mock tsp-client api calls or stub responses since jest could not reference the requests being made by the component's methods. But I believe we can later add to this test once we figure out a way to do so. Also included snapshot and unit tests for TableRendererComponents which are used by TableOutputComponent


hriday-panchasara:
New PR: https://github.com/theia-ide/theia-trace-extension/pull/596

PR #578, opened 2021-11-24
Provide front-end sorting of trees for based on the column

### LINKED ISSUE #577: Trees in XY or timeline charts chart are by strings by default
Some of the columns represent number values, however they are formatted on the server. For example, in the CPU usage view the
 column `%` shows `number + %` and the `Time` column shows formatted time values with unit, e.g. `100 ms`.

cpu-usage-sorting

Right now the cell content is provided as strings through the TSP and the formatting is done on the trace server. The TSP needs to be augmented to allow better sorting in either the front-end or the data is pre-sorted by the server based on a query parameter. This change is a full-stack change and starts with the TSP specification.

However, as stop-gap measure the front-end can try to parse the string and extract numbers for some know formats.

bhufmann:
Right now the sorting is done using string sorting. If the data in the
columns are numbers, then number sorting should be used. Moreover,
some known string formatting can be reversed to extract the number and
do a number comparison. For example, some data providers provide columns
with duration, where the value is formatted to show 'ns' for nanosecond,
'us' micro-second and so on. Other example, are percentages shown.

The tree of the CPU usage view when using the Trace Compass trace server
is good example for different columns and formatting.

Note that this implementation is a temporary solution until the TSP and
server back-end is augmented to provide a different mechanism to
support sorting.

fixes #577

### DIFF
--- a/packages/react-components/src/components/utils/filtrer-tree/sort.tsx
+++ b/packages/react-components/src/components/utils/filtrer-tree/sort.tsx
@@ -35,31 +35,31 @@ export const sortNodes = (nodes: TreeNode[], sortConfig: SortConfig[]): TreeNode
     if (orderToSort) {
         sortedNodes.sort((node1: TreeNode, node2: TreeNode) => {
             const index = orderToSort.columnIndex;
-            const order = (orderToSort.sortState === sortState.asc) ? 'asc' : 'desc';
-            const value1 = node1.labels[index];
-            const value2 = node2.labels[index];
-            let result = 0;
+            const value1: string = node1.labels[index];
+            const value2: string = node2.labels[index];
+            let comp = 0;
             if (!value1 && value2) {
-                result = -1;
+                comp = -1;
             } else if (value1 && !value2) {
-                result = 1;
+                comp = 1;
             } else if (!value1 && !value2) {
-                result = 0;
+                comp = 0;
             } else {
-                if (typeof value1 === 'string' && typeof value2 === 'string') {
-                    const comp = (value1 as string).localeCompare(value2);
-                    result = (order === 'asc') ? -comp : comp;
-                } else {
-                    if (value1 < value2) {
-                        result = (order === 'asc') ? -1 : 1;
-                    } else if (value1 > value2) {
-                        result = (order === 'asc') ? 1 : -1;
+                const number1 = parseValue(value1);
+                const number2 = parseValue(value2);
+                if (number1 !== undefined && number2 !== undefined) {
+                    if (number1 < number2) {
+                        comp = -1;
+                    } else if (number1 > number2) {
+                        comp = 1;
                     } else {
-                        result = 0;
+                        comp = 0;
                     }
+                } else {
+                    comp = value1.localeCompare(value2);
                 }
             }
-            return result;
+            return (orderToSort.sortState === sortState.asc) ? comp : -comp;
         });
         sortedNodes.forEach((node: TreeNode) => {
             if (node.children.length) {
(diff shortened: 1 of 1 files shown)


packages/react-components/src/components/utils/filtrer-tree/sort.tsx
```diff
@@ -47,13 +47,26 @@ export const sortNodes = (nodes: TreeNode[], sortConfig: SortConfig[]): TreeNode
                 result = 0;
             } else {
                 if (typeof value1 === 'string' && typeof value2 === 'string') {
```
PatrickTasse, line 49: I think value1 and value2 are strings by definition and already checked for null/undefined, so is this if-condition unnecessary?

packages/react-components/src/components/utils/filtrer-tree/sort.tsx
```diff
@@ -47,13 +47,26 @@ export const sortNodes = (nodes: TreeNode[], sortConfig: SortConfig[]): TreeNode
...
+                            result = (order === 'asc') ? -1 : 1;
+                        } else {
+                            result = 0;
+                        }
+                    } else {
+                        comp = (value1 as string).localeCompare(value2);
+                        result = (order === 'asc') ? -comp : comp;
+                    }
                 } else {
                     if (value1 < value2) {
```
PatrickTasse, line 66: ... and this code unreachable?

packages/react-components/src/components/utils/filtrer-tree/sort.tsx
```diff
@@ -69,3 +82,33 @@ export const sortNodes = (nodes: TreeNode[], sortConfig: SortConfig[]): TreeNode
...
+    let floatNumber = NaN;
+    let factor = -1;
+    const valueArray = valueString.split(' ');
+    if (valueArray.length === 1) {
+        floatNumber = Number.parseFloat(valueString);
+        factor = 1;
+    } else if (valueArray.length === 2) {
+        const value = valueArray[0];
+        const unit = valueArray[1];
+        if (unit === 'ns' || unit === '%') {
```
PatrickTasse, line 96: It's probably OK as this is only a temporary fix, but it seems that for comparing a unit-less number with a unit-suffixed number, the factors should be 1E-9 / 1E-6 / 1E-3 / 0.01 for ns / us / ms / %.

PatrickTasse (changes requested):

packages/react-components/src/components/utils/filtrer-tree/sort.tsx
bhufmann: you're right ... value1 and value2 are strings.

packages/react-components/src/components/utils/filtrer-tree/sort.tsx
bhufmann: you're right ... value1 and value2 are strings.

packages/react-components/src/components/utils/filtrer-tree/sort.tsx
bhufmann: not necessary... if it's unit less then we don't know what the value represent. They could be ms. So, mixing different units will lead incorrect sorting. This should be fixed once a fullstack solution is implemented.

PatrickTasse (commented):
From the UI everything is fine. But internally, the sort method doesn't respect the API, but it is inversed in a double-negative way that makes it work:

- In ascending order, val1 < val2 returns 1.
- The default sort order is 'desc' and it expects elements in ascending order

I let you decide if this should be fixed or not...

bhufmann:
> From the UI everything is fine. But internally, the sort method doesn't respect the API, but it is inversed in a double-negative way that makes it work:
> 
>     * In ascending order, val1 < val2 returns 1.
> 
>     * The default sort order is 'desc' and it expects elements in ascending order
> 
> 
> I let you decide if this should be fixed or not...

I updated it. Hopefully it's fine now.

PatrickTasse (approved):

MatthewKhouzam (approved):
Verks great!

PR #580, opened 2021-11-29
Remove unused Legend column of the tree in the xy-output-component

### LINKED ISSUE #579: Empty legend column in XY Charts
Each xy-chart has a legend column as last column in the tree which is currently not populated. This was done to mimic the way it was done in Eclipse Trace Compass. 

image

The column should be removed from the view because it doesn't contain any data.

bhufmann:
fixes #579


PatrickTasse (dismissed):

PatrickTasse (approved):

ebugden:
Could this be merged? My understanding is that if the submitter is a committer we just need one more approval.

I also gave it a quick try and it looks like the code behaves as intended.

MatthewKhouzam (approved):

PR #581, opened 2021-11-30
Issue/479 specify server cli arguments

### LINKED ISSUE #479: Add a setting in the UI to specify trace server CLI arguments
The default workspace location is in the user's home folder (~/.tracecompass-webapp). That can be an issue with NFS mounted home folders that can fill-up quickly, when opening very large traces, making the system slow and unusable. There's a way to override the default workspace path, by passing it with `-data` to the trace server. There should be a UI setting added to make it possible to change it from the UI (under the existing Trace Viewer section).

hriday-panchasara:
fixes #479 

As discussed, we're changing the settings field to contain just _Trace Server Path_ and _Tracer Server Arguments_ (set by default to '-vmargs -Dtraceserver.port=8080').

Previous implementation was to have a _Trace Server Port_ argument. This PR assumes that valid arguments are provided and makes use of regEx to extract the port from the arguments string. What validation can we put to check if the provided arguments are valid?

Screen Shot 2021-11-30 at 11 14 06 AM

### DIFF
--- a/theia-extensions/viewer-prototype/src/browser/trace-server-preference.ts
+++ b/theia-extensions/viewer-prototype/src/browser/trace-server-preference.ts
@@ -1,8 +1,9 @@
 import { PreferenceSchema, PreferenceProxy, PreferenceScope } from '@theia/core/lib/browser';
-import { TRACE_SERVER_DEFAULT_PORT } from '../common/trace-server-url-provider';
+import { TRACE_VIEWER_DEFAULT_PORT } from '../common/trace-server-url-provider';
 
-export const TRACE_PATH = 'traceViewer.traceServer.path';
-export const TRACE_PORT = 'traceViewer.traceServer.port';
+export const TRACE_PATH = 'trace Viewer.trace Server.path';
+export const TRACE_PORT = 'trace Viewer.trace Viewer.port';
+export const TRACE_ARGS = 'trace Viewer.trace Server.arguments';
 
 export const ServerSchema: PreferenceSchema = {
     scope: PreferenceScope.Folder,
@@ -14,8 +15,14 @@ export const ServerSchema: PreferenceSchema = {
         },
         [TRACE_PORT]: {
             type: 'integer',
-            default: TRACE_SERVER_DEFAULT_PORT,
-            description: 'Specify the port on which you want to execute the server. This change will take effect the next time you open a trace',
+            default: TRACE_VIEWER_DEFAULT_PORT,
+            description: 'Specify the port the Trace Viewer would use to connect to the trace server',
+        },
+        [TRACE_ARGS]: {
+            type: 'string',
+            default: '',
+            description: 'Specify trace-server command line arguments. This change will take effect the next time the trace server starts.'+'\n'+
+            'E.g. for Trace Compass server: -data /home/user/server-workspace -vmargs -Dtraceserver.port=8080',
         }
     },
 };
@@ -23,6 +30,7 @@ export const ServerSchema: PreferenceSchema = {
 interface TracePreferenceContribution {
     [TRACE_PATH]?: string
     [TRACE_PORT]: number
+    [TRACE_ARGS]: string
 }
 
 export const TracePreferences = Symbol('TracePreferences');
--- a/theia-extensions/viewer-prototype/src/browser/trace-viewer/trace-viewer-contribution.ts
+++ b/theia-extensions/viewer-prototype/src/browser/trace-viewer/trace-viewer-contribution.ts
@@ -7,7 +7,7 @@ import { FileDialogService, OpenFileDialogProps } from '@theia/filesystem/lib/br
 import { WorkspaceService } from '@theia/workspace/lib/browser/workspace-service';
 import { OpenTraceCommand, StartServerCommand, StopServerCommand, TraceViewerCommand } from './trace-viewer-commands';
 import { PortBusy, TraceServerConfigService } from '../../common/trace-server-config';
-import { TracePreferences, TRACE_PATH, TRACE_PORT } from '../trace-server-preference';
+import { TracePreferences, TRACE_PATH, TRACE_ARGS } from '../trace-server-preference';
 import { TspClient } from 'tsp-typescript-client/lib/protocol/tsp-client';
 import { TspClientProvider } from '../tsp-client-provider-impl';
 
@@ -41,8 +41,8 @@ export class TraceViewerContribution extends WidgetOpenHandler<TraceViewerWidget
         return this.tracePreferences[TRACE_PATH];
     }
 
-    protected get port(): number | undefined {
-        return this.tracePreferences[TRACE_PORT];
+    protected get args(): string | undefined {
+        return this.tracePreferences[TRACE_ARGS];
     }
 
     protected createWidgetOptions(uri: URI, options?: TraceViewerWidgetOpenerOptions): TraceViewerWidgetOptions {
@@ -61,18 +61,27 @@ export class TraceViewerContribution extends WidgetOpenHandler<TraceViewerWidget
         } catch (outer) {
             const progress = await this.messageService.showProgress({ text: '' });
             progress.report({ message: 'Launching trace server... ', work: { done: 10, total: 100 } });
-            const { path, port } = this;
+            const { path, args } = this;
             try {
-                const resolve = await this.traceServerConfigService.startTraceServer({ path, port });
+                const resolve = await this.traceServerConfigService.startTraceServer({ path, args });
                 if (resolve === 'success') {
-                    progress.report({ message: `Trace server started on port: ${port}.`, work: { done: 100, total: 100 } });
+                    if (this.args && this.args.length > 0) {
+                        progress.report({ message: `Trace server started using the following arguments:  ${this.args}.`, work: { done: 100, total: 100 } });
+                    } else {
+                        progress.report({ message: 'Trace server started.', work: { done: 100, total: 100 } });
+                    }
                     progress.cancel();
                     this.openDialog();
                 }
             } catch (inner) {
                 // eslint-disable-next-line @typescript-eslint/no-explicit-any
                 if (PortBusy.is(inner as any)) {
-                    this.messageService.error(`Error opening serial port ${port}. (Port busy)`);
+                    if (this.args && this.args.length > 0) {
+                        this.messageService.error(
+                            `Error starting the server (port busy) using the following arguments: ${this.args}`);
+                    } else {
+                        this.messageService.error('Error starting the server (port busy)');
+                    }
                 } else {
                     this.messageService.error(
                         'Failed to start the trace server: no such file or directory. Please make sure that the path is correct in Trace Viewer settings and retry'
@@ -106,18 +115,27 @@ export class TraceViewerContribution extends WidgetOpenHandler<TraceViewerWidget
         } catch (outer) {
             return this.messageService.showProgress({ text: '' }).then(async progress => {
                 progress.report({ message: 'Launching trace server... ', work: { done: 10, total: 100 } });
-                const { path, port } = this;
+                const { path, args } = this;
                 try {
-                    const resolve = await this.traceServerConfigService.startTraceServer({ path, port });
+                    const resolve = await this.traceServerConfigService.startTraceServer({ path, args });
                     if (resolve === 'success') {
                         await this.waitForTraceServer(10_000);
-                        progress.report({ message: `Trace server started on port: ${port}.`, work: { done: 100, total: 100 } });
+                        if (this.args && this.args.length > 0) {
+                            progress.report({ message: `Trace server started using the following arguments:  ${args}.`, work: { done: 100, total: 100 } });
+                        } else {
+                            progress.report({ message: 'Trace server started.', work: { done: 100, total: 100 } });
+                        }
                         return super.open(traceURI, options);
                     }
                 } catch (inner) {
                     // eslint-disable-next-line @typescript-eslint/no-explicit-any
                     if (PortBusy.is(inner as any)) {
-                        this.messageService.error(`Error opening serial port ${port}. (Port busy)`);
+                        if (this.args && this.args.length > 0) {
+                            this.messageService.error(
+                                `Error starting the server (port busy) using the following arguments: ${this.args}`);
+                        } else {
+                            this.messageService.error('Error starting the server (port busy)');
+                        }
                     } else {
                         this.messageService.error(
                             'Failed to start the trace server: no such file or directory. Please make sure that the path is correct in Trace Viewer settings and retry'
@@ -142,14 +160,23 @@ export class TraceViewerContribution extends WidgetOpenHandler<TraceViewerWidget
         });
         registry.registerCommand(StartServerCommand, {
             execute: async () => {
-                const { path, port } = this;
+                const { path, args } = this;
                 try {
-                    await this.traceServerConfigService.startTraceServer({ path, port });
-                    this.messageService.info(`Trace server started successfully on port: ${port}.`);
+                    await this.traceServerConfigService.startTraceServer({ path, args });
+                    if (this.args && this.args.length > 0) {
+                        this.messageService.info(`Trace server started using the following arguments:  ${this.args}.`);
+                    } else {
+                        this.messageService.info('Trace server started.');
+                    }
                 } catch (error) {
                     // eslint-disable-next-line @typescript-eslint/no-explicit-any
                     if (PortBusy.is(error as any)) {
-                        this.messageService.error(`Error opening serial port ${port}. (Port busy)`);
+                        if (this.args && this.args.length > 0) {
+                            this.messageService.error(
+                                `Error starting the server (port busy) using the following arguments: ${this.args}`);
+                        } else {
+                            this.messageService.error('Error starting the server (port busy)');
+                        }
                     } else {
                         console.error(error);
                         this.messageService.error(
@@ -164,9 +191,9 @@ export class TraceViewerContribution extends WidgetOpenHandler<TraceViewerWidget
             execute: async () => {
                 try {
                     await this.traceServerConfigService.stopTraceServer();
-                    this.messageService.info(`Trace server terminated successfully on port: ${this.port}.`);
+                    this.messageService.info('Trace server terminated successfully.');
                 } catch (err) {
-                    this.messageService.error(`Failed to stop the trace server on port: ${this.port}.`);
+                    this.messageService.error('Failed to stop the trace server.');
                 }
             }
         });
--- a/theia-extensions/viewer-prototype/src/common/trace-server-config.ts
+++ b/theia-extensions/viewer-prototype/src/common/trace-server-config.ts
@@ -8,7 +8,7 @@ export const PortBusy = ApplicationError.declare(-32650, code => ({
 
 export interface StartTraceServerOptions {
     path?: string
-    port?: number
+    args?: string
 }
 
 export const TraceServerConfigService = Symbol('TraceServerConfigService');
--- a/theia-extensions/viewer-prototype/src/common/trace-server-url-provider.ts
+++ b/theia-extensions/viewer-prototype/src/common/trace-server-url-provider.ts
@@ -1,5 +1,5 @@
 export const TRACE_SERVER_DEFAULT_URL = 'http://localhost:{}/tsp/api';
-export const TRACE_SERVER_DEFAULT_PORT = 8080;
+export const TRACE_VIEWER_DEFAULT_PORT = 8080;
 
 export const TraceServerUrlProvider = Symbol('TraceServerUrlProvider');
 export interface TraceServerUrlProvider {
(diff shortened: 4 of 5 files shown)


theia-extensions/viewer-prototype/src/browser/trace-explorer/trace-explorer-sub-widgets/trace-explorer-placeholder-widget.tsx
```diff
@@ -95,18 +95,22 @@ export class TraceExplorerPlaceholderWidget extends ReactWidget {
             }).then(async progress => {
                 progress.report({ message: 'Launching trace server... ', work: { done: 10, total: 100 } });
                 try {
-                    const resolve = await this.traceServerConfigService.startTraceServer(this.path, this.port);
+                    const resolve = await this.traceServerConfigService.startTraceServer(this.path, this.args);
                     if (resolve === 'success') {
-                        progress.report({ message: `Trace server started on port: ${this.port}.`, work: { done: 100, total: 100 } });
+                        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+                        const port = this.args!.match(/(?<=port=)\d+/g)![0];
```
mirsky-work, line 101: I wonder if we can get rid of using the port at all, since it still makes an assumption about the structure of the CLI args. Instead we can output something like:
"Trace server started using the following arguments: ..."

If we still decide to keep the port, can we move that Regex to something like PreferenceService, to avoid repeating it in other locations?

@bhufmann , any thoughts?

bhufmann: > @bhufmann , any thoughts?

I'll get back to you later on today.

bhufmann: Here I think we shouldn't mention the port anymore. I agree with the suggestion of @mirsky-work.

theia-extensions/viewer-prototype/src/browser/trace-server-preference.ts
```diff
@@ -1,8 +1,8 @@
 import { PreferenceSchema, PreferenceProxy, PreferenceScope } from '@theia/core/lib/browser';
-import { TRACE_SERVER_DEFAULT_PORT } from '../common/trace-server-url-provider';
+import { TRACE_SERVER_DEFAULT_ARGS } from '../common/trace-server-url-provider';
 
-export const TRACE_PATH = 'trace-viewer.path';
-export const TRACE_PORT = 'trace-viewer.port';
+export const TRACE_PATH = 'trace Viewer.trace Server.path';
+export const TRACE_ARGS = 'trace Viewer.trace Server.arguments';
```
bhufmann, line 5: The port preference should not be removed because it's needed in the URL of the REST API.

theia-extensions/viewer-prototype/src/browser/trace-viewer/trace-viewer-contribution.ts
```diff
@@ -123,13 +127,17 @@ export class TraceViewerContribution extends WidgetOpenHandler<TraceViewerWidget
         registry.registerCommand(StartServerCommand, {
             execute: async () => {
                 try {
-                    await this.traceServerConfigService.startTraceServer(this.path, this.port);
-                    this.messageService.info(`Trace server started successfully on port: ${this.port}.`);
+                    await this.traceServerConfigService.startTraceServer(this.path, this.cliArgs);
+                    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+                    const port = this.cliArgs!.match(/(?<=port=)\d+/g)![0];
+                    this.messageService.info(`Trace server started successfully on port: ${port}.`);
```
bhufmann, line 133: The message should not mention the port. Either just mention `Trace Server started successfully` or add the whole command-line executed: `Trace Server started successfully: ./tracecompass-server -vmargs -Dtraceserver.port=8080`.

Make it consistent with the above with toast message in the other file (TraceExplorerPlaceholderWidget) above.

bhufmann (changes requested):
Thanks for the update

bhufmann:
A heads-up that #484 has been merged and you'll need to rebase your change on top of it. It touches the spawning of the server.

hriday-panchasara:
> A heads-up that #484 has been merged and you'll need to rebase your change on top of it. It touches the spawning of the server.

Here's the updated preferences:
Screen Shot 2021-12-06 at 12 43 17 PM

theia-extensions/viewer-prototype/src/browser/trace-server-preference.ts
mirsky-work: Having the port specified in 2 fields is confusing and error prone. If Theia supported referencing preferences [1], perhaps in Arguments we could specify something like: `-vmargs -Dtraceserver.port=${config.trace Viewer.trace Viewer.port}`. But alas, looks like it's not implemented in Theia [2]. I guess we'll have to live with these 2 preferences for now, but at least we should modify the description of "Trace Server: Port" to clarify that it's purpose has changed and now it's used to specify what port to use by the client.

@paul-marechal, can you please confirm that  https://github.com/eclipse-theia/theia/issues/8805 isn't indeed implemented?

[1] https://code.visualstudio.com/docs/editor/variables-reference#_configuration-variables 
[2] https://github.com/eclipse-theia/theia/issues/8805

paul-marechal: Variable substitution in settings is a delicate matter, here's what VS Code does:
https://code.visualstudio.com/docs/editor/variables-reference#_is-variable-substitution-supported-in-user-and-workspace-settings

We might support as much as VS Code, but generically we can't use variables, let alone refer to other settings.

You can always manually substitute string patterns at use time in your logic though.

paul-marechal:
> This PR assumes that valid arguments are provided and makes use of regEx to extract the port from the arguments string

Isn't there a paradox here? We assume that valid arguments are provided, but we still check the port using RegExps?

What happens when we swap to a different Trace Server with a CLI that doesn't have the same port argument format?

I'd leave the CLI validation to the spawned program and get a way to display initialization errors coming from the child?

theia-extensions/viewer-prototype/src/browser/trace-server-preference.ts
bhufmann: yeah, I understand it seems confusing and error prone to specify the port at 2 different places. However, to make the way a server is launched independent of the actual server implementation, I think there is no other way. The front-end needs to be told which port to use and hence there is a separate preference.

Alternatively, we could standardize the the command-line which each trace server implementation needs to follow. It would be an API that needs to be supported (and probably versioned). For this solution the port is an obvious CLI parameter. The server workspace location is something else and might Trace Compass server specific.

Maybe we could inspired by how language servers are spawn, configured (port, cache for indexing etc).

theia-extensions/viewer-prototype/src/node/trace-server-service.ts
```diff
@@ -11,25 +11,33 @@ export class TraceServerServiceImpl implements TraceServerConfigService {
...
-    async startTraceServer(path?: string, port?: number): Promise<string> {
+    async startTraceServer(path?: string, cliArgs?: string): Promise<string> {
         if (path === undefined) {
             throw new Error('no Trace Server path specified');
-        } else if (port === undefined) {
+        } else if (cliArgs === undefined) {
+            throw new Error('no Trace Server cli arguments specified');
+        }
+
+        const port = cliArgs.match(/(?<=port=)\d+/g);
```
bhufmann, line 21: The port doesn't need to be parsed because it's Trace Compass server specific and this patch aims the server to be independent of the server implementation specifics.

theia-extensions/viewer-prototype/src/node/trace-server-service.ts
```diff
@@ -11,25 +11,33 @@ export class TraceServerServiceImpl implements TraceServerConfigService {
     private server?: ChildProcess;
     private port?: number;
 
-    async startTraceServer(path?: string, port?: number): Promise<string> {
+    async startTraceServer(path?: string, cliArgs?: string): Promise<string> {
         if (path === undefined) {
             throw new Error('no Trace Server path specified');
-        } else if (port === undefined) {
+        } else if (cliArgs === undefined) {
```
bhufmann, line 17: The cli arguments should be optional. For example, it should be possible to start the Trace Compass server without any arguments. This will start the TC server with default port and workspace.

theia-extensions/viewer-prototype/src/node/trace-server-service.ts
```diff
@@ -11,25 +11,33 @@ export class TraceServerServiceImpl implements TraceServerConfigService {
...
+
+        const argsArray: Array<string> | undefined = cliArgs?.split(' ');
+        const server = spawn(path, argsArray);
         if (server.pid === undefined) {
             // When pid is undefined it usually means we're about to get an error.
             return new Promise<never>((_, reject) => server.once('error', reject));
         }
         this.server = server;
-        this.port = port;
+        this.port = parseInt(port[0]);
```
bhufmann, line 40: instead of saving the port you should save the cliArgs and replace the port with it in the file (line 23, line 54, line 63).

bhufmann (changes requested):

theia-extensions/viewer-prototype/src/node/trace-server-service.ts
paul-marechal: Something I learned recently with JSON-RPC services is that if the method parameters are positional (just like they are now) then `undefined` values are transformed into `null` values.

Two possible fixes:
- Just write `if (!path) {` so that either `undefined` or `null` work the same way.
- Replace the positional arguments with an `options` object containing the potentially undefined fields as those will properly carry the `undefined` values...

I actually changed to the latter in a different PR. This is a limitation of JSON-RPC :/

hriday-panchasara: Replace the positional arguments with an options object as done in PR #594

bhufmann (changes requested):
I think we should have a README.md about configuring the trace server. This is not in scope of this PR.

theia-extensions/viewer-prototype/src/browser/trace-explorer/trace-explorer-sub-widgets/trace-explorer-placeholder-widget.tsx
```diff
@@ -94,19 +94,22 @@ export class TraceExplorerPlaceholderWidget extends ReactWidget {
...
                     }
                 }
                 catch (err) {
                     // eslint-disable-next-line @typescript-eslint/no-explicit-any
                     if (PortBusy.is(err as any)) {
+                        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+                        const port = this.args!.match(/(?<=port=)\d+/g)![0];
                         this.messageService.error(
-                            `Error opening serial port ${this.port}. (Port busy)`);
+                            `Error opening serial port ${port}. (Port busy)`);
```
mirsky-work, line 112: Extracting the port parameter is server-specific. The message can just output the whole args string:
`Error starting the server (port busy) using the following arguments: ${this.args}`

theia-extensions/viewer-prototype/src/browser/trace-server-preference.ts
```diff
@@ -15,15 +16,21 @@ export const ServerSchema: PreferenceSchema = {
         },
         [TRACE_PORT]: {
             type: 'number',
-            default: TRACE_SERVER_DEFAULT_PORT,
-            description: 'Specify the port on which you want to execute the server. This change will take effect the next time you open a trace',
+            default: TRACE_VIEWER_DEFAULT_PORT,
+            description: 'Specify the port the view is using to connect to the trace server',
```
mirsky-work, line 20: I suggest:
`Specify the port the Trace Viewer would use to connect to the trace server'

mirsky-work (changes requested):

theia-extensions/viewer-prototype/src/browser/trace-explorer/trace-explorer-sub-widgets/trace-explorer-placeholder-widget.tsx
```diff
@@ -40,16 +40,16 @@ export class TraceExplorerPlaceholderWidget extends ReactWidget {
...
         this.preferenceService.onPreferenceChanged(event => {
-            if (event.preferenceName === TRACE_PORT) {
-                this.port = event.newValue;
+            if (event.preferenceName === TRACE_ARGS) {
+                this.args = event.newValue;
             }
             if (event.preferenceName === TRACE_PATH) {
                 this.path = event.newValue;
             }
         });
```
mirsky-work, line 49: Port change events should still be handled, no?

hriday-panchasara: We replaced the port variable with args in this file, hence Port change events wouldn't affect any implementation

theia-extensions/viewer-prototype/src/node/trace-server-service.ts
```diff
@@ -1,35 +1,38 @@
...
 
-    async startTraceServer(path?: string, port?: number): Promise<string> {
-        if (path === undefined) {
+    async startTraceServer({ path, args }: StartTraceServerOptions = {}): Promise<string> {
+        if (!path) {
             throw new Error('no Trace Server path specified');
-        } else if (port === undefined) {
-            throw new Error('no Trace Server port specified');
+        } else if (!args) {
+            args = TRACE_SERVER_DEFAULT_ARGS;
```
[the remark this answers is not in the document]
bhufmann: Mhhh, maybe, we can provide some default arguments or some help string in the package.json

mirsky-work (approved):
LGTM. The args are now somewhat decoupled from the source code.

theia-extensions/viewer-prototype/src/browser/trace-explorer/trace-explorer-sub-widgets/trace-explorer-placeholder-widget.tsx
```diff
@@ -94,10 +94,11 @@ export class TraceExplorerPlaceholderWidget extends ReactWidget {
                 text: ''
             }).then(async progress => {
                 progress.report({ message: 'Launching trace server... ', work: { done: 10, total: 100 } });
+                const { path, args } = this;
                 try {
-                    const resolve = await this.traceServerConfigService.startTraceServer(this.path, this.port);
+                    const resolve = await this.traceServerConfigService.startTraceServer({ path, args });
                     if (resolve === 'success') {
-                        progress.report({ message: `Trace server started on port: ${this.port}.`, work: { done: 100, total: 100 } });
+                        progress.report({ message: `Trace server started using the following arguments:  ${this.args}.`, work: { done: 100, total: 100 } });
```
bhufmann, line 101: If the ${this.args} is empty then the message is: `Trace server started using the following arguments: .`

If `this.args` is empty the message should be clearer.

theia-extensions/viewer-prototype/src/browser/trace-explorer/trace-explorer-sub-widgets/trace-explorer-placeholder-widget.tsx
```diff
@@ -106,7 +107,7 @@ export class TraceExplorerPlaceholderWidget extends ReactWidget {
                     // eslint-disable-next-line @typescript-eslint/no-explicit-any
                     if (PortBusy.is(err as any)) {
                         this.messageService.error(
-                            `Error opening serial port ${this.port}. (Port busy)`);
+                            `Error starting the server (port busy) using the following arguments: ${this.args}`);
```
bhufmann, line 110: same comment her about about the empty string `this.args`

theia-extensions/viewer-prototype/src/browser/trace-server-preference.ts
```diff
@@ -15,15 +18,21 @@ export const ServerSchema: PreferenceSchema = {
...
         [TRACE_PORT]: {
             type: 'number',
-            default: TRACE_SERVER_DEFAULT_PORT,
-            description: 'Specify the port on which you want to execute the server. This change will take effect the next time you open a trace',
+            default: TRACE_VIEWER_DEFAULT_PORT,
+            description: 'Specify the port the Trace Viewer would use to connect to the trace server',
+        },
+        [TRACE_ARGS]: {
+            type: 'string',
+            default: traceServerDefaultArgs,
```
bhufmann, line 26: the default argument should be an empty string.

theia-extensions/viewer-prototype/src/browser/trace-server-preference.ts
```diff
@@ -1,8 +1,11 @@
 import { PreferenceSchema, PreferenceProxy, PreferenceScope } from '@theia/core/lib/browser';
-import { TRACE_SERVER_DEFAULT_PORT } from '../common/trace-server-url-provider';
+import { TRACE_VIEWER_DEFAULT_PORT } from '../common/trace-server-url-provider';
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const { traceServerDefaultArgs } = require('../../package.json');
```
bhufmann, line 4: I'm not for providing it in the package.json of the theia-trace-extension. It practically the same as having it as constant in the source code. 

I suggested to have the default arguments an empty string and it can be part of the source code.

theia-extensions/viewer-prototype/src/browser/trace-viewer/trace-viewer-contribution.ts
```diff
@@ -98,7 +98,7 @@ export class TraceViewerContribution extends WidgetOpenHandler<TraceViewerWidget
                     // eslint-disable-next-line @typescript-eslint/no-explicit-any
                     if (PortBusy.is(err as any)) {
                         this.messageService.error(
-                            `Error opening serial port ${port}. (Port busy)`);
+                            `Error starting the server (port busy) using the following arguments: ${this.args}`);
```
bhufmann, line 101: same comment her about about the empty string `this.args`

theia-extensions/viewer-prototype/src/browser/trace-viewer/trace-viewer-contribution.ts
```diff
@@ -122,14 +122,15 @@ export class TraceViewerContribution extends WidgetOpenHandler<TraceViewerWidget
 
         registry.registerCommand(StartServerCommand, {
             execute: async () => {
+                const { path, args } = this;
                 try {
-                    await this.traceServerConfigService.startTraceServer(this.path, this.port);
-                    this.messageService.info(`Trace server started successfully on port: ${this.port}.`);
+                    await this.traceServerConfigService.startTraceServer({ path, args });
+                    this.messageService.info(`Trace server started using the following arguments:  ${this.args}.`);
```
bhufmann, line 128: same comment her about about the empty string `this.args`

theia-extensions/viewer-prototype/src/browser/trace-viewer/trace-viewer-contribution.ts
```diff
@@ -122,14 +122,15 @@ export class TraceViewerContribution extends WidgetOpenHandler<TraceViewerWidget
...
                 } catch (err) {
                     // eslint-disable-next-line @typescript-eslint/no-explicit-any
                     if (PortBusy.is(err as any)) {
                         this.messageService.error(
-                            `Error opening serial port ${this.port}. (Port busy)`);
+                            `Error starting the server (port busy) using the following arguments: ${this.args}`);
```
bhufmann, line 133: same comment her about about the empty string `this.args`

theia-extensions/viewer-prototype/package.json
```diff
@@ -44,5 +44,6 @@
       "frontend": "lib/browser/trace-viewer/trace-viewer-frontend-module",
       "backend": "lib/node/trace-server-backend-module"
     }
-  ]
+  ],
+  "traceServerDefaultArgs": "-vmargs -Dtraceserver.port=8080"
```
bhufmann, line 48: Remove (see explanation below)

bhufmann (changes requested):
The default command-line arguments should be an empty string. Providing trace compass specific arguments in the package.json or in the source code is pretty much the same.

theia-extensions/viewer-prototype/package.json
hriday-panchasara: > Remove (see explanation below)

Set default command line arguments to an empty string, and resolved merge conflicts

hriday-panchasara:
@bhufmann Default cli args provided through Preference Description
Screen Shot 2022-01-18 at 5 02 49 PM

bhufmann (changes requested):
One small update and then it's fine from my side.

bhufmann (approved):
Thanks a lot for this contribution. With this the theia-trace-extension will work with other servers than the Trace Compass server.

PatrickTasse (approved):

PR #582, opened 2021-11-30
Rename opened traces context menu delete command

### LINKED ISSUE #277: Rename Opened Traces widget and commands

hriday-panchasara:
fixes #277

'Delete trace' renamed to 'Remove trace' to make it consistent with the remove traces tooltip

Screen Shot 2021-11-30 at 12 38 34 PM

### DIFF
--- a/theia-extensions/viewer-prototype/src/browser/trace-explorer/trace-explorer-commands.ts
+++ b/theia-extensions/viewer-prototype/src/browser/trace-explorer/trace-explorer-commands.ts
@@ -13,8 +13,8 @@ export namespace TraceExplorerCommands {
         label: 'Close Trace'
     };
 
-    export const DELETE_TRACE: Command = {
-        id: 'trace-explorer:delete-trace',
-        label: 'Delete Trace'
+    export const REMOVE_TRACE: Command = {
+        id: 'trace-explorer:remove-trace',
+        label: 'Remove Trace'
     };
 }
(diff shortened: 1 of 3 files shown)


theia-extensions/viewer-prototype/src/browser/trace-explorer/trace-explorer-commands.ts
```diff
@@ -15,6 +15,6 @@ export namespace TraceExplorerCommands {
 
     export const DELETE_TRACE: Command = {
```
bhufmann, line 16: DELETE_TRACE should be changed too

theia-extensions/viewer-prototype/src/browser/trace-explorer/trace-explorer-commands.ts
```diff
@@ -15,6 +15,6 @@ export namespace TraceExplorerCommands {
...
         id: 'trace-explorer:delete-trace',
```
bhufmann, line 17: delete-trace too

bhufmann (changes requested):
Thanks for the contribution.

The tooltip of the delete button says "Remove Traces ...". I think the tooltip should say "Remove Trace ..." instead. Could you please also update that in this PR? I think it fits well in its scope.

bhufmann (approved):
Thanks a lot. Looks good to me.

PatrickTasse (approved):

PR #591, opened 2021-12-03
Add intuitive way to open additional traces

### LINKED ISSUE #193: Add intuitive way to open additional traces
Currently, there is no intuitive way to open additional traces if a trace is already open. Without this, users could become frustrated early in their user experience or assume that you can only open one trace at a time. 

Initially, this use case was going to be covered by adding an "Open Trace" option to the File menu #159, but the VS Code extensions API does not allow contributions to the main menu: 
* VS Code menu contribution points [documentation](https://code.visualstudio.com/api/references/contribution-points#contributes.menus)
* [Comment](https://github.com/theia-ide/theia-trace-extension/issues/159#issuecomment-733749083) on Add “Open Trace” option to File menu #159 

The other currently planned options for opening a trace do not cover this use case:
* The "Open Trace" button #160 just covers the case where no traces are open.
* The current alternatives (Right-click in the file explorer, CTRL+SHIFT+P + "Open Trace") would not be the user's first reflex.

## Proposed solution
Keep the "Open Trace" button (from PR #186) in the side panel menu even after opening the first trace (ex. under the opened traces)

image

hriday-panchasara:
fixes #193 

- Button added to the 'Opened Traces' widget header to allow opening additional traces. Button appears when mouse is hovered over widget
- In my approach I'm emitting a new signal when button is clicked, but I can reuse one of the existing signals instead

Named button tooltip "Open New Trace" because I think it differentiates from the context menu "Open Trace" command:
Screen Shot 2021-12-03 at 2 46 48 PM

Screen Shot 2021-12-03 at 2 49 16 PM

Unable to add '+' to the icon because theia commands accept a single iconClass param and I couldn't find a way to stack multiple icons in one element. We can upgrade font-awesome to v5 to get access to a folder-plus icon

### DIFF
--- a/theia-extensions/viewer-prototype/src/browser/trace-explorer/trace-explorer-sub-widgets/trace-explorer-placeholder-widget.tsx
+++ b/theia-extensions/viewer-prototype/src/browser/trace-explorer/trace-explorer-sub-widgets/trace-explorer-placeholder-widget.tsx
@@ -3,14 +3,8 @@ import { ReactWidget } from '@theia/core/lib/browser';
 import * as React from 'react';
 import { CommandService } from '@theia/core';
 import { OpenTraceCommand } from '../../trace-viewer/trace-viewer-commands';
-import { PortBusy, TraceServerConfigService } from '../../../common/trace-server-config';
 import { PreferenceService } from '@theia/core/lib/browser';
 import { TRACE_PATH, TRACE_PORT } from '../../trace-server-preference';
-import { TspClient } from 'tsp-typescript-client/lib/protocol/tsp-client';
-import { TspClientProvider } from '../../tsp-client-provider-impl';
-import { TspClientResponse } from 'tsp-typescript-client/lib/protocol/tsp-client-response';
-import { HealthStatus } from 'tsp-typescript-client/lib/models/health';
-import { MessageService } from '@theia/core';
 
 @injectable()
 export class TraceExplorerPlaceholderWidget extends ReactWidget {
@@ -18,24 +12,17 @@ export class TraceExplorerPlaceholderWidget extends ReactWidget {
     static LABEL = 'Trace Exploerer Placeholder Widget';
     protected path: string | undefined;
     protected port: number | undefined;
-    private tspClient: TspClient;
 
     state = {
         loading: false
     };
 
-    private constructor(
-        @inject(TspClientProvider) private tspClientProvider: TspClientProvider
-    ) {
+    private constructor() {
         super();
-        this.tspClient = this.tspClientProvider.getTspClient();
-        this.tspClientProvider.addTspClientChangeListener(tspClient => this.tspClient = tspClient);
     }
 
     @inject(CommandService) protected readonly commandService!: CommandService;
     @inject(PreferenceService) protected readonly preferenceService: PreferenceService;
-    @inject(TraceServerConfigService) protected readonly traceServerConfigService: TraceServerConfigService;
-    @inject(MessageService) protected readonly messageService: MessageService;
 
     @postConstruct()
     init(): void {
@@ -79,44 +66,10 @@ export class TraceExplorerPlaceholderWidget extends ReactWidget {
     protected handleOpenTrace = async (): Promise<void> => this.doHandleOpenTrace();
 
     private async doHandleOpenTrace() {
-        try {
-            this.state.loading = true;
-            this.update();
-            const healthResponse = await this.tspClient.checkHealth();
-            if ((healthResponse as TspClientResponse<HealthStatus>).getModel()?.status === 'UP') {
-                this.state.loading = false;
-                this.update();
-                this.commandService.executeCommand(OpenTraceCommand.id);
-            }
-        }
-        catch (e) {
-            this.messageService.showProgress({
-                text: ''
-            }).then(async progress => {
-                progress.report({ message: 'Launching trace server... ', work: { done: 10, total: 100 } });
-                try {
-                    const resolve = await this.traceServerConfigService.startTraceServer(this.path, this.port);
-                    if (resolve === 'success') {
-                        progress.report({ message: `Trace server started on port: ${this.port}.`, work: { done: 100, total: 100 } });
-                        progress.cancel();
-                        this.commandService.executeCommand(OpenTraceCommand.id);
-                    }
-                }
-                catch (err) {
-                    // eslint-disable-next-line @typescript-eslint/no-explicit-any
-                    if (PortBusy.is(err as any)) {
-                        this.messageService.error(
-                            `Error opening serial port ${this.port}. (Port busy)`);
-                    } else {
-                        this.messageService.error(
-                            'Failed to start the trace server: no such file or directory. Please make sure that the path is correct in Trace Viewer settings and retry');
-                    }
-                    progress.cancel();
-                }
-                this.state.loading = false;
-                this.update();
-            });
-        }
+        this.state.loading = true;
+        this.update();
+        await this.commandService.executeCommand(OpenTraceCommand.id);
+        this.state.loading = false;
+        this.update();
     }
-
 }
--- a/theia-extensions/viewer-prototype/src/browser/trace-viewer/trace-viewer-contribution.ts
+++ b/theia-extensions/viewer-prototype/src/browser/trace-viewer/trace-viewer-contribution.ts
@@ -57,6 +57,40 @@ export class TraceViewerContribution extends WidgetOpenHandler<TraceViewerWidget
         };
     }
 
+    protected async launchTraceServer(): Promise<void> {
+        try {
+            const healthResponse = await this.tspClient.checkHealth();
+            if ((healthResponse as TspClientResponse<HealthStatus>).getModel()?.status === 'UP') {
+                this.openDialog();
+            }
+        } catch (e) {
+            this.messageService.showProgress({
+                text: ''
+            }).then(async progress => {
+                progress.report({ message: 'Launching trace server... ', work: { done: 10, total: 100 } });
+                try {
+                    const resolve = await this.traceServerConfigService.startTraceServer(this.path, this.port);
+                    if (resolve === 'success') {
+                        progress.report({ message: `Trace server started on port: ${this.port}.`, work: { done: 100, total: 100 } });
+                        progress.cancel();
+                        this.openDialog();
+                    }
+                }
+                catch (err) {
+                    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+                    if (PortBusy.is(err as any)) {
+                        this.messageService.error(
+                            `Error opening serial port ${this.port}. (Port busy)`);
+                    } else {
+                        this.messageService.error(
+                            'Failed to start the trace server: no such file or directory. Please make sure that the path is correct in Trace Viewer settings and retry');
+                    }
+                    progress.cancel();
+                }
+            });
+        }
+    }
+
     public async openDialog(): Promise<void> {
         const props: OpenFileDialogProps = {
             title: 'Open Trace',
@@ -113,7 +147,9 @@ export class TraceViewerContribution extends WidgetOpenHandler<TraceViewerWidget
 
     registerCommands(registry: CommandRegistry): void {
         registry.registerCommand(OpenTraceCommand, {
-            execute: () => { this.openDialog(); }
+            execute: async () => {
+                await this.launchTraceServer();
+            }
         });
 
         registry.registerCommand(TraceViewerCommand, {
--- a/theia-extensions/viewer-prototype/src/browser/trace-viewer/trace-viewer-toolbar-commands.ts
+++ b/theia-extensions/viewer-prototype/src/browser/trace-viewer/trace-viewer-toolbar-commands.ts
@@ -31,6 +31,11 @@ export namespace TraceViewerToolbarCommands {
         label: 'Markers',
         iconClass: 'fa fa-bars fa-lg',
     };
+    export const OPEN_TRACE: Command = {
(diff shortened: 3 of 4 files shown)


ebugden (changes requested):
Requested changes:

- [x] Change new button tooltip to "Open Trace"
- [ ] Suggestion: Put code related to trace opening in a different file? Rather than having it in the placeholder widget

----

> Button added to the 'Opened Traces' widget header to allow opening additional traces. Button appears when mouse is hovered over widget

* The button interactions match the File Explorer icons perfectly! Good consistency.
* I like this backup icon choice. If we can't have a file with a plus, I think having an open file will make it more clear that this button is for opening traces.

Thank you for these changes @hriday-panchasara! This is one of the most noticeable UX improvements this year! ☀️

mirsky-work (approved):
Functionality works well! Thanks for this useful contribution!

theia-extensions/viewer-prototype/src/browser/trace-explorer/trace-explorer-sub-widgets/trace-explorer-placeholder-widget.tsx
```diff
@@ -54,6 +55,12 @@ export class TraceExplorerPlaceholderWidget extends ReactWidget {
         this.id = TraceExplorerPlaceholderWidget.ID;
         this.title.label = TraceExplorerPlaceholderWidget.LABEL;
         this.update();
+        signalManager().on(Signals.OPEN_NEW_TRACE, this.handleOpenTrace);
```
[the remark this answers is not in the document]
hriday-panchasara: @PatrickTasse @bhufmann any suggestions? I catch the signal in the placeholder widget because the button would be in that component and the widget already has a method to handle opening of traces. An alternative could be to manage the signal in a parent component like trace-viewer.tsx or trace-explorer-widget.tsx and move handleOpenTrace functionality to a utils file or service

theia-extensions/viewer-prototype/src/browser/trace-viewer/trace-viewer-toolbar-contribution.tsx
```diff
@@ -82,6 +83,18 @@ export class TraceViewerToolbarContribution implements TabBarToolbarContribution
...
+        registry.registerCommand(
+            TraceViewerToolbarCommands.OPEN_NEW_TRACE, {
+            isVisible: (w: Widget) => {
+                if (w instanceof TraceExplorerOpenedTracesWidget) {
+                    return true;
+                }
+                return false;
+            },
+            execute: () => {
+                signalManager().fireOpenNewTraceSignal();
```
bhufmann, line 95: you can just call the other open command: registry.executeCommand(OpenTraceCommand.id); No need for the signal.

theia-extensions/viewer-prototype/src/browser/trace-viewer/trace-viewer-toolbar-commands.ts
```diff
@@ -31,6 +31,12 @@ export namespace TraceViewerToolbarCommands {
         label: 'Markers',
         iconClass: 'fa fa-bars fa-lg',
     };
+
+    export const OPEN_NEW_TRACE: Command = {
```
bhufmann, line 35: OPEN_TRACE

theia-extensions/viewer-prototype/src/browser/trace-viewer/trace-viewer-toolbar-commands.ts
```diff
@@ -31,6 +31,12 @@ export namespace TraceViewerToolbarCommands {
...
+        id: 'trace.viewer.openTrace',
+        label: 'Open New Trace',
```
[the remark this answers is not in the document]
bhufmann: I agree, it should be just `Open Trace` since it's the same command than the other one

bhufmann (changes requested):
It can be simplified, no need for all the signal related code.

theia-extensions/viewer-prototype/src/browser/trace-viewer/trace-viewer-toolbar-contribution.tsx
bhufmann: I understand now why the it was moved to the place holder, i.e. to execute the health check as well. So, the command handler of the open command should to the heath check as part of the command execution. That means the code from the placeholder widget has to move into the command-handler. Then the command can be executed from multiple places and the behavior is the same.

hriday-panchasara: @bhufmann Command handler of the open command now does the health check as well. Works well, but not sure if my reorganization is 100% correct

ebugden (changes requested):
Requested changes:

- [ ] Split changes into two commits (same PR). These are distinct changes even if one is a prerequisite for the other. 
  * Commit 1: Move server health check to Open Trace command (the commit message should justify why this change was made)
  * Commit 2: Add intuitive way to open additional traces
- [ ] Nitpick: Remove the grey translucent rectangle that appears over the right half of the button when you long hover (see video below)

https://user-images.githubusercontent.com/28311615/145594722-87c804a7-7a33-472e-a243-a9b5088ce4d3.mp4

----

I can look at the code more closely after the commit is split because then it will take me less time to understand.  

I'm excited to merge this! It really is a great UX improvement. Thank you for the update @hriday-panchasara! 🌿 This new code organization makes sense to me. Thank you for the recommendation @bhufmann!

hriday-panchasara:
> Requested changes:
> 
> * [ ]  Split changes into two commits (same PR). These are distinct changes even if one is a prerequisite for the other.
>   
>   * Commit 1: Move server health check to Open Trace command (the commit message should justify why this change was made)
>   * Commit 2: Add intuitive way to open additional traces
> * [ ]  Nitpick: Remove the grey translucent rectangle that appears over the right half of the button when you long hover (see video below)
> 
>  TraceCompassTutorialTraces.Theia-Trace.Example.Application.-.Google.Chrome.2021-12-06.09-18-40.mp4 
> I can look at the code more closely after the commit is split because then it will take me less time to understand.
> 
> I'm excited to merge this! It really is a great UX improvement. Thank you for the update @hriday-panchasara! 🌿 Also this new code organization makes sense to me. Thank you for the recommendation @bhufmann!

- Split the changes into two commits so its more readable
- I'm unable to replicate the issue with the button overlay, its appearing normal for me in light and dark themes. Might be a theia bug since we're contributing a command through its api:
Screen Shot 2021-12-10 at 10 34 05 AM
Screen Shot 2021-12-10 at 10 34 33 AM

theia-extensions/viewer-prototype/src/browser/trace-viewer/trace-viewer-toolbar-contribution.tsx
```diff
@@ -82,6 +84,18 @@ export class TraceViewerToolbarContribution implements TabBarToolbarContribution
                 signalManager().fireResetZoomSignal();
             }
         });
+        registry.registerCommand(
+            TraceViewerToolbarCommands.OPEN_TRACE, {
```
ebugden, line 88: It's necessary to register an additional command? It's not possible to just directly use the "OpenTraceCommand"? (This is a genuine question. I don't know how commands work.)

ebugden (changes requested):
Requested change:
- [ ]  Remove "fixes #193" from the "Move server health check to Open Trace command" commit. This commit is a prerequisite for fixing that issue, but it doesn't fix it. If you want to link it to the issue you could say "Required for fixing #193".

I'm being particular about commit structure and messages because it makes things much easier for future devs to understand and make changes to the code. Imagine if you're trying to understand a file in a new code base and by using "git blame" you can get clear descriptions of the purpose of each line of code from the devs that originally added them. That sounds great to me!

Apart from that, these changes look good to me!. The division of the code in each commit makes sense to me as well as the changes themselves. Thank you for taking the time to adjust the commits. Again, I'll leave it to more experienced coders to give the final approval.

ebugden:
> * I'm unable to replicate the issue with the button overlay, its appearing normal for me in light and dark themes. Might be a theia bug since we're contributing a command through its api:
>   Screen Shot 2021-12-10 at 10 34 05 AM
>   Screen Shot 2021-12-10 at 10 34 33 AM

No problem! It's not happening for me either in the most recent version of these changes so who knows 😄

hriday-panchasara:
> Requested change:
> 
> * [ ]   Remove "fixes #193" from the "Move server health check to Open Trace command" commit. This commit is a prerequisite for fixing that issue, but it doesn't fix it. If you want to link it to the issue you could say "Required for fixing #193".
> 
> I'm being particular about commit structure and messages because it makes things much easier for future devs to understand and make changes to the code. Imagine if you're trying to understand a file in a new code base and by using "git blame" you can get clear descriptions of the purpose of each line of code from the devs that originally added them. That sounds great to me!
> 
> Apart from that, these changes look good to me!. The division of the code in each commit makes sense to me as well as the changes themselves. Thank you for taking the time to adjust the commits. Again, I'll leave it to more experienced coders to give the final approval.

Reworded the commit message

theia-extensions/viewer-prototype/src/browser/trace-viewer/trace-viewer-toolbar-contribution.tsx
hriday-panchasara: Since we cannot add buttons to phosphor widget headers, we're creating a toolbar command (https://theia-ide.org/docs/commands_keybindings/) to create a custom action in the widget toolbar, similar to the way we add zoom-in & zoom-out buttons to the TabBar Widget when Resources view is opened 

The OPEN_TRACE command has the label and folder icon assigned to it, and we register the command using the registerCommand function where we specify that we want it visible only for the opened traces widget. The registerCommand function also specifies callback function (OpenTraceCommand) for when its triggered.

theia-extensions/viewer-prototype/src/browser/trace-explorer/trace-explorer-sub-widgets/trace-explorer-placeholder-widget.tsx
```diff
@@ -79,44 +66,10 @@ export class TraceExplorerPlaceholderWidget extends ReactWidget {
...
-                    }
-                    progress.cancel();
-                }
-                this.state.loading = false;
-                this.update();
-            });
-        }
+        this.state.loading = true;
+        this.update();
+        this.commandService.executeCommand(OpenTraceCommand.id);
```
bhufmann, line 71: I think `await` is needed here. What do you think?

theia-extensions/viewer-prototype/src/browser/trace-viewer/trace-viewer-toolbar-contribution.tsx
```diff
@@ -82,6 +84,18 @@ export class TraceViewerToolbarContribution implements TabBarToolbarContribution
...
+            isVisible: (w: Widget) => {
+                if (w instanceof TraceExplorerOpenedTracesWidget) {
+                    return true;
+                }
+                return false;
+            },
+            execute: () => {
+                registry.executeCommand(OpenTraceCommand.id);
```
bhufmann, line 96: `await` here to, right? If yes, you will need to change line 95 to add async modifier (`execute: async`)

bhufmann (changes requested):

theia-extensions/viewer-prototype/src/browser/trace-viewer/trace-viewer-contribution.ts
```diff
@@ -113,7 +113,39 @@ export class TraceViewerContribution extends WidgetOpenHandler<TraceViewerWidget
...
+                            if (PortBusy.is(err as any)) {
+                                this.messageService.error(
+                                    `Error opening serial port ${this.port}. (Port busy)`);
+                            } else {
+                                this.messageService.error(
+                                    'Failed to start the trace server: no such file or directory. Please make sure that the path is correct in Trace Viewer settings and retry');
+                            }
+                            progress.cancel();
+                        }
+                    });
```
mirsky-work, line 146: Extracting that block to a separate function, called something like launchTraceServer() would improve readability - for your discretion.

mirsky-work (commented):
Works well. Thanks for the contribution!

Please format the message of the commit 8fe9de8 to look like a paragraph (capital letter at the begging of a sentence, punctuations, lines capped at 80 chars, etc.).

mirsky-work:
> * I'm unable to replicate the issue with the button overlay, its appearing normal for me in light and dark themes. Might be a theia bug since we're contributing a command through its api:

I've seen it several times in the latest patch, but can't reliably reproduce it. Anyway, it's not a big issue and shouldn't block this PR.

theia-extensions/viewer-prototype/src/browser/trace-viewer/trace-viewer-contribution.ts
```diff
@@ -57,6 +57,40 @@ export class TraceViewerContribution extends WidgetOpenHandler<TraceViewerWidget
         };
     }
 
+    protected async launchTraceServer(): Promise<void> {
```
PatrickTasse, line 60: Looks like this method extraction belongs in the previous commit?

PatrickTasse (changes requested):

ebugden (commented):
Thank you for addressing the comments @hriday-panchasara! From a functional perspective I'm comfortable merging this and I'll leave it to others to give approval of the code.

ebugden (approved):

bhufmann (approved):
Looks good to me. Thanks for the contribution!

Rodrigoplp-work (approved):

PatrickTasse (approved):

PR #592, opened 2021-12-06
Fix filter tree alignment

PatrickTasse:
Replace spans with divs in first column of filter tree. Use flexbox to
create fixed width indentation box containing an optional right-aligned
expand/collapse button. Remove unnecessary padding in all buttons.

### DIFF
--- a/packages/react-components/src/components/utils/filtrer-tree/checkbox-component.tsx
+++ b/packages/react-components/src/components/utils/filtrer-tree/checkbox-component.tsx
@@ -30,8 +30,8 @@ export class CheckboxComponent extends React.Component<CheckboxProps> {
     };
 
     render(): JSX.Element {
-        return <span style={{padding: 5}} onClick={this.handleClick}>
+        return <div style={{ paddingRight: 5, display: 'inline' }} onClick={this.handleClick}>
                 {this.renderCheckbox(this.props.checkedStatus)}
-            </span>;
+            </div>;
     }
 }
--- a/packages/react-components/src/components/utils/filtrer-tree/table-row.tsx
+++ b/packages/react-components/src/components/utils/filtrer-tree/table-row.tsx
@@ -32,13 +32,12 @@ export class TableRow extends React.Component<TableRowProps> {
     };
 
     renderToggleCollapse = (): React.ReactNode => {
-        const marginLeft = this.props.level * 15 + 'px';
+        const width = (this.props.level + 1) * 12;
         return (
             (this.props.node.children.length === 0)
-                ? <span style={{ marginLeft: marginLeft, paddingLeft: 20 }}></span>
-                : <span style={{ marginLeft: marginLeft, paddingRight: 5 }} onClick={this.handleCollapse}>
-                    {(this.isCollapsed() ? icons.expand : icons.collapse)}
-                </span>
+                ? <div style={{ width, paddingRight: 5, display: 'inline-block' }} />
+                : <div style={{ width, paddingRight: 5, textAlign: 'right', display: 'inline-block' }} onClick={this.handleCollapse}>
+                    {(this.isCollapsed() ? icons.expand : icons.collapse)}</div>
         );
     };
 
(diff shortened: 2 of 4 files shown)


bhufmann (dismissed):
Looks good to me and looks cleaner. Thanks.

Rodrigoplp-work (changes requested):
Good work. There is a small misalignment between the bottom border of adjacent cells. I'm afraid our solution with "flex" caused that issue.
I propose to use "inline" instead, which seems to work. I added my solution as comments in the review below, for your consideration.

packages/react-components/src/components/utils/filtrer-tree/checkbox-component.tsx
```diff
@@ -30,8 +30,8 @@ export class CheckboxComponent extends React.Component<CheckboxProps> {
     };
 
     render(): JSX.Element {
-        return <span style={{padding: 5}} onClick={this.handleClick}>
+        return <div style={{paddingRight: 5}} onClick={this.handleClick}>
```
Rodrigoplp-work, line 33: Use inline as in: `return <div style={{ paddingRight: 5, display: 'inline' }} onClick={this.handleClick}>`

packages/react-components/src/components/utils/filtrer-tree/table-cell.tsx
```diff
@@ -12,9 +12,10 @@ export class TableCell extends React.Component<TableCellProps> {
     }
 
     render(): React.ReactNode {
+        const style = this.props.index === 0 ? { display: 'flex' } : undefined;
```
Rodrigoplp-work, line 15: Remove flex. Use instead: `const style = this.props.index === 0 ? { margin: '0 5px' } : undefined;`

packages/react-components/src/components/utils/filtrer-tree/table-row.tsx
```diff
@@ -32,13 +32,14 @@ export class TableRow extends React.Component<TableRowProps> {
     };
 
     renderToggleCollapse = (): React.ReactNode => {
-        const marginLeft = this.props.level * 15 + 'px';
+        const width = (this.props.level + 1) * 12 + 'px';
```
Rodrigoplp-work, line 35: Replace with: `const indentation = (this.props.level + 1) * 12;`

packages/react-components/src/components/utils/filtrer-tree/table-row.tsx
```diff
@@ -32,13 +32,14 @@ export class TableRow extends React.Component<TableRowProps> {
...
             (this.props.node.children.length === 0)
-                ? <span style={{ marginLeft: marginLeft, paddingLeft: 20 }}></span>
-                : <span style={{ marginLeft: marginLeft, paddingRight: 5 }} onClick={this.handleCollapse}>
-                    {(this.isCollapsed() ? icons.expand : icons.collapse)}
-                </span>
+                ? <div style={{ width: width, paddingRight: 5 }}>&nbsp;</div>
+                : <div style={{ width: width, paddingRight: 5, display: 'flex' }} onClick={this.handleCollapse}>
+                    <div style={{ marginLeft: 'auto', marginRight: 0 }}>{(this.isCollapsed() ? icons.expand : icons.collapse)}
+                    </div>
+                </div>
```
Rodrigoplp-work, line 42: Works with inline:
```
(this.props.node.children.length === 0
  ? <div style={{ marginLeft: indentation + 12, display: 'inline' }} />
  : <div style={{ marginLeft: indentation, paddingRight: 5, display: 'inline' }} onClick={this.handleCollapse}>
    <div style={{ marginLeft: 'auto', marginRight: 0, display: 'inline' }}>
      {(this.isCollapsed() ? icons.expand : icons.collapse)}
    </div>
  </div>
)
```

PatrickTasse:
> Good work. There is a small misalignment between the bottom border of adjacent cells. I'm afraid our solution with "flex" caused that issue. I propose to use "inline" instead, which seems to work. I added my solution as comments in the review below, for your consideration.

Thanks! I noticed that 1 pixel misalignment but had no idea how to fix it!

packages/react-components/src/components/utils/filtrer-tree/table-cell.tsx
PatrickTasse: I didn't find that this margin was necessary.

packages/react-components/src/components/utils/filtrer-tree/table-row.tsx
PatrickTasse: This reverts the fix because it puts back the variable icon width into the indentation. But it's on the right track, using 'inline-block' the desired indentation width can be respected. We can also set text-align: right to align the icon inside.

Rodrigoplp-work (approved):

bhufmann (approved):
Thanks!

PR #596, opened 2021-12-07
add unit tests for the events table

hriday-panchasara:
Created some unit tests for user-defined functions and components used in the events table. Was unable to create comprehensive tests for TableOutputComponent since its entirely built upon the AG-Grid library, and snapshot testing does not show the rendered HTML of AG-Grid. Another way to test TableOutputComponent could be to stub api calls made by the component to fetch table data and check if the internal state of the component (table columns & lines) updates. I was unable to mock tsp-client api calls or stub responses since jest could not reference the requests being made by the component's methods. But I believe we can later add to this test once we figure out a way to do so. Also included snapshot and unit tests for TableRendererComponents which are used by TableOutputComponent

### DIFF
--- a/packages/react-components/src/components/__tests__/__snapshots__/table-renderer-components.test.tsx.snap
+++ b/packages/react-components/src/components/__tests__/__snapshots__/table-renderer-components.test.tsx.snap
@@ -1,0 +1,76 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`Table output component tests Cell renderer 1`] = `"test cell"`;
+
+exports[`Table output component tests Cell renderer with search selection 1`] = `
+Array [
+  <span>
+    
+    <span
+      style={
+        Object {
+          "backgroundColor": "#FFFF00",
+        }
+      }
+    >
+      test
+    </span>
+  </span>,
+  <span>
+     cell
+  </span>,
+]
+`;
+
+exports[`Table output component tests Empty search filter renderer 1`] = `
+<div
+  data-testid="search-filter-element-parent"
+  onClick={[Function]}
+  onMouseEnter={[Function]}
+  onMouseLeave={[Function]}
+>
+  <svg
+    aria-hidden="true"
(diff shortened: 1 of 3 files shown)


packages/react-components/src/components/__tests__/events-table-component-tests.tsx
```diff
@@ -0,0 +1,243 @@
...
+});
+
+describe('Table output component tests', () => {
+    
+    test('Renders AG-Grid table with provided props & state', async () => {
+        const tableOutputComponentProps: AbstractOutputProps = {
+            tooltipComponent: null,
+            style: {
+                width: 0,
+                chartWidth: 0,
```
PatrickTasse, line 26: Sorry for taking so long to review. The AbstractOutputProps have been updated and some of these properties have changed. The unit test fails to compile on latest master.

PatrickTasse (changes requested):

packages/react-components/src/components/__tests__/events-table-component-tests.tsx
hriday-panchasara: @PatrickTasse Rebased on top of master. The unit tests should pass now

PatrickTasse (approved):

Rodrigoplp-work (changes requested):
Excellent work!

Please change the name of the test file from `event-table-component-tests.tsx` to `table-renderer-components.test.tsx`, and the respective snapshot.

That follows the way we did with the other tests and makes it easier for other developers to identify what has already been covered when writing more tests.

hriday-panchasara:
> Excellent work!
> 
> Please change the name of the test file from `event-table-component-tests.tsx` to `table-renderer-components.test.tsx`, and the respective snapshot.
> 
> That follows the way we did with the other tests and makes it easier for other developers to identify what has already been covered when writing more tests.

@Rodrigoplp-work renamed the relevant files

PatrickTasse (approved):

Rodrigoplp-work (approved):

Issue #329, opened 2021-03-26
Discover ability to open a group of traces vs a single trace file from Open Traces dialog


ebugden:
Right now, the only way to open a group of traces together is by knowing you can select a folder that contains multiple traces (completely implicit). We need a more discoverable way to allow users to open a group of traces. 

Distinguish between "merged trace" and "trace group"? See [#194 (comment)](https://github.com/theia-ide/theia-trace-extension/issues/194#issuecomment-808519438)

----
Issue source: @ssmagula's user tests + brainstorming with TC and tool team members

ebugden:
[#332 (comment)](https://github.com/theia-ide/theia-trace-extension/issues/332#issuecomment-812049646) also relevant for this issue:
> **ebugden:** @tahini and @arfio recently explained that **all traces are opened as experiments in the Trace Viewer (even if there's only one trace)**. So maybe there doesn't need to be a huge visual distinction between a single trace and a trace group, it just needs to be clear how many traces are in the trace group.

Issue #324, opened 2021-03-26
Discover compatible trace files


ebugden:
Finding compatible trace files is required for using the Trace Viewer. Users who can't find compatible trace files are 100% blocked from using the tool. Users must be able to:
* **Locate trace files** - Where are trace files?
* **Identify compatible trace files** - Which files are trace file? Which trace files are compatible formats?
* **Identify what to select to correctly open the trace** - Ex. What is the root of the trace file? 

## Proposed solution elements
* **Magic Folder:** A user-configurable folder. User identifies folder paths that should be watched by the Trace Viewer and scanned for traces. Detected files automatically appear in UI.
* **Scan system for compatible trace files** #8 
   * Should file hierarchy be kept or not? Ex. Fade out non-compatible files in the browsable file structure? Only show a list of detected compatible files (squash all file hierarchy)?
* **Explicitly filter for certain trace formats:** Open Traces dialog presents a user editable filter, initially set to 'CTF'? NB: User could always dismiss the filter, and browse to any and all files and a tree filter already implemented in Theia. Default filters could be configurable by settings. 

----
Issue source: @ssmagula's user tests + brainstorming with TC and tool team members

bhufmann:
Discover of trace files are domain and tracer specific, where the tracer is the application that produces traces. 

First of all the location where traces are going to be stored is not fixed across tracers.  For example LTTng produces their traces by default into the user's home directory under ${user.home}/lttng-traces. The location can be configured by users. 

Secondly, the file format, filenames, file endings are not fix either. LTTng, for example, use the CTF format which is a directory with a metadata file and binary files. Other tracer may create traces in JSON format some text format. Just looking on the file extension won't be sufficient.  

Discover of compatible trace files need to be configurable and extensible.

ebugden:
Thank you for the details! I'll update the description with these challenges.

ebugden:
**What would our version of the "popular trace formats" list be?**

Ex. Perfetto claims it's interoperable because it "can import and export popular trace formats: Chromium JSON trace format, Android Systrace, ftrace, CSV"

![image](https://user-images.githubusercontent.com/28311615/113933983-38113200-97c3-11eb-9a7a-645395e8b593.png)

@bhufmann @MatthewKhouzam @arfio What are the 3-5 most common trace file formats:
* For the whole tracing universe?
* For our intended users?

----

[...]

PR #597, opened 2021-12-08
Clarify in the UI that open traces searches recursively

hriday-panchasara:
Users who are new to the tool, or are unfamiliar with CTF, having trouble navigating to the exact location of traces on the disk
- Added a label above the Open Trace button saying: "Select the root directory to begin recursive search for trace files."
- Label is open to suggestions

Screen Shot 2021-12-08 at 11 02 30 AM

### DIFF
--- a/theia-extensions/viewer-prototype/src/browser/trace-explorer/trace-explorer-sub-widgets/trace-explorer-placeholder-widget.tsx
+++ b/theia-extensions/viewer-prototype/src/browser/trace-explorer/trace-explorer-sub-widgets/trace-explorer-placeholder-widget.tsx
@@ -59,7 +59,8 @@ export class TraceExplorerPlaceholderWidget extends ReactWidget {
     render(): React.ReactNode {
         const { loading } = this.state;
         return <div className='theia-navigator-container' tabIndex={0}>
-            <div className='center'>{'You have not yet opened a trace.'}</div>
+            <div className='center'>{'You have not yet opened a trace.'}</div><br />
+            <div className='center'>{'Select the root directory to begin recursive search for trace files.'}</div>
             <div className='open-workspace-button-container'>
                 <button className='theia-button open-workspace-button' title='Select a trace to open'
                     onClick={this.handleOpenTrace} disabled={loading}>


ebugden:
Thank you for taking the initiative to start clarifying this information!

Could you clarify the purpose of the message?
* Is this message to help users who are "having trouble navigating to the exact location of traces on the disk"? The message proposed doesn't explain where trace files are.
* Is the message to explain that if you select a folder that recursively contains traces that it will open all the CTF traces recursively as a single group? Creating a trace group is a special case that only makes sense with traces we want to compare on the same timeline. It shouldn't be the default.
* Is the message to explain that to open a CTF trace you need to select the root folder?

I agree that all of the above points need clarification (and that maybe a label could help with this while the trickier trace opening design problems are handled next year), but I don't think the current proposed phrasing helps with these issues.

I also think it would be more appropriate to have a message like this in the file explorer window after users click on "Open Trace".

mirsky-work:
> Could you clarify the purpose of the message?
> 
> * Is this message to help users who are "having trouble navigating to the exact location of traces on the disk"? The message proposed doesn't explain where trace files are.

That was the intention more or less. The use case is a complex folder structure that contains test source files, as well as build and run artifacts, and we'd like to spare the user from having to drill down that folder to locate the trace files. The label cannot explain where the traces are located - it can be anywhere on the file system...

> * Is the message to explain that if you select a folder that recursively contains traces that it will open all the CTF traces recursively as a single group? Creating a trace group is a special case that only makes sense with traces we want to compare on the same timeline. It shouldn't be the default.

That's a good point - we can add to the message that traces would be "merged" or overlayed (I don't think users are exposed to the term _Trace Group_ in the UI).

ebugden:
After chatting with @ssmagula I would suggest closing this PR and taking a different approach to the problem. There are several pieces of information (about opening traces) that are unclear for the user at the moment, but this information is too complex to be communicated clearly with a single message.

## What information is missing?
During user tests ssmagula observed that user don't know:
* Where trace files typically are
* What a trace files looks like. What the structure of a trace file is (ex. CTF traces are a folder with several files)
* That traces are autogrouped if the parent folder is selected #329 

## Helping users find traces
Helping users effectively find compatible traces #324 is a complex design problem that will be tackled seriously next year. 

If we want to make this slightly easier for users in time for the MVP, ssmagula and I suggest adding a message at the top of the "Open Trace" dialogue, ex. "Having trouble finding trace? Learn more about typical trace locations." that has a configurable link to documentation. (see [comment](https://github.com/theia-ide/theia-trace-extension/issues/324#issuecomment-995216563) on #324)

## Helping users discover how to open a trace group
Since this feature is specialized and easy to misuse as it is currently implemented (it only works on CTF, it only makes sense when you want to compare traces), I would suggest we don't explicitly point to it yet.

mirsky-work:
> Since this feature is specialized and easy to misuse as it is currently implemented (it only works on CTF, it only makes sense when you want to compare traces), I would suggest we don't explicitly point to it yet.

Opening a trace group is actually a very common use case, when a user wants to get a complete picture and examine several traces that were produced during the same tracing session, overlayed together.

I agree to drop this PR and rethink how to best approach that. Thanks for your feedback!

PR #598, opened 2021-12-14
Add data tree output component implementation

bhufmann:
The implementation is the first iteration. It introduces the component
class which uses the tree implementation also used in the time-graph and
XY components. The component is not enabled in this patch.

The tree implementation need some UI/UX improvement, for example for
styling or column resizing which will be done in a separated commit.

Once some improvements are available, the data tree component
will be shown as part of the available views list for a given trace.


bhufmann:
To test replace the implementation of method `getOutputDescriptors()` in file  `trace-explorere-views-widget.tsx` with the implementation below:
``` javascript
    protected async getOutputDescriptors(experiment: Experiment): Promise<OutputDescriptor[]> {
        const outputDescriptors: OutputDescriptor[] = [];
        const descriptors = await this._experimentManager.getAvailableOutputs(experiment.UUID);
        }
        /*
         * Remove certain scatter specific outputs since they are not supported in the theia-traceviewer
         * There is no generic way to identify such outputs. This is not ideal because other server
         * implementations are not covered.
         */
        return outputDescriptors.filter(value => !value.id.includes('scatter'));
    }
```

packages/react-components/src/components/trace-context-component.tsx
```diff
@@ -357,12 +358,14 @@ export class TraceContextComponent extends React.Component<TraceContextProps, Tr
...
                             return <TableOutputComponent key={output.id} {...outputProps} />;
+                        case 'DATA_TREE':
+                            return <DataTreeOutputComponent key={output.id} {...outputProps} />;
                         default:
                             return <NullOutputComponent key={output.id} {...outputProps} />;
                     }
                 })}
             </ResponsiveGridLayout>
-            {(outputs.length > 1 || outputs[0].type !== 'TABLE') &&
+            {(outputs.length > 1 || outputs[0].type !== 'TABLE' && outputs[0].type !== 'DATA_TREE') &&
```
PatrickTasse, line 368: It shouldn't add the TimeNavigatorComponent unconditionally where there are 2 or more outputs. It should do some fancy loop lookup to see if at least one of the outputs has a time axis. I wonder if it would be better to do a positive check for timegraph/xy instead of a negative check for table/datatree?

Also this condition should be applied to determine if TimeAxisComponent should be created. I would put the comment above but you know, github.

PatrickTasse (changes requested):

PatrickTasse (approved):

PR #602, opened 2021-12-15
Add 'Nothing selected' when a chart is empty

MatthewKhouzam:
I can't find the issue that it was solving. It's something like "show data by default"

### DIFF
--- a/packages/react-components/src/components/xy-output-component.tsx
+++ b/packages/react-components/src/components/xy-output-component.tsx
@@ -232,7 +232,15 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
                 }
             </React.Fragment>;
         }
-
+        if (this.state.outputStatus === ResponseStatus.COMPLETED && this.state.xyData?.datasets?.length===0 ) {
+            return <React.Fragment>
+                {
+                    <p className='no-data' style={{fontSize: 20, marginRight: '5px', marginLeft: '5px', justifyContent:'center', alignItems:'center'}}>
+                        Select a checkbox to see analysis results
+                    </p>
+                }
+            </React.Fragment>;
+        }
         return <React.Fragment>
             {this.state.outputStatus === ResponseStatus.COMPLETED ?
                 <div id='xy-main' tabIndex={0}


MatthewKhouzam:
Before a chart would be empty if there was no data in it. Now there's a warning. 
nothing-selected

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -232,6 +232,16 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
                 }
             </React.Fragment>;
         }
+        if (this.state.outputStatus === ResponseStatus.COMPLETED && this.state.checkedSeries.length === 0 ) {
```
MatthewKhouzam, line 235: How about !!this.state.xyData? to handle those views that have total?

MatthewKhouzam (commented):
Needs an issue number

ebugden:
> I can't find the issue that it was solving. It's something like "show data by default"

That's my fault! I never got around to creating the issue. Here you go: #604

ebugden (changes requested):
Requested changes:

- [x] Simplify message. Suggestion: "Check a box to see analysis results"

I still have a strong preference for just checking the first box by default (for the views that don't have an obvious default) instead of showing a message: 
* In my mind, seeing a checked box shows that checking boxes is possible and where they can be selected in the UI. A checked box clearly tutorializes the action way more explicitly than a message. 
* I also think this better tutorialization outweighs the fact that if we just check the first box, the default won't necessarily be what they want to look at or the most relevant data.

That being said, this code is done and makes the UI more robust so after simplifying the message, let's merge it even if we do something else after! Thank you for these changes @MatthewKhouzam! 🍰 

Are there other reasons this solution was proposed instead of checking a box by default (ex. A message is significantly easier to implement)?

----

Justification for message change:
* We don't need to say "Trace analysis completed" since this is implied when we tell them they can view results 
* I don't think users would understand what "the tree" is. I only understood because I've seen the code. "Check a box" is more specific guidance.

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -232,6 +232,16 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
                 }
             </React.Fragment>;
         }
+        if (this.state.outputStatus === ResponseStatus.COMPLETED && this.state.xyData?.datasets?.length===0 ) {
+            return <React.Fragment>
+                {
+                    <p className='no-selected' style={{fontSize: 20, marginRight: '5px', marginLeft: '5px', justifyContent:'center', alignItems:'center'}}>
+                        <p>Trace analysis completed.</p>
+                        <p>Select an element in the tree to visualize output.</p>
```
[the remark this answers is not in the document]
bhufmann: What about: "Select a checkbox to see analysis results"

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -232,6 +232,16 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
                 }
             </React.Fragment>;
         }
+        if (this.state.outputStatus === ResponseStatus.COMPLETED && this.state.xyData?.datasets?.length===0 ) {
+            return <React.Fragment>
+                {
+                    <p className='no-selected' style={{fontSize: 20, marginRight: '5px', marginLeft: '5px', justifyContent:'center', alignItems:'center'}}>
```
bhufmann, line 238: I see the following error in the console:
```
root ERROR Warning: validateDOMNesting(...): <p> cannot appear as a descendant of <p>.
    in p (created by XYOutputComponent)
    in p (created by XYOutputComponent)
    in div (created by XYOutputComponent)
    in div (created by XYOutputComponent)
    in div (created by XYOutputComponent)
    in XYOutputComponent (created by TraceContextComponent)
    in Resizable (created by GridItem)
    in DraggableCore (created by GridItem)
    in GridItem (created by ReactGridLayout)
    in div (created by ReactGridLayout)
    in ReactGridLayout (created by ResponsiveReactGridLayout)
    in ResponsiveReactGridLayout (created by WidthProvider)
    in WidthProvider (created by TraceContextComponent)
    in div (created by TraceContextComponent)
    in TraceContextComponent
    in div
```

bhufmann (changes requested):
Nice improvements. Please reference issue #406 and fix linting failure.

packages/react-components/src/components/xy-output-component.tsx
MatthewKhouzam: Huh... I'll remove it then... it works though. ?!?

packages/react-components/src/components/xy-output-component.tsx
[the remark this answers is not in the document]
MatthewKhouzam: Picking Bernd's because I checked my email late.

MatthewKhouzam:
Fixes #604. Updated commit message.

Rodrigoplp-work (approved):

bhufmann (approved):
Thanks. Looks good to me.

MatthewKhouzam:
@ebugden Any thoughts on the changes?

ebugden (approved):
Looks good to me! Thank you for these changes @MatthewKhouzam!

bhufmann (approved):
It still looks good to me. Thanks!

Rodrigoplp-work (approved):

MatthewKhouzam:
Thanks all!

PR #603, opened 2021-12-15
Available views cleared when browser refreshed

### LINKED ISSUE #599: Available views not empty after browser refresh
When there is no view open (e.g. events table) the list of available views is cleared. However, after doing a browser refresh the list of available views is populated using the first trace in the list of opened traces. The list should be empty instead.

avail-views-not-empty

hriday-panchasara:
fixes #599

I think in opened traces widget the default value for selectedExperimentIndex should be -1 so that it does not select the first experiment by default (index 0) when the component is initialized.


bhufmann (approved):
Thanks for providing this fix!

Rodrigoplp-work (approved):

PR #605, opened 2021-12-15
Fix output component scrolling and layout

PatrickTasse:
Since the timegraph-output-component creates two trees that each have
their own scrollbar, remove the scroll overflow from the parent
output-component-tree.

Add scrollable wrapper to the xy-output-component and
datatree-output-component trees. Make the timegraph-output-component
wrappers scrollable.

Set the main-output-component overflow to hidden to prevent the
timegraph-output-component markers tree from overflowing when the output
area is too small.

Remove the tree scroll synchronization from the parent class
AbstractTreeOutputComponent and instead add it to the only tree that
needs it, the timegraph tree.

Fix tree width calculation in AbstractTreeOutputComponent.

Fix markers layer height calculation to ignore scrollbar height when the
tree is collapsed.

Remove filler row from markers tree and chart when the model is empty.

Remove arbitrary main area width from abstract-output-component.

### DIFF
--- a/packages/react-components/src/components/abstract-output-component.tsx
+++ b/packages/react-components/src/components/abstract-output-component.tsx
@@ -101,7 +101,7 @@ export abstract class AbstractOutputComponent<P extends AbstractOutputProps, S e
         if (this.mainAreaContainer.current) {
             return this.mainAreaContainer.current.clientWidth;
         }
-        return 1000;
+        return 0;
     }
 
     public getHandleWidth(): number {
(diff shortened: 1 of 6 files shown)


bhufmann:
@PatrickTasse , do you have a screenshot for the before and after?

Also, how can I reproduce what you are fixing with this PR. 

Thanks
Bernd

PatrickTasse:
Before and after fix:

before-#605

after-#605

You can see in the after screenshot:
- Top time graph's marker chart doesn't overflow over the event table
- Bottom time graph's tree vertical scroll bar doesn't have unused space to its right between the scroll bar and the chart
- Bottom time graph's collapsed marker tree and chart take less space

bhufmann (approved):
Looks good to me. It fixes issues as described.

Rodrigoplp-work (approved):

Rodrigoplp-work (approved):

bhufmann (approved):
Still good for me after the rebase. Thanks!

PR #606, opened 2021-12-16
Make table-tree columns resizable

PatrickTasse:
Add resizable property to ColumnHeader model.

Add a resize handle to each resizable column header cell.

Add listeners to resize columns when the handle is dragged, and reset to
preferred column width when the handle is double-clicked.

Add filler column to make table-tree extend to full container width.

Change the table display, styling, borders, and header color.

Make all output component trees use the same base table-tree styling
with only timegraph-tree adding some specific styling overrides.

### DIFF
--- a/packages/react-components/style/output-components-style.css
+++ b/packages/react-components/style/output-components-style.css
@@ -97,27 +97,84 @@ canvas {
     align-self: center;
 }
 
-.table-tree>tbody>tr:nth-child(1)>th {
-    background-color: var(--theia-editor-background);
-    position: sticky;
-    top: 0;
+.table-tree {
+    display: grid;
+    border-collapse: collapse;
+    border-left: 1px solid var(--theia-tree-inactiveIndentGuidesStroke);
 }
 
-.table-tree th, .table-tree td {
-    padding: 3px 5px;
+.table-tree thead, .table-tree tbody, .table-tree tr {
+    display: contents;
+}
+
+.table-tree th {
+    position: relative;
+    padding: 3px 0px 3px 5px;
+    background-color: var(--theia-list-hoverBackground) !important;
+    border-top: 1px solid var(--theia-tree-inactiveIndentGuidesStroke);
+    border-bottom: 1px solid var(--theia-tree-inactiveIndentGuidesStroke);
     text-align: left;
-    border-bottom: 1px solid #333;
-    border-right: 1px solid #333;
+}
+
+.table-tree tr:hover td {
+    background-color: var(--theia-list-hoverBackground) !important;
+}
+
+.table-tree td {
+    padding: 3px 5px;
+    border-bottom: 1px solid var(--theia-tree-inactiveIndentGuidesStroke);
+    border-right: 1px solid var(--theia-tree-inactiveIndentGuidesStroke);
+    text-align: left;
+}
+
+.table-tree th span {
+    overflow: hidden;
     white-space: nowrap;
-    min-width: 50px;
+    display: block;
+}
+
+.table-tree td span {
+    text-overflow: ellipsis;
+    overflow: hidden;
+    white-space: nowrap;
+    display: block;
+}
+
+.resize-handle {
+    float: right;
+    position: absolute;
+    top: 1px;
+    right: 0px;
+    color: var(--theia-tree-inactiveIndentGuidesStroke);
+    background-color: var(--theia-list-hoverBackground);
+    cursor: col-resize;
+}
+
+.sort-icon {
+    position: absolute;
+    top: 3px;
+    right: 3px;
+    padding: 0px 3px;
+    background-color: var(--theia-list-hoverBackground) !important;
+}
+
+.filler {
+    padding: 0px !important;
+    border-right: 1px solid var(--theia-tree-inactiveIndentGuidesStroke);
+}
+
+.timegraph-tree {
+    border: 0px; 
 }
 
 .timegraph-tree tr {
     /* TODO: Fix row alignment, this number is arbitrary, it works [on my machine], but it should match line height in timeline-chart */
     line-height: 18px;
-    position: relative;
-    white-space: nowrap;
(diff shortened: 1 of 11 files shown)


packages/react-components/style/output-components-style.css
```diff
@@ -93,27 +97,80 @@ canvas {
     align-self: center;
 }
 
-.table-tree>tbody>tr:nth-child(1)>th {
-    background-color: var(--theia-editor-background);
-    position: sticky;
-    top: 0;
+.table-tree {
```
bhufmann, line 100: I think it looks good to highlight the row when hovering on a row... the events table, file explorer, open trace view etc. do that. I only manged to highlight a cell by adding the lines below. Maybe you can figure it out.

.table-tree td:hover {
    background-color: var(--theia-list-hoverBackground) !important;
}

packages/react-components/style/output-components-style.css
```diff
@@ -93,27 +97,80 @@ canvas {
...
+    display: grid;
+    border-collapse: collapse;
+    border-left: 1px solid #CCC;
```
bhufmann, line 103: #CCC is to bright when in dark theme. We should use a themed color of theia. I could find one that is meant for tree borders,  but var(--theia-tree-inactiveIndentGuidesStroke) seems to be a good one. Also add a right border which looks good to have on the right.
```
     border-left: 1px solid var(--theia-tree-inactiveIndentGuidesStroke) !important;
     border-right: 1px solid var(--theia-tree-inactiveIndentGuidesStroke) !important;
```

packages/react-components/style/output-components-style.css
```diff
@@ -93,27 +97,80 @@ canvas {
...
+}
+
+.table-tree thead, .table-tree tbody, .table-tree tr {
+    display: contents;
+}
+
+.table-tree th {
+    position: relative;
+    padding: 3px 0px 3px 5px;
+    background-color: var(--theia-list-hoverBackground) !important;
```
bhufmann, line 113: It would be good to match the header background color of the events table. However, that's not straight forward.  Probably it should have a separate theming task to handle themes better in all the views.

packages/react-components/style/output-components-style.css
```diff
@@ -93,27 +97,80 @@ canvas {
...
+    border-top: 1px solid #CCC;
```
bhufmann, line 114: var(--theia-tree-inactiveIndentGuidesStroke)

packages/react-components/style/output-components-style.css
```diff
@@ -93,27 +97,80 @@ canvas {
...
+    border-bottom: 1px solid #CCC;
```
bhufmann, line 115: var(--theia-tree-inactiveIndentGuidesStroke)

packages/react-components/style/output-components-style.css
```diff
@@ -93,27 +97,80 @@ canvas {
...
+    text-align: left;
 }
 
-.table-tree th, .table-tree td {
+.table-tree td {
     padding: 3px 5px;
+    border-bottom: 1px solid #CCC;
```
bhufmann, line 121: var(--theia-tree-inactiveIndentGuidesStroke)

packages/react-components/style/output-components-style.css
```diff
@@ -93,27 +97,80 @@ canvas {
...
+    border-right: 1px solid #CCC;
```
bhufmann, line 122: var(--theia-tree-inactiveIndentGuidesStroke)

bhufmann (changes requested):
It works well and functionality is there. Some suggestion for the theming.

packages/react-components/style/output-components-style.css
Rodrigoplp-work: Here is a solution to highlight the whole line. I suggest adding it below line 138.
```
.table-tree td:hover, .table-tree td:hover ~ td {
    background-color: var(--theia-list-hoverBackground);
}
```

bhufmann: I think you need the ` !important` at the end.

Rodrigoplp-work: Not needed:
Screenshot 2021-12-21 at 15 35 53

bhufmann: I just saw it for all other `var(...)` references in that css file and thought it would be needed.

PatrickTasse: > Here is a solution to highlight the whole line. I suggest adding it below line 138.
> 
> ```
> .table-tree td:hover, .table-tree td:hover ~ td {
>     background-color: var(--theia-list-hoverBackground);
> }
> ```

This is some kind of sibling notation right? Good to know!

I had missed your comment and found that this also works:

.table-tree tr:hover td {

packages/react-components/style/output-components-style.css
PatrickTasse: I got a weird result adding the border-right to the table. The table border was not moved when resizing the columns.

The border-right is provided by the td elements. I had disabled it on the filler column. I removed that and forced it on the filler elements also, which includes the th filler, since th don't have that right border because of the resize handle.

Note that I made cells have bottom and right border only because border-collapse didn't seem to work right (so there was double-border between cells). Maybe that could have been fixed some other way.

packages/react-components/style/output-components-style.css
Rodrigoplp-work: Yes Patrick, it's using CSS siblings. But your solution works, too.
And we should only use `!important` when really necessary, for that it breaks CSS inheritance. In this case, it works fine without it.

bhufmann (approved):
Thanks for updating the colors and the hover background. I think it looks good now and a nice improvement over the previous tree implementation (e.g. the CPU usage view tree).

I approve, but will give the 2nd reviewers  the final word .

Rodrigoplp-work (approved):

PR #607, opened 2021-12-16
Enable data tree outputs

PatrickTasse:


MatthewKhouzam:
@ssmagula @ebugden This provides `DATA_TREE` support to show function duration statistics. Do we want to merge this as-is, we would like some user-focused feedback. We feel it is good, we want to make sure we have a good first impression as we think these tables will be the first thing people look at.

The main ask is: **Do you agree with the look and feel?**

bhufmann (dismissed):
Looks good to me  and fixes the issues as described.

ssmagula:
I just looked at this with Patrick. The name made it seem more intimidating than it was. Nutshell: Looks pretty good to me. The "Filter" text entry box could use a boost in discoverability.Take a look at the Filter text entry box in the Events Table. Follow that pattern for styling/interaction. A Filter icon might help it look more actionable.

Totally nitpicky: the table header labels are different color / styling vs the Events Table header labels. I'd choose one <th> style and stick with it.

These are minor things, but there are other, more significant improvements that could be made

PatrickTasse:
Added issues for suggested future improvements:

https://github.com/theia-ide/theia-trace-extension/issues/622
https://github.com/theia-ide/theia-trace-extension/issues/623

bhufmann (approved):

Rodrigoplp-work (approved):

PR #609, opened 2021-12-16
Add undo/redo functionality(hardcode)

ywang001:
Currently, the timeRange is stored in undo/redo stack so that the user is able to redo/undo zoom in/zoom out/pan left/pan right. I tried to make it generic by storing the entire context instead of just the timeRange so that other settings such as backgroundTheme can be undone/redone as well. I tried undoStack.push(this) but it doesn’t seem to be the right way.

https://github.com/typescript-eslint/typescript-eslint/blob/v3.10.1/packages/eslint-plugin/docs/rules/no-this-alias.md
Also, 'const self = this' is not allowed as indicated in the link above. That's why one check is not successful, but it works well locally on my laptop.

Any advice is appreciated!
Signed-off-by: Yining Wang <yining.wang@ericsson.com>

### DIFF
--- a/packages/react-components/src/components/trace-context-component.tsx
+++ b/packages/react-components/src/components/trace-context-component.tsx
@@ -55,8 +55,8 @@ interface TraceContextState {
 interface UndoRedoFunctionality {
     undoStack: bigint[][];
     redoStack: bigint[][];
-    undo: () => void;
-    redo: () => void;
+    undo: () => (bigint[] | undefined);
+    redo: () => (bigint[] | undefined);
 }
 
 export class TraceContextComponent extends React.Component<TraceContextProps, TraceContextState, UndoRedoFunctionality> {
@@ -141,7 +141,6 @@ export class TraceContextComponent extends React.Component<TraceContextProps, Tr
         this.props.addResizeHandler(this.onResize);
         this.initialize();
         this.subscribeToEvents();
-        const self = this;
         this.undoRedo = {
             undoStack: [],
             redoStack: [],
@@ -151,9 +150,7 @@ export class TraceContextComponent extends React.Component<TraceContextProps, Tr
                     if (lastUndo !== undefined) {
                         console.log('undo');
                         this.redoStack.push(lastUndo);
-                        const prevStart = lastUndo[0];
-                        const prevEnd = lastUndo[1];
-                        self.unitController.viewRange = { start: prevStart, end: prevEnd };
+                        return lastUndo;
                     }
                     else {
                         console.log('undefined', lastUndo);
@@ -162,6 +159,7 @@ export class TraceContextComponent extends React.Component<TraceContextProps, Tr
                 else {
                     console.log('can\'t undo anymore');
                 }
+                return undefined;
             },
             redo: function () {
                 if (this.redoStack.length > 0) {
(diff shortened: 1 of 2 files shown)


packages/react-components/src/components/trace-context-component.tsx
```diff
@@ -52,7 +52,14 @@ interface TraceContextState {
     backgroundTheme: string;
 }
 
-export class TraceContextComponent extends React.Component<TraceContextProps, TraceContextState> {
+interface UndoRedoFunctionality {
+    undoStack: bigint[][];
+    redoStack: bigint[][];
+    undo: () => void;
```
MatthewKhouzam, line 58: Maybe add TraceContextComponent to this component as an argument...

packages/react-components/src/components/trace-context-component.tsx
```diff
@@ -132,6 +141,48 @@ export class TraceContextComponent extends React.Component<TraceContextProps, Tr
         this.props.addResizeHandler(this.onResize);
         this.initialize();
         this.subscribeToEvents();
+        const self = this;
+        this.undoRedo = {
```
MatthewKhouzam, line 145: (self : TraceContextComponent) => { blabla

packages/react-components/src/components/trace-context-component.tsx
```diff
@@ -132,6 +141,48 @@ export class TraceContextComponent extends React.Component<TraceContextProps, Tr
         this.props.addResizeHandler(this.onResize);
         this.initialize();
         this.subscribeToEvents();
+        const self = this;
```
MatthewKhouzam, line 144: remove me!

MatthewKhouzam (commented):
I like this, as a first step, we need to eventually serialize/deserialize the whole context. I suggest looking into json conversions... But the suggestion I put should help.

packages/react-components/src/components/trace-context-component.tsx
wangyining-221: I tried to store TraceContextComponent in the stack and add TraceContextComponent as an argument, but it is passed by reference. As a result, no value can be changed. Instead, if undo function returns the viewRange, it is possible to do 'this.viewRange = undo()', and it passed four checks. I also pushed modified code and a new commit message.

wangyining-221:
I pushed a new commit, and the link is attached below.
https://github.com/theia-ide/theia-trace-extension/pull/609/commits/c164f655336c2bbe199576b6a1b7dee0bfd86446

Since it is not allowed to assign 'this', the value of viewRange cannot be changed in the function. Thus, I change the return type of undo/redo function from void to bigint[]. In specific, it returns a list of bigint which are start and end of viewRange. Then, this.unitController.viewRange can be changed after the function is called.

I tried a few ways to expand it to a generic undo/redo function. However, none of it works.

1. To store TraceContextComponent
I tried to store 'this' which is TraceContextComponent in the undoStack. However, it is similar to passing by reference, so no value can be changed if we do 'this = undoStack.pop()'.

2. To store TraceContextProps
I also used undoStack to store TraceContextProps, but 'this.props = undoStack.pop()' is not allowed because props is read only.

3. To store state
I used 'this.setState(currState: prevState)' but this.state does not include every property we want to keep track of. For example, this.state does not include unitController which contains viewRange.

bhufmann:
Thanks a lot again for this initial effort!

PR #733 finalized this feature. I'm closing this one since it is not needed anymore.

PR #611, opened 2021-12-20
Make filter width match first column width

PatrickTasse:
Add resize observer on first column header to synchronize filter width.

Remove text message on empty filter result as this disrupts the observer
and prevents synchronization after the non-matching filter is removed.


bhufmann (commented):
The commit "Make filter width first column width" looks good to me and works as described. This PR depends on other PR to be merged. So, I can approve it when dependent PRs are merged.

Rodrigoplp-work (approved):

PatrickTasse (commented):
Bernd's previous approval applied.

Issue #311, opened 2021-03-19
Set keyboard focus on graph/table when selecting an analysis


PatrickTasse:
I would suggest the following improvements:

- When clicking on an analysis in the Available Analyses list, in the selected trace tab, this analysis's graph/table should be given keyboard focus and show the focus blue border. The trace tab should be scrolled to make that analysis visible, whether it is being opened for the first time or if it was already created previously. If the visible page is already full, opening a new analysis currently is only noticed by the scrollbar thumb position/size being updated, as the new analysis is out of view.

- When clicking on a trace tab, the keyboard focus should be given to the first, or last selected, analysis's graph/table in the tab.

- When clicking on an analysis title in the trace tab, the keyboard focus should be given to its graph/table

ebugden:
Related: #271

hriday-panchasara:
Closed via #612

PR #612, opened 2021-12-23
Set keyboard focus on graph/table when selecting an analysis

### LINKED ISSUE #414: Pulse view title on click if view is already open
If a View that is already open is selected in the sidebar, the View's title area should pulse with a highlight to temporarily bring attention to the View panel. Right now, there's no feedback for the user in this situation so it looks like the open action failed.

Related: #271

hriday-panchasara:
This PR has changes for the following features suggested by @PatrickTasse in #311
- When clicking on an analysis in the Available Analysis list, in the selected trace tab, the analysis is scrolled into view and given keyboard focus. User can click tab to access the tree and chart of the focused graph.
focus-example-1

- The trace tab should be scrolled to make that analysis visible, whether it is being opened for the first time or if it was already created previously.
focus-example-2

- When clicking on an analysis title in the trace tab, keyboard focus should be the analysis' graph/chart region. This is indicated by a blue focus border, allowing the user to directly use keyboard interactions/shortcuts
focus-example-3

fixes #414
- If a View that is already open is selected in the sidebar, the View's title area should pulse to temporarily bring attention to the View panel. Keyboard focus is also given to the analysis' graph/chart region.
focus-example-4

For some reason the focus outline is not being shown properly in some of the screen recordings. This is the default tab focus styling but we can change the css to make the outline more prominent.

### DIFF
--- a/packages/react-components/src/components/abstract-output-component.tsx
+++ b/packages/react-components/src/components/abstract-output-component.tsx
@@ -63,7 +63,8 @@ export abstract class AbstractOutputComponent<P extends AbstractOutputProps, S e
         const localStyle = Object.assign({}, this.props.style);
         localStyle.width = this.props.widthWPBugWorkaround;
         return <div style={localStyle}
-            id={this.props.outputDescriptor.id}
+            id={this.props.traceId + this.props.outputDescriptor.id}
+            tabIndex={-1}
             className={'output-container ' + this.props.className}
             onMouseUp={this.props.onMouseUp}
             onMouseDown={this.props.onMouseDown}
@@ -71,7 +72,11 @@ export abstract class AbstractOutputComponent<P extends AbstractOutputProps, S e
             onTouchEnd={this.props.onTouchEnd}
             data-tip=''
             data-for="tooltip-component">
-            <div className='widget-handle' style={{ width: this.props.style.handleWidth, height: this.props.style.height }}>
+            <div
+                id={this.props.traceId + this.props.outputDescriptor.id + 'handle'}
+                className='widget-handle'
+                style={{ width: this.props.style.handleWidth, height: this.props.style.height }}
+            >
                 {this.renderTitleBar()}
             </div>
             <div className='main-output-container' ref={this.mainOutputContainer}
@@ -88,7 +93,7 @@ export abstract class AbstractOutputComponent<P extends AbstractOutputProps, S e
             <button className='remove-component-button' onClick={this.closeComponent}>
                 <FontAwesomeIcon icon={faTimes} />
             </button>
-            <div className='title-bar-label' title={outputName}>
+            <div className='title-bar-label' title={outputName} onClick={() => this.setFocus()}>
                 {outputName}
             </div>
         </React.Fragment>;
@@ -120,6 +125,7 @@ export abstract class AbstractOutputComponent<P extends AbstractOutputProps, S e
 
         return this.renderMainArea();
     }
+    abstract setFocus(): void;
 
     abstract renderMainArea(): React.ReactNode;
 
--- a/packages/react-components/src/components/datatree-output-component.tsx
+++ b/packages/react-components/src/components/datatree-output-component.tsx
@@ -81,7 +81,11 @@ export class DataTreeOutputComponent extends AbstractOutputComponent<AbstractOut
         this.onToggleCollapse = this.onToggleCollapse.bind(this);
         this.onOrderChange = this.onOrderChange.bind(this);
         return this.state.xyTree.length
-            ? <div className='scrollable' style={{ height: this.props.style.height, width: this.getMainAreaWidth() }}>
+            ?   <div
+                    tabIndex={0}
+                    id={this.props.traceId + this.props.outputDescriptor.id + 'focusContainer'}
+                    className='scrollable' style={{ height: this.props.style.height, width: this.getMainAreaWidth() }}
+                >
                 <EntryTree
                     entries={this.state.xyTree}
                     showCheckboxes={false}
@@ -94,6 +98,7 @@ export class DataTreeOutputComponent extends AbstractOutputComponent<AbstractOut
             : undefined
             ;
     }
+
     renderMainArea(): React.ReactNode {
         return <React.Fragment>
             {this.state.outputStatus === ResponseStatus.COMPLETED ?
@@ -102,13 +107,22 @@ export class DataTreeOutputComponent extends AbstractOutputComponent<AbstractOut
                 >
                     {this.renderTree()}
                 </div> :
-                <div className='analysis-running-main-area'>
+                <div tabIndex={0} id={this.props.traceId + this.props.outputDescriptor.id + 'focusContainer'} className='analysis-running-main-area'>
                     <i className='fa fa-refresh fa-spin' style={{ marginRight: '5px' }} />
                     <span>Analysis running</span>
                 </div>
             }
         </React.Fragment>;
     }
+
+    setFocus(): void {
+        if (document.getElementById(this.props.traceId + this.props.outputDescriptor.id + 'focusContainer')) {
+            document.getElementById(this.props.traceId + this.props.outputDescriptor.id + 'focusContainer')?.focus();
+        } else {
+            document.getElementById(this.props.traceId + this.props.outputDescriptor.id)?.focus();
+        }
+    }
+
     private onToggleCollapse(id: number, nodes: TreeNode[]) {
         let newList = [...this.state.collapsedNodes];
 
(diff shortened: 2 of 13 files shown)


yarn.lock
```diff
@@ -4728,6 +4728,11 @@ alphanum-sort@^1.0.1, alphanum-sort@^1.0.2:
   resolved "https://registry.yarnpkg.com/alphanum-sort/-/alphanum-sort-1.0.2.tgz#97a1119649b211ad33691d9f9f486a8ec9fbe0a3"
   integrity sha1-l6ERlkmyEa0zaR2fn0hqjsn74KM=
 
+animate.css@^4.1.1:
+  version "4.1.1"
+  resolved "https://registry.yarnpkg.com/animate.css/-/animate.css-4.1.1.tgz#614ec5a81131d7e4dc362a58143f7406abd68075"
+  integrity sha512-+mRmCTv6SbCmtYJCN4faJMNFVNN5EuCTTprDTAo7YzIGji2KADmakjVA3+8mVDkZ2Bf09vayB35lSQIex2+QaQ==
+
```
PatrickTasse, line 4735: These lines get removed automatically for me when I do yarn and then it fails to compile. I think you need to add a dependency in theia-extensions/viewer-prototype/package.json "dependencies"

    "animate.css": "^4.1.1"

PatrickTasse (changes requested):
When clicking on an output title, the focus is given to the whole output container. I think it should be given to the output's chart or table:
- for a timegraph or xy chart, so that we can then use WASD keys to zoom and pan
- for the event table so that we can use the arrow keys to navigate the table

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -359,6 +362,10 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
         </React.Fragment>;
     }
 
+    setFocus(): void {
+        document.getElementById(this.props.outputDescriptor.id+'chartContainer')?.focus();
```
PatrickTasse, line 366: It appears the call to focus() also makes the component scroll into view.

Therefore this scrolling does not happen when there is no series selected, when there are no results or when the analysis is still running (text displayed instead of chart).

Perhaps there could be a super() implementation to give focus to the container as a fall-back? This could be used also as a fall-back for timegraph and as a default for data-tree.

packages/react-components/src/components/null-output-component.tsx
```diff
@@ -24,4 +24,7 @@ export class NullOutputComponent extends AbstractOutputComponent<NullOutputProps
             </div>
         </React.Fragment>;
     }
+    setFocus(): void {
```
PatrickTasse, line 27: add blank line

packages/react-components/src/components/datatree-output-component.tsx
```diff
@@ -128,4 +128,7 @@ export class DataTreeOutputComponent extends AbstractOutputComponent<AbstractOut
         // fix Warning: Can't perform a React state update on an unmounted component
         this.setState = (_state, _callback) => undefined;
     }
+    setFocus(): void {
```
PatrickTasse, line 131: Add blank line

packages/react-components/src/components/datatree-output-component.tsx
```diff
@@ -128,4 +128,7 @@ export class DataTreeOutputComponent extends AbstractOutputComponent<AbstractOut
...
+        return;
```
PatrickTasse, line 132: Something should be done to allow data trees so scroll into view even if there is no need for keyboard focus.

packages/react-components/src/components/table-output-component.tsx
```diff
@@ -88,20 +88,22 @@ export class TableOutputComponent extends AbstractOutputComponent<TableOutputPro
...
+        this.checkFocus = this.checkFocus.bind(this);
     }
 
     renderMainArea(): React.ReactNode {
         if (this.state.outputStatus === ResponseStatus.FAILED) {
             return this.analysisFailedMessage();
         }
 
-        return <div id='events-table'
+        return <div id={this.props.outputDescriptor.id+'chartContainer'}
```
PatrickTasse, line 104: Not a big deal, but chartContainer isn't appropriate for this component that only has a table.

PatrickTasse (changes requested):

packages/react-components/src/components/xy-output-component.tsx
hriday-panchasara: @PatrickTasse Made use of conditional statements to call focus() on the parent abstract-output-component if the focus container does not exist

packages/react-components/src/components/timegraph-output-component.tsx
```diff
@@ -454,14 +456,22 @@ export class TimegraphOutputComponent extends AbstractTreeOutputComponent<Timegr
...
+            id={this.props.outputDescriptor.id+'focusContainer'}
             layers={[
                 grid, this.chartLayer, selectionRange, this.chartCursors, this.arrowLayer, this.rangeEventsLayer
             ]}
         >
         </ReactTimeGraphContainer>;
     }
 
+    setFocus(): void {
+        if (document.getElementById(this.props.outputDescriptor.id+'focusContainer')) {
```
PatrickTasse, line 468: It doesn't seem to work correctly when you open the same output for a different trace tab. I am guessing there are multiple elements with the same id.

Perhaps there is a way to find an element that is strictly a child of another specific element?

This comment applies to all outputs.

PatrickTasse (changes requested):

theia-extensions/viewer-prototype/src/browser/trace-viewer/trace-viewer.tsx
```diff
@@ -316,7 +317,22 @@ export class TraceViewerWidget extends ReactWidget {
                 await this.fetchAnnotationCategories(output);
                 this.update();
             } else {
-                document.getElementById(exist.id)?.scrollIntoView();
+                const traceId = this.openedExperiment.UUID;
+                if (document.getElementById(traceId + exist.id + 'focusContainer')) {
+                    document.getElementById(traceId + exist.id + 'focusContainer')?.focus();
```
PatrickTasse, line 322: This doesn't work for the datatree, it outputs this warning to console:

> root WARN Widget was activated, but did not accept focus after 2000ms

It appears that datatree has no component that accepts focus so focus() cannot be used to make it scroll into view?

Also, my intention was to use setFocus() method for an output component to implement to decide which of its children should receive focus. But here we do not use this method, instead the id suffix 'focusContainer' is used to indicate which one it should be. It seems we should not have two ways to achieve the same goal? Maybe scrolling into view and assigning focus should be two different steps?

PatrickTasse (changes requested):

theia-extensions/viewer-prototype/src/browser/trace-viewer/trace-viewer.tsx
hriday-panchasara: @PatrickTasse updated datatree-output-component and rebased to resolve conflicts

packages/react-components/src/components/abstract-output-component.tsx
```diff
@@ -62,15 +62,20 @@ export abstract class AbstractOutputComponent<P extends AbstractOutputProps, S e
         const localStyle = Object.assign({}, this.props.style);
         localStyle.width = this.props.widthWPBugWorkaround;
         return <div style={localStyle}
-            id={this.props.outputDescriptor.id}
+            id={this.props.traceId+this.props.outputDescriptor.id}
```
PatrickTasse, line 65: spaces around +

packages/react-components/src/components/abstract-output-component.tsx
```diff
@@ -62,15 +62,20 @@ export abstract class AbstractOutputComponent<P extends AbstractOutputProps, S e
...
             className={'output-container ' + this.props.className}
             onMouseUp={this.props.onMouseUp}
             onMouseDown={this.props.onMouseDown}
             onTouchStart={this.props.onTouchStart}
             onTouchEnd={this.props.onTouchEnd}
             data-tip=''
             data-for="tooltip-component">
-            <div className='widget-handle' style={{ width: this.props.style.handleWidth, height: this.props.style.height }}>
+            <div
+                id={this.props.traceId+this.props.outputDescriptor.id+'handle'}
```
PatrickTasse, line 75: spaces around +

PatrickTasse (approved):

PatrickTasse (approved):

bhufmann:
@hriday-panchasara could you please rebase this PR? Then a final round of review can be done. Thanks

hriday-panchasara:
> @hriday-panchasara could you please rebase this PR? Then a final round of review can be done. Thanks

@bhufmann Rebased on top of master

PatrickTasse (approved):

Rodrigoplp-work (approved):

PR #614, opened 2022-01-05
Display loading message on data tree components

Rodrigoplp-work:
While the data tree component is loading data, display an "analysis
running" message in the middle of the component.

The data tree components are still hidden, as metioned in #598.

Contributes to #144.

### DIFF
--- a/packages/react-components/src/components/datatree-output-component.tsx
+++ b/packages/react-components/src/components/datatree-output-component.tsx
@@ -85,11 +85,17 @@ export class DataTreeOutputComponent extends AbstractOutputComponent<AbstractOut
     }
     renderMainArea(): React.ReactNode {
         return <React.Fragment>
-            <div ref={this.treeRef} className='output-component-tree'
-                style={{ height: this.props.style.height, width: this.props.widthWPBugWorkaround }}
-            >
-                {this.renderTree()}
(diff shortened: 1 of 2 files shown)


Rodrigoplp-work:
To display the data tree component and test this PR, replace the implementation of method `getOutputDescriptors()` in file `trace-explorere-views-widget.tsx` with:
```js
    protected async getOutputDescriptors(experiment: Experiment): Promise<OutputDescriptor[]> {
        const outputDescriptors: OutputDescriptor[] = [];
        const descriptors = await this._experimentManager.getAvailableOutputs(experiment.UUID);
        }
        /*
         * Remove certain scatter specific outputs since they are not supported in the theia-traceviewer
         * There is no generic way to identify such outputs. This is not ideal because other server
         * implementations are not covered.
         */
        return outputDescriptors.filter(value => !value.id.includes('scatter'));
    }
```
Also, if the "analysis running" message is not showing up because the component is too fast at loading data, replace line 88 of `packages/react-components/src/components/datatree-output-component.tsx` with:

```js
{this.state.outputStatus === ResponseStatus.RUNNING ?
```

PatrickTasse (approved):

MatthewKhouzam (approved):

PR #615, opened 2022-01-06
Add keyboard shortcuts cheatsheet

hriday-panchasara:
'Trace Viewer Keyboard and Mouse Shortcuts' dialog:
Screen Shot 2022-01-06 at 15 26 02 PM
Screen Shot 2022-01-06 at 15 26 46 PM

Dialog can be triggered by:
1) Clicking info icon that appears in toolbar when a time-graph or xy-chart is opened:
Screen Shot 2022-01-06 at 3 31 44 PM
2) Using Theia's 'find command' search:
Screen Shot 2022-01-06 at 15 34 36 PM
3) Using keybinding 'Shift+F1'

### DIFF
--- a/packages/react-components/src/components/xy-output-component.tsx
+++ b/packages/react-components/src/components/xy-output-component.tsx
@@ -746,7 +746,7 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
 
     private buildScatterData(seriesObj: XYSeries[]) {
         const dataSetArray = new Array<any>();
-        let xValues: number[] = [];
+        let xValues: bigint[] = [];
         let yValues: number[] = [];
         let pairs: xyPair[] = [];
         const offset = this.getChartOffset();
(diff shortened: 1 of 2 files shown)


PatrickTasse (changes requested):
- The rows alternate background color and the darker ones have a hover effect. I suggest all to be the same background color with no hover effect as it's not selectable.
- I would suggest to reorder the SELECT section: Select element/Select time on the left, Select time range/Extend time range on the right. Or top and bottom, but just don't have the two time range diagonal to each other...

hriday-panchasara:
> * The rows alternate background color and the darker ones have a hover effect. I suggest all to be the same background color with no hover effect as it's not selectable.
> * I would suggest to reorder the SELECT section: Select element/Select time on the left, Select time range/Extend time range on the right. Or top and bottom, but just don't have the two time range diagonal to each other...

Here's the updated look:
Screen Shot 2022-01-12 at 16 42 37 PM

PR #620, opened 2022-01-14
Add horizontal resize sash to output components

PatrickTasse:
Add a sash between the tree and the chart. For XY chart insert the sash
between the Y-axis and the chart.

Add mouse handling to the sash to call the setSashOffset() callback when
moving the mouse while the sash is clicked, to propage the new sash
offset to all output components.

Refactor the width handling to use handle width, sash offset and sash
width to compute the tree width and chart width.

Add hover color and cursor to the sash.


bhufmann (approved):
Nice improvement. Thanks for the contribution!

PR #627, opened 2022-01-19
Data Tree: Add "analysis failed" and "no results" messages

ebugden:
Messages are now handled in the abstract output component so less code is required in each output component.

When error messages are handled in each component rather (than by the
parent class), it is easy to forget to handle these errors when
implementing new output components. Handling errors in the parent class
makes these bugs less likely.

"No results" message: Instead of needing to remember to call the
function in each output components the abstract class structure imposes
that every output component implement a check for empty results (and
displaying the message is handled by the parent class).

Analysis failed message: The developper still needs to remember to set
the output status to failed when implementing new output components.
This change does not help with remembering to do that.

### DIFF
--- a/packages/react-components/src/components/abstract-output-component.tsx
+++ b/packages/react-components/src/components/abstract-output-component.tsx
@@ -111,11 +111,11 @@ export abstract class AbstractOutputComponent<P extends AbstractOutputProps, S e
 
     private renderMainOutputContainer(): React.ReactNode {
         if (this.state.outputStatus === ResponseStatus.FAILED) {
-            return this.analysisFailedMessage();
+            return this.renderAnalysisFailed();
         }
 
         if (this.state.outputStatus === ResponseStatus.COMPLETED && this.resultsAreEmpty()) {
-            return this.emptyResultsMessage();
+            return this.renderEmptyResults();
         }
 
         return this.renderMainArea();
@@ -125,7 +125,7 @@ export abstract class AbstractOutputComponent<P extends AbstractOutputProps, S e
 
     abstract resultsAreEmpty(): boolean;
 
-    protected analysisFailedMessage(): React.ReactFragment {
+    protected renderAnalysisFailed(): React.ReactFragment {
         return <React.Fragment>
             <div className='message-main-area'>
                 Trace analysis failed.
@@ -133,7 +133,7 @@ export abstract class AbstractOutputComponent<P extends AbstractOutputProps, S e
         </React.Fragment>;
     }
 
-    protected emptyResultsMessage(): React.ReactFragment {
+    protected renderEmptyResults(): React.ReactFragment {
         return <React.Fragment>
                 <div className='message-main-area'>
                     Trace analysis complete.
--- a/packages/react-components/src/components/abstract-output-component.tsx
+++ b/packages/react-components/src/components/abstract-output-component.tsx
@@ -9,6 +9,7 @@ import { OutputComponentStyle } from './utils/output-component-style';
 import { OutputStyleModel } from 'tsp-typescript-client/lib/models/styles';
 import { TooltipComponent } from './tooltip-component';
 import { TooltipXYComponent } from './tooltip-xy-component';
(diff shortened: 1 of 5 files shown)


packages/react-components/src/components/abstract-output-component.tsx
```diff
@@ -108,8 +109,22 @@ export abstract class AbstractOutputComponent<P extends AbstractOutputProps, S e
         return this.props.style.handleWidth;
     }
 
+    private checkAndRenderMainArea(): React.ReactNode {
```
ebugden, line 112: I wasn't sure what to call this function. Suggestions welcome!

packages/react-components/src/components/datatree-output-component.tsx
```diff
@@ -63,6 +63,9 @@ export class DataTreeOutputComponent extends AbstractOutputComponent<AbstractOut
             }
```
PatrickTasse, line 63: We need to set the state also when the tspClientResponse is OK but the treeResponse is not (e.g. treeResponse.model is null). It needs an else-clause, see timegraph-output-component.tsx.

PatrickTasse (changes requested):

packages/react-components/src/components/abstract-output-component.tsx
```diff
@@ -75,7 +76,7 @@ export abstract class AbstractOutputComponent<P extends AbstractOutputProps, S e
             </div>
             <div className='main-output-container' ref={this.mainAreaContainer}
                 style={{ width: this.props.widthWPBugWorkaround - this.props.style.handleWidth, height: this.props.style.height }}>
-                {this.renderMainArea()}
+                {this.checkAndRenderMainArea()}
```
PatrickTasse, line 79: I would suggest renderMainOutputContainer(), see line 77 className. I would perhaps also rename the ref variable to mainOutputContainer, to be consistent.

packages/react-components/src/components/abstract-output-component.tsx
```diff
@@ -108,8 +109,22 @@ export abstract class AbstractOutputComponent<P extends AbstractOutputProps, S e
...
+        }
+
+        return this.renderMainArea();
+    }
+
     abstract renderMainArea(): React.ReactNode;
 
+    abstract resultsAreEmpty(): boolean;
+
     protected analysisFailedMessage(): React.ReactFragment {
```
PatrickTasse, line 128: We suggest to rename this method and the next to renderAnalysisFailed() and renderEmptyResults(). Just in case a subclass overrides to show something else than a message.

packages/react-components/src/components/datatree-output-component.tsx
ebugden: Good catch! Thanks 🌼

packages/react-components/src/components/abstract-output-component.tsx
ebugden: Great suggestion! It feels clean and consistent.

packages/react-components/src/components/abstract-output-component.tsx
ebugden: My reflex tends to be to make things more general when it's needed rather than a priori, but I don't have very strong feelings about it in this case. It will mainly take a little more code reading to understand what these functions do.

I will make these changes!

ebugden:
Update:

- [x] Data Tree: Set state when treeResponse is not ok
- [x] Rename `checkAndRenderMainArea` to `renderMainOutputContainer`
- [x] Rename ref variable from `mainAreaContainer` to `mainOutputContainer`
- [x] Rename `analysisFailedMessage()` to `renderAnalysisFailed()`
- [x] Rename `emptyResultsMessage` to `renderEmptyResults()`
- [x] Rebase on master

Thank you @PatrickTasse for the feedback!

PatrickTasse (approved):

bhufmann (approved):
LGTM. Thanks!

PR #629, opened 2022-01-21
Enable tooltips for scatter plots.

Rodrigoplp-work:
These tooltips behave as in the line charts: when the mouse stops moving for a
couple of seconds the tooltip is triggered.

If there are points close to the mouse pointer, the tooltip displays a
list with the closest point of each series. If the points are too far, the
tooltip is not shown.

The tooltip displays the timestamp that corresponds to the position of
the mouse, and y-value of the closest points.

This commit also enables the scatter plots, since the two issues that
were blocking their release (bug about lack of precision and the absence
of tooltip) have been resolved.

https://user-images.githubusercontent.com/88502808/150555054-b6efa227-0e23-4c0e-a3ef-0ac72edecd72.mov

### DIFF
--- a/packages/react-components/src/components/xy-output-component.tsx
+++ b/packages/react-components/src/components/xy-output-component.tsx
@@ -28,6 +28,8 @@ type XYOuputState = AbstractOutputState & {
     // FIXME Type this properly
     xyData: any;
     columns: ColumnHeader[];
+    allMax: number;
+    allMin: number;
 };
 const RIGHT_CLICK_NUMBER = 2;
 const ZOOM_IN = true;
@@ -41,7 +43,7 @@ class xyPair {
     constructor(x: number, y: number) {
         this.x = x;
         this.y = y;
-      }
+    }
 }
 
 export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutputProps, XYOuputState> {
@@ -52,11 +54,13 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
     private yAxisRef: any;
     private mouseIsDown = false;
     private positionXMove = 0;
+    private positionYMove = 0;
     private isRightClick = false;
     private isMouseMove = false;
     private posPixelSelect = 0;
     private isMouseLeave = false;
     private startPositionMouseRightClick = BigInt(0);
+    private margin = { top: 15, right: 0, bottom: 5, left: this.props.style.yAxisWidth };
     private isScatterPlot: boolean = this.props.outputDescriptor.id.includes('scatter');
     private plugin = {
         afterDraw: (chartInstance: Chart, _easing: Chart.Easing, _options?: any) => { this.afterChartDraw(chartInstance); }
@@ -85,6 +89,8 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
             orderedNodes: [],
             xyData: {},
             columns: [{title: 'Name', sortable: true}],
+            allMax: 0,
+            allMin: 0
         };
 
         this.afterChartDraw = this.afterChartDraw.bind(this);
@@ -197,26 +203,13 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
     renderYAxis(): React.ReactNode {
         // Y axis with D3
         const chartHeight = parseInt(this.props.style.height.toString());
-        const margin = { top: 15, right: 0, bottom: 5, left: this.props.style.yAxisWidth };
-        let allMax = 0;
-        let allMin = 0;
-        this.state.xyData?.datasets?.forEach((dSet: any, i: number) => {
-            let rowMax;
-            let rowMin;
-            if (this.isScatterPlot) {
-                rowMax = Math.max(...dSet.data.map((d: any) => d.y));
-                rowMin = Math.min(...dSet.data.map((d: any) => d.y));
-            } else {
-                rowMax = Math.max(...dSet.data);
-                rowMin = Math.min(...dSet.data);
-            }
-            allMax = Math.max(allMax, rowMax);
-            allMin = i === 0 ? rowMin : Math.min(allMin, rowMin);
-        });
+
         const yScale = scaleLinear()
-            .domain([allMin, Math.max(allMax, 1)])
-            .range([chartHeight - margin.bottom, margin.top]);
-        const yTransform = `translate(${margin.left}, 0)`;
+            .domain([this.state.allMin, Math.max(this.state.allMax, 1)])
+            .range([chartHeight - this.margin.bottom, this.margin.top]);
+
+        const yTransform = `translate(${this.margin.left}, 0)`;
+
         // Abbreviate large numbers
         const scaleYLabel = (d: number) => (
             d >= 1000000000000 ? Math.round(d / 100000000000) / 10 + 'G' :
@@ -225,19 +218,22 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
             d >= 1000 ? Math.round(d / 100) / 10 + 'K':
             Math.round(d * 10) / 10
         );
-        if (allMax > 0) {
+
+        if (this.state.allMax > 0) {
             select(this.yAxisRef.current).call(axisLeft(yScale).tickSizeOuter(0).ticks(4)).call(g => g.select('.domain').remove());
             select(this.yAxisRef.current).selectAll('.tick text').style('font-size', '11px').text((d: any) => scaleYLabel(d));
         }
 
         return <React.Fragment>
-            <svg height={chartHeight} width={margin.left}>
+            <svg height={chartHeight} width={this.margin.left}>
                 <g className='y-axis' ref={this.yAxisRef} transform={yTransform} />
             </svg>
         </React.Fragment>;
     }
 
     chooseChart(): JSX.Element {
+        const offset = this.props.viewRange?.getOffset() ?? BigInt(0);
+
         const lineOptions: Chart.ChartOptions = {
             responsive: true,
             elements: {
@@ -267,14 +263,21 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
                     id: 'time-axis',
                     display: false,
                     ticks: {
-                        min: Number(this.props.viewRange?.getStart() - this.getChartOffset()),
-                        max: Number(this.props.viewRange?.getEnd() - this.getChartOffset())
+                        min: Number(this.props.viewRange?.getStart() - offset),
+                        max: Number(this.props.viewRange?.getEnd() - offset)
                     }
                 }],
-                yAxes: [{ display: false, stacked: false }]
+                yAxes: [{
+                    display: false,
+                    stacked: false,
+                    ticks: {
+                        max: this.state.allMax,
+                        min: this.state.allMin
+                    }
+                }]
             },
             animation: { duration: 0 },
-            events: [ 'mousedown' ],
+            events: ['mousedown'],
         };
 
         if (!this.isScatterPlot) {
@@ -312,7 +315,6 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
     }
 
     renderChart(): React.ReactNode {
-        // width={this.props.style.chartWidth}
         if (this.state.outputStatus === ResponseStatus.COMPLETED && this.state.xyData?.datasets?.length === 0) {
             return <React.Fragment>
                 <div className='chart-message'>
@@ -335,21 +337,41 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
                     {this.chooseChart()}
                 </div> :
                 <div className='analysis-running'>
-                {(
-                        <i
-                            className='fa fa-refresh fa-spin'
-                            style={{ marginRight: '5px' }}
-                        />
-                    )}
-                    {
-                    'Analysis running'
-                    }
-                </div>}
+                    <i className='fa fa-refresh fa-spin' style={{ marginRight: '5px' }} />
+                    <span>Analysis running</span>
+                </div>
+            }
         </React.Fragment>;
     }
 
     resultsAreEmpty(): boolean {
         return this.state.xyTree.length === 0;
+    }
+
+    private calculateYRange() {
+        let localMax = 0;
+        let localMin = 0;
+
+        if (this.state && this.state.xyData) {
+            this.state.xyData?.datasets?.forEach((dSet: any, i: number) => {
+                let rowMax;
+                let rowMin;
+                if (this.isScatterPlot) {
+                    rowMax = Math.max(...dSet.data.map((d: any) => d.y));
+                    rowMin = Math.min(...dSet.data.map((d: any) => d.y));
+                } else {
+                    rowMax = Math.max(...dSet.data);
+                    rowMin = Math.min(...dSet.data);
+                }
+                localMax = Math.max(localMax, rowMax);
+                localMin = i === 0 ? rowMin : Math.min(localMin, rowMin);
+            });
+        }
+
+        this.setState({
+            allMax: localMax * 1.01,
+            allMin: localMin * 0.99
+        });
     }
 
     private afterChartDraw(chart: Chart) {
@@ -513,8 +535,7 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
         if (wheel.shiftKey) {
             if (wheel.deltaY < 0) {
                 this.pan(PAN_LEFT);
-            }
-            else if (wheel.deltaY > 0) {
+            } else if (wheel.deltaY > 0) {
                 this.pan(PAN_RIGHT);
             }
         } else if (wheel.ctrlKey) {
@@ -547,13 +568,6 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
         return x;
     }
 
-    private getChartOffset(): bigint {
-        // Work around loss of precision when working with BigInt
-        let offset = this.props.viewRange.getStart() ?? BigInt(0);
-        offset = offset / BigInt(1000000000) * BigInt(1000000000);
-        return offset;
-    }
-
     private updateSelection(): void {
         if (this.props.unitController.selectionRange){
             const xStartPos = this.props.unitController.selectionRange.start;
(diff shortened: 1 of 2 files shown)


PatrickTasse (changes requested):
- Since it is a scatter plot, the displayed time should be the time of the point being tooltipped, not the hover time of the mouse. So maybe they should be showed in pairs?
- The tooltip sometimes shows a value (usually 0) for non-existing points (it looks like the code for line chart tooltip when lines < 10 is interfering).
- The tooltip sometimes appears when the mouse is not near any point.
- It would be nice improvement to have that the point being tooltipped be highlighted somehow.
- The chart.js highlights the point that is selected, but that's not necessarily the point being tooltipped. Could we replace point selection by point hover?

PatrickTasse:
Note: I tested with Futex Contention Analysis - Latency vs Time where the total number of series was only 5.

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -564,37 +578,90 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
     }
 
-    private tooltip(x: number, y: number): void {
+    private closestPoint(array: any, point: any, xRatio: number, yRatio: number, chartHeight: number): number | undefined {
+        const hypos: any[] = [];
+
+        array.forEach((ar: any) => {
+            // Convert x values to screen pixels
+            const pointX = point.x * xRatio;
+            const x = ar.x * xRatio;
```
PatrickTasse, line 587: When zoomed in the x values don't seem correct, e.g. a range of 40 px instead of 600 px for a time range of 180,000,000 ns.

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -564,37 +578,90 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
         if (this.props.unitController.numberTranslator) {
             timeLabel = this.props.unitController.numberTranslator(timeForX);
         }
         const chartWidth = this.chartRef.current.chartInstance.width;
+        const chartHeight = this.chartRef.current.chartInstance.height;
         const arraySize = this.state.xyData.labels.length;
         const index = Math.max(Math.round((xPos / chartWidth) * (arraySize - 1)), 0);
+        const offset = this.props.viewRange?.getOffset() ?? BigInt(0);
+        const end = Number(this.props.viewRange.getEnd() - offset);
+        const xRatio = chartWidth / end;
```
PatrickTasse, line 622: This should probably be chartWidth / (end - start). But there is still something wrong...

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -564,37 +578,90 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
+            // Convert x values to screen pixels
+            const pointX = point.x * xRatio;
+            const x = ar.x * xRatio;
+
+            // Convert y values to screen pixels
+            const pointY: number = chartHeight - point.y;
+            const y = (ar.y - this.state.allMin) * yRatio;
+
+            // Calculate hypotenuses
+            const hypotenuse = Math.sqrt(Math.pow(Math.abs(pointX - x), 2) + Math.pow(Math.abs(pointY - y), 2));
```
PatrickTasse, line 594: If you print the hypotenuse in real time, you will see that distance 0 is not exactly where the mouse is over a point. So there is something slightly off in the calculations somewhere.

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -564,37 +578,90 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
         }
     }
 
-    private tooltip(x: number, y: number): void {
+    private closestPoint(array: any, point: any, xRatio: number, yRatio: number, chartHeight: number): number | undefined {
+        const hypos: any[] = [];
+
+        array.forEach((ar: any) => {
+            // Convert x values to screen pixels
+            const pointX = point.x * xRatio;
```
PatrickTasse, line 586: pointX and pointY are constant in this method so they can be calculated before the loop.

packages/react-components/src/components/xy-output-component.tsx
PatrickTasse: Could it be something to do with the margins?

Is there a way for the chart to provide the exact coordinates of a data point instead of the client trying to figure it out?

packages/react-components/src/components/xy-output-component.tsx
Rodrigoplp-work: True. xRatio was being calculated with `offset` instead of `start`, which works only without zooming. Fixed.

packages/react-components/src/components/xy-output-component.tsx
Rodrigoplp-work: Using `end - start` instead of `end - offset` fixed it.

packages/react-components/src/components/xy-output-component.tsx

Rodrigoplp-work:
ChartJS allows the highlight of a point when hovering, which works fine if you are looking at a single process. But for multiple processes our tooltip shows a list with the closest point of each process, and that would not make sense with ChartJS highlighting only one point.

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -238,6 +231,8 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
     }
 
     chooseChart(): JSX.Element {
+        const offset = this.props.viewRange?.getOffset() ?? BigInt(0);
```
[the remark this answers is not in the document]
PatrickTasse: I realize it's an issue for the whole application to worry about later. We can keep it like that for now.

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -742,8 +818,7 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
         let xValues: bigint[] = [];
         let yValues: number[] = [];
         let pairs: xyPair[] = [];
-        const offset = this.getChartOffset();
-
+        const offset = this.props.viewRange.getOffset() ?? BigInt(0);
```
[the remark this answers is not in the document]
PatrickTasse: I realize it's an issue for the whole application to worry about later. We can keep it like that for now.

packages/react-components/src/components/xy-output-component.tsx
MatthewKhouzam: remove the square root, please, useless calculations

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -564,37 +578,93 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
+            // If there are more than 10 lines in the chart, summarise the ones that are equal to 0.
+            if (!invalidPoint) {
+                if (this.state.xyData.datasets.length <= 10 || rounded > 0) {
+                    const point: any = {
+                        label: d.label,
+                        color: d.borderColor,
+                        background: d.backgroundColor,
+                        value: formatted
+                    };
+                    points.push(point);
```
MatthewKhouzam, line 663: Can you only create the point and push it, if it's closer than the previous point?

MatthewKhouzam: Screenshot from 2022-01-27 15-51-30
Look at all that GC! :)

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -564,37 +578,93 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
+        const hypos: any[] = [];
+        // Convert x and y values to screen pixels
+        const pointX = point.x * xRatio;
+        const pointY: number = chartHeight - point.y;
+
+        array.forEach((ar: any) => {
+            const x = ar.x * xRatio;
+            const y = (ar.y - this.state.allMin) * yRatio;
+            const hypotenuse = Math.sqrt(Math.pow(Math.abs(pointX - x), 2) + Math.pow(Math.abs(pointY - y), 2));
+            hypos.push(hypotenuse);
```
MatthewKhouzam, line 591: This is the performance killer right here. instead, save the closest point, not an array. If not, you spend 60% of your time doing a GC per mouse move. I felt it trying out the code on an older computer.

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -564,37 +578,93 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
+        });
+
+        const min = Math.min(...hypos);
+
+        if (min < 20) {
```
MatthewKhouzam, line 596: if you remove the squareroots, min < 400

packages/react-components/src/components/xy-output-component.tsx
MatthewKhouzam: You also really don't need an abs when it comes to squares. a*a  is always positive, if not, I will panic since we're dealing with imaginary points.

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -564,37 +578,93 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
         }
     }
 
-    private tooltip(x: number, y: number): void {
+    private closestPoint(array: any, point: any, xRatio: number, yRatio: number, chartHeight: number): number | undefined {
+        const hypos: any[] = [];
+        // Convert x and y values to screen pixels
+        const pointX = point.x * xRatio;
+        const pointY: number = chartHeight - point.y;
```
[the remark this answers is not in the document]

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -564,37 +578,93 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
+
+        array.forEach((ar: any) => {
+            const x = ar.x * xRatio;
```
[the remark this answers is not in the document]

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -564,37 +578,93 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
+            const y = (ar.y - this.state.allMin) * yRatio;
```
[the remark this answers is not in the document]

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -564,37 +578,93 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
         }
         const chartWidth = this.chartRef.current.chartInstance.width;
+        const chartHeight = this.chartRef.current.chartInstance.height;
         const arraySize = this.state.xyData.labels.length;
         const index = Math.max(Math.round((xPos / chartWidth) * (arraySize - 1)), 0);
+        const start = this.props.viewRange.getStart();
+        const end = this.props.viewRange.getEnd();
+        const xRatio = chartWidth / Number(end - start);
+        const yRatio = chartHeight / (this.state.allMax - this.state.allMin);
+        const chartTop = bounds ? bounds.top : 0;
```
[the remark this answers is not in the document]
Rodrigoplp-work: Indeed. Removed.

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -564,37 +578,93 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
+                    yValue = d.data[closest].y;
+                    xValue = d.data[closest].x;
+                }
+                else {
+                    invalidPoint = true; // Too far from mouse
+                }
             }
             else {
-                zeros += 1;
+                // In case there are less data points than pixels in the chart,
```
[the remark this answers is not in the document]
Rodrigoplp-work: It may require duplicating some code. To be considered in a future ticket.

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -564,37 +578,93 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
         }
     }
 
-    private tooltip(x: number, y: number): void {
+    private closestPoint(array: any, point: any, xRatio: number, yRatio: number, chartHeight: number): number | undefined {
```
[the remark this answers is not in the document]

packages/react-components/src/components/xy-output-component.tsx
Rodrigoplp-work: Yes, thank you for the help there.

packages/react-components/src/components/xy-output-component.tsx
Rodrigoplp-work: Yes, thank you for the help there.

packages/react-components/src/components/xy-output-component.tsx
Rodrigoplp-work: Yes, thank you for the help there.

packages/react-components/src/components/xy-output-component.tsx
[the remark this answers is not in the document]
PatrickTasse: We end up here if the series has no points in the current view range (d.data.length == 0) and then the point appears in the tooltip with the time equal to start time and value 0.

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -572,29 +616,55 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
+                // In case there are less data points than pixels in the chart,
+                // calculate nearest value.
+                yValue = d.data[index];
+            }
+            const rounded: number = isNaN(yValue) ? 0 : (Math.round(Number(yValue) * 100) / 100);
+            let formatted: string = new Intl.NumberFormat().format(rounded);
+            if (this.isScatterPlot && this.props.unitController.numberTranslator) {
+                const time = this.props.unitController.numberTranslator(BigInt(xValue));
+                formatted = '(' + time + ') ' + formatted;
+                timeLabel = 'Process (time stamp) value';
```
PatrickTasse, line 650: It is not a given that the series label is a Process, we don't know what it represents, it depends on the analysis. I wonder if we really need a label in this case anyways?

Also, we might want to ask UX experts (now or later) what is the best format to display in the tooltip:
series (time) value
series value (time)
series: value @ time
something else? and should the formatted string be aligned (table) when there are multiple rows?

PatrickTasse (changes requested):

packages/react-components/src/components/xy-output-component.tsx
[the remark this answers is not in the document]
Rodrigoplp-work: The scatter plot never ends here, as this if-clause checks for `this.isScatterPlot` in line 629. This part of the code works for the line charts, in which the tooltip displays series with value 0.

packages/react-components/src/components/xy-output-component.tsx
Rodrigoplp-work: Changed "Process" to "Series".
Yes, we can change the format of the text as soon as the UX experts choose one.

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -564,37 +578,93 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
             timeLabel = this.props.unitController.numberTranslator(timeForX);
         }
         const chartWidth = this.chartRef.current.chartInstance.width;
+        const chartHeight = this.chartRef.current.chartInstance.height;
         const arraySize = this.state.xyData.labels.length;
         const index = Math.max(Math.round((xPos / chartWidth) * (arraySize - 1)), 0);
+        const start = this.props.viewRange.getStart();
+        const end = this.props.viewRange.getEnd();
+        const xRatio = chartWidth / Number(end - start);
+        const yRatio = chartHeight / (this.state.allMax - this.state.allMin);
```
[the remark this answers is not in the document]

PatrickTasse (changes requested):

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -564,6 +579,35 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
         }
     }
 
+    private closestPoint(array: { x: number, y: number}[], chartWidth: number, chartHeight: number): any | undefined {
```
MatthewKhouzam, line 582: points?

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -564,6 +579,35 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
+        let min_hypotenuse = Number.MAX_VALUE;
+        let closestPoint = undefined;
+        const offset = this.props.viewRange.getOffset() ?? BigInt(0);
+        const start = this.props.viewRange.getStart();
+        const end = this.props.viewRange.getEnd();
+        const xRatio = chartWidth / Number(end - start);
+        const yRatio = (chartHeight - this.margin.top - this.margin.bottom) / (this.state.allMax - this.state.allMin);
+
+        array.forEach((ar: {x: number, y: number}) => {
```
MatthewKhouzam, line 591: point?

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -564,6 +579,35 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
+            const distY = chartHeight - this.positionYMove - y;
+            const hypotenuse = distX * distX + distY * distY;
+
+            if (min_hypotenuse > hypotenuse){
+                closestPoint = ar;
+                min_hypotenuse = hypotenuse;
+            }
+        });
+
+        if (min_hypotenuse < 400) {
```
MatthewKhouzam, line 604: comment explaining 20 pixels in a circle, so max 14 x and 14 y?

MatthewKhouzam (changes requested):
Just comments and nits

packages/react-components/src/components/xy-output-component.tsx
MatthewKhouzam: To be clear, for future devs need to modify this, explain why 400 is picked.

packages/react-components/src/components/xy-output-component.tsx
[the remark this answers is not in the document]
Rodrigoplp-work: It's true. Fixed the if-clause.

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -572,29 +616,55 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
+            let xValue = 0;
+            let invalidPoint = false;
+            if (this.isScatterPlot && d.data.length > 0) {
+                // Find the data point that is the closest to the mouse position
+                const closest = this.closestPoint(d.data, chartWidth, chartHeight);
+                if (closest !== undefined) {
+                    yValue = closest.y;
+                    xValue = closest.x;
+                }
+                else {
```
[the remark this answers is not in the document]

packages/react-components/src/components/xy-output-component.tsx

packages/react-components/src/components/xy-output-component.tsx

packages/react-components/src/components/xy-output-component.tsx
Rodrigoplp-work: Done.

MatthewKhouzam (approved):
Works good enough for me. I think a transitive approval is easy here. ;)

PatrickTasse (approved):

PR #632, opened 2022-01-24
Fix event table time synchronization

PatrickTasse:
Setting the outputStatus state before the tableColumns made the output
render the emptyResultsMessage temporarily, which destroyed the table
created at construction. This caused the columnApi to fail to return all
columns and left the timestamp column undefined.

Set both state variables at the same time in the init() method.


Rodrigoplp-work (approved):

bhufmann (approved):
Thanks for fixing this regression. Looks good to me.

PR #638, opened 2022-01-27
Prepare virtual environment for Python. Initialize ADRs-

marco-miller:
This grown (yet mostly simple) PR can be reviewed on a per-commit basis if needed:
1. the README and git-ignore commits are now pushed through #670, for faster merging;
2. the python commits should be simple enough;
3. the GraphQL ADR commit was already reviewed in Gerrit;
4. the Mermaid (near-top) commit might require slightly more review time;
5. the Apollo (top) commit shouldn't contain any more risks.


bhufmann:
A PR was just merged that updated the README. Could you please rebase this PR? Thanks.

marco-miller:
> > A PR was just merged that updated the README. Could you please rebase this PR? Thanks.
> 
> Sure, will work on this by next Monday or so; thanks.

Done.

bhufmann (changes requested):
The commits upto SHA 3a7653e look good to me and can could be merged. For subsequent changes I have some basic comments. I need some more time for the last one (Apollo)

I think it would be a good idea to start merging parts of it otherwise it will be harder and harder to continue.  Maybe we could merge up-to 3a7653e in a separate PR.

marco-miller:
> The commits upto SHA [3a7653e](https://github.com/theia-ide/theia-trace-extension/commit/3a7653e7cbe8ee202a032acec12d9f1341dbfeff) look good to me and can could be merged. For subsequent changes I have some basic comments. I need some more time for the last one (Apollo)

Ack; thanks.

> I think it would be a good idea to start merging parts of it otherwise it will be harder and harder to continue. Maybe we could merge up-to [3a7653e](https://github.com/theia-ide/theia-trace-extension/commit/3a7653e7cbe8ee202a032acec12d9f1341dbfeff) in a separate PR.

Sure; done through #670.

README.md
```diff
@@ -334,3 +334,14 @@ The following command prints a coverage report to the terminal. As of now, it co
 ```bash
 yarn test --coverage --collectCoverageFrom='src/**/*.ts'
 ```
+
```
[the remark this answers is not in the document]
marco-miller: > The Python virtual environment is only needed when updating/adding ADRs. It's not mandatory. Please clarify this in the README.

Done. I ended up squashing the two related commits together, while inverting the ADR and Python sections order; former now refers to the latter which follows it.

> Should the main README just have a short chapter about ADRs and then reference to a README under doc directory? Just a thought.

The ADR steps need to stay at root level, given that required location (relative to `./doc/adr`) for the `adr-` commands.

doc/adr/0002/graphql-server-example/package.json
```diff
@@ -0,0 +1,16 @@
...
+  "name": "graphql-server-example",
+  "version": "1.0.0",
+  "description": "",
+  "main": "index.js",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "keywords": [],
+  "author": "",
+  "license": "ISC",
```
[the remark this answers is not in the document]
marco-miller: > The theia-trace-extension is currently MIT. We are planning to move to the Eclipse foundation and re-license to EPL-2.0.

Ack.

> Is it possible license the example code under MIT as well?

Done; also added a short commit message paragraph about this.

README.md
```diff
@@ -342,6 +342,21 @@ To initialize a local virtual environment, type the following commands in the ro
...
+pip install -r requirements.txt
 ````
 
 The virtual environment can be replaced with another setup locally.
+
+### About ADRs
+
+ADRs are [Architectural Decision Records][adr].
+
+* The `./adr` directory was initialized using `adr-init` based on [adr-tools-python][tools].
```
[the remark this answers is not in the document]
marco-miller: > the directory is `doc/adr`. Maybe state, that ADRs stored under `doc/adr` directory.

Oops, sure; thanks! -Done.

> In which directory do we need to run the ADR python commands? In the root directory or doc directory? Please clarify. Should this chapter be in a README.md in the doc directory and here just referenced?

Root directory; now clarified in README. Thanks again.

.github/workflows/render-md-mermaid.yml
```diff
@@ -0,0 +1,16 @@
+# https://github.com/nielsvaneck/render-md-mermaid#github-action--workflow-example
+name: render-md-mermaid
```
bhufmann, line 2: will this action be executed when pushing it on forked repos? The publish-next.yml prevents running this action when it's a fork.

.github/workflows/render-md-mermaid.yml
```diff
@@ -0,0 +1,16 @@
...
+jobs:
+  render-md-mermaid:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+      - name: Render images for every Mermaid diagram in all Markdown files (*.md) in the repo
+        uses: nielsvaneck/render-md-mermaid@v2
+      - name: Commit rendered png and svg files
+        uses: stefanzweifel/git-auto-commit-action@v4
```
bhufmann, line 13: Are only new and updated images committed? 
If the mermaid source doesn't change there should not be any new commits or uncommitted files in the repo after running this github action. Uncommitted files, would make the publishing fail.

How does it detect, that a new version of the file needs to be created when the mermaid source changes? Or does it blindly create an image every time until the created image is not the same anymore.

Another question, GitHub is caching images referenced in Markdown files. Change the image file in the repo won't clear the cache and the old image is shown. That's why in doc/images there a files with version numbers to make rendering work.

doc/adr/0002-GraphQL.md
```diff
@@ -0,0 +1,317 @@
...
+    GraphQL
+  end
+  subgraph Servers
+    trace-server
+    GraphQL-Server
+  end
+```
+
+1. GraphQL would be offered as a TSP option alongside current REST endpoints.
+1. Resolver functions in GraphQL Server would rely on current REST endpoints.
```
bhufmann, line 196: Is it required that the resolver translate the GraphQL queries to REST API calls that are currently implemented in the trace server? If I don't misunderstand it, that would add 1 serialization and 1 deserialization to the query execution path, which will increase the overall round-trip for a query.

doc/adr/0002/graphql-server-example/index.js
```diff
@@ -0,0 +1,51 @@
+// https://www.apollographql.com/docs/apollo-server/getting-started/#step-3-define-your-graphql-schema
```
bhufmann, line 1: Do we need to provide the example code here or could we just have a reference to this example in the GraphQL ADR? It's to demonstrate it with a simple example and it might at some maintenance tasks.

bhufmann (changes requested):
The first commit looks good to me. For the others I have some comments.

.github/workflows/render-md-mermaid.yml
marco-miller: Good point, so this action would have needed that guard too. Thanks for the flag. Now I'm about to remove this whole action file anyway, will push that next.

.github/workflows/render-md-mermaid.yml
marco-miller: > Are only new and updated images committed? If the mermaid source doesn't change there should not be any new commits or uncommitted files in the repo after running this github action. Uncommitted files, would make the publishing fail.

Agreed, and these were my expectations too while cooking that action; will remove it all very soon though.

> How does it detect, that a new version of the file needs to be created when the mermaid source changes? Or does it blindly create an image every time until the created image is not the same anymore.

Action itself was yet to be tested (in action), but the test script mentioned in this original commit message dirtied the image files only when their mermaid source code changed; AFAIK that was.

> Another question, GitHub is caching images referenced in Markdown files. Change the image file in the repo won't clear the cache and the old image is shown. That's why in doc/images there a files with version numbers to make rendering work.

Good point, which mainly leads to me pushing another approach next.

doc/adr/0002-GraphQL.md
marco-miller: Yes only for the time being, and totally right; indeed. As previously shared, we will still be able to reconsider that approach eventually or at will. -Just not yet or for this initial ADR version of ours. (Or so was the proposal so far.)

doc/adr/0002/graphql-server-example/index.js
marco-miller: Agreed. -Will push an amend, accordingly; thanks.

doc/adr/0002-GraphQL.md
bhufmann: Thanks for clarifying.

doc/adr/0002/graphql-server-example/index.js
bhufmann: Thanks

bhufmann (approved):
It looks good me to merge this investigation increment. Thanks.

Issue #634, opened 2022-01-26
tree component doesn't disable browser context sensitive menu


bhufmann:
The context sensitive menu of the the tree component is not disable so that it shows when the user does a right-mouse click action. 
The browser menu should be disabled. Right now no context sensitive actions are defined for such tree.

The component is used for the following views based on:
- DATA_TREE  (e.g. Futex Contention Analysis - Latency Statistics)
- TREE_TIME_XY (the tree on the left side, e.g. Histogram)
- TIME_GRAPH (the tree on the left side, e.g. Resources View)

![tree-right-mouse-bug](https://user-images.githubusercontent.com/413424/151177146-cce428cf-dbd4-453b-8f83-54c4686ec10a.gif)

Issue #635, opened 2022-01-26
Browser context sensitive menu is not disabled when not showing graphs


bhufmann:
When the empty chart message is shown, the context sensitive menu of the is not disable so that it shows when the user does a right-mouse click action. The browser menu should be disabled. This is true for any of the chart types, e.g. xy-chart and for empty chart message, error message etc.

Example screenshots:

![empty-trace-right-mouse-bug](https://user-images.githubusercontent.com/413424/151181076-dd2360ab-787d-45b1-aa13-a783f05e2a8d.gif)

![xy-chart-right-mouse-bug](https://user-images.githubusercontent.com/413424/151179544-c906d812-e4ce-4744-8575-8095ed56e1a4.gif)

bhufmann:
fixed by #667

PR #667, opened 2022-02-02
Remove context menu from TraceContextComponent

williamsyang-work:
Disabled the browser's context-menu for the entire TraceContextComponent, which effects the area highlighted in blue:

PreventDefaultContextMenu

There are still areas in the extension where we can access the browser's context-menu: 

- Empty space under "Opened Traces"
- Empty space under "Item Properties"
- Empty space next to Trace Tabs (near top)

If we want the right-click disabled there too, we can move this to a higher-level component.

Fixes: #634, #635

Signed-off-by: William Yang william.yang@ericsson.com

### DIFF
--- a/packages/react-components/src/components/trace-context-component.tsx
+++ b/packages/react-components/src/components/trace-context-component.tsx
@@ -293,8 +293,13 @@ export class TraceContextComponent extends React.Component<TraceContextProps, Tr
         }));
     }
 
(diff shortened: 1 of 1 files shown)


Rodrigoplp-work (approved):

bhufmann (approved):
Looks good to me. Thanks.

PR #671, opened 2022-02-04
Clarify when the Trace Server is Offline

### LINKED ISSUE #321: Clarify that Trace Viewer not functional when server offline
When the server is offline, this information need to be communicated in a permanent and obvious way:

| Critical information | Solution ideas |
|----|----|
| **The tool cannot be used ...**             |   Make the tool look and feel broken/unresponsive: Prevent interactions (ex. opening analyses), Fade out the associated UI (ex. Trace viewer side panel, trace tabs)                     |
| **... because the trace server is offline**              |   Write a more precise permanent message like "Trace Viewer Critical Error: Trace server offline"                     |
| Bonus: Point the user towards how they can fix the problem | Should they restart the server? Change the configuration? |

Related: [#53 comment](https://github.com/theia-ide/theia-trace-extension/issues/53#issuecomment-782396960), #234

----

~~Right now, when the trace server is offline the only hint to users is the bottom bar saying "Offline".~~ PR #298 removed the server start/stop buttons from the bottom bar.

image

This is not clear enough considering that the **Trace Viewer is unusable if the trace server is offline**. 

| When the server is offline, it needs to be obvious to users that: | Neither of these things are clear in the current implementation: |
|----|----|
| **The tool cannot be used** - This is a critical failure              |   Right now, users can continue to attempt to add views (or interact with current views) which will never successfully update                     |
| **Why the tool cannot be used** - The trace server is offline              |   It's virtually impossible to guess that "Offline" means "The trace server is offline"                     |

----

CC: @ssmagula

hriday-panchasara:
Screen Shot 2022-02-01 at 11 43 17 AM
Screen Shot 2022-02-01 at 11 43 57 AM

fixes #321 

### DIFF
--- a/packages/react-components/style/trace-explorer.css
+++ b/packages/react-components/style/trace-explorer.css
@@ -285,3 +285,8 @@
     align-items: center;
     justify-content: flex-end !important;
 }
+
+#trace\.viewer\.serverCheck {
+    cursor: default;
+    color: green;
+}
--- a/theia-extensions/viewer-prototype/src/browser/trace-explorer/trace-explorer-widget.tsx
+++ b/theia-extensions/viewer-prototype/src/browser/trace-explorer/trace-explorer-widget.tsx
@@ -6,6 +6,7 @@ import { TraceExplorerOpenedTracesWidget } from './trace-explorer-sub-widgets/th
 import { TraceExplorerPlaceholderWidget } from './trace-explorer-sub-widgets/trace-explorer-placeholder-widget';
 import { signalManager, Signals } from 'traceviewer-base/lib/signals/signal-manager';
 import { OpenedTracesUpdatedSignalPayload } from 'traceviewer-base/src/signals/opened-traces-updated-signal-payload';
+import { TraceServerConnectionStatusService } from '../trace-server-status';
 
 @injectable()
 export class TraceExplorerWidget extends BaseWidget {
@@ -18,6 +19,7 @@ export class TraceExplorerWidget extends BaseWidget {
     @inject(TraceExplorerTooltipWidget) protected readonly tooltipWidget!: TraceExplorerTooltipWidget;
     @inject(TraceExplorerPlaceholderWidget) protected readonly placeholderWidget!: TraceExplorerPlaceholderWidget;
     @inject(ViewContainer.Factory) protected readonly viewContainerFactory!: ViewContainer.Factory;
+    @inject(TraceServerConnectionStatusService) protected readonly connectionStatusService: TraceServerConnectionStatusService;
 
     openExperiment(traceUUID: string): void {
         return this.openedTracesWidget.openExperiment(traceUUID);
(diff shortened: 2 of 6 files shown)


Rodrigoplp-work:
Good work!

Do we really need a "keep alive" for this? It generates a high traffic between the front-end and the server.

Wouldn't it be good enough to show the red icon only after the server fails to respond? It could be done by catching an error in `tsp-typescript-client/sac/protocol/rest-client.ts`.

Then we could start the periodic (5 seconds) calls until the server is back online.

This way we only make the `/health` requests if the server is off-line.

hriday-panchasara:
> Good work!
> 
> Do we really need a "keep alive" for this? It generates a high traffic between the front-end and the server.
> 
> Wouldn't it be good enough to show the red icon only after the server fails to respond? It could be done by catching an error in `tsp-typescript-client/sac/protocol/rest-client.ts`.
> 
> Then we could start the periodic (5 seconds) calls until the server is back online.
> 
> This way we only make the `/health` requests if the server is off-line.

That's a good idea! @bhufmann what do you think of this approach?

Rodrigoplp-work:
> > > Good work!
> > > Do we really need a "keep alive" for this? It generates a high traffic between the front-end and the server.
> > > Wouldn't it be good enough to show the red icon only after the server fails to respond? It could be done by catching an error in `tsp-typescript-client/sac/protocol/rest-client.ts`.
> > > Then we could start the periodic (5 seconds) calls until the server is back online.
> > > This way we only make the `/health` requests if the server is off-line.
> > 
> > 
> > That's a good idea! @bhufmann what do you think of this approach? I think we might've discussed something similar while brainstorming
> 
> Yes, that sounds like good idea and it will avoid sending unnecessary requests. One thing though, would want to limit the time how long (or how often) the health check is sent to detect if the server is back?

I agree. Maybe something like this, if possible:
- Query every 5s for 1 minute;
- After that, every 30s for 4 minutes;
- Stop.

hriday-panchasara:
@Rodrigoplp-work Once we catch the error in tsp-typescript-client how do we report back to the trace-viewer? In order to change the status LED we can either use DOM manipulation, fire a signal, or add a custom action to Theia. But all these approaches would have to be performed in the trace-extension repo and can’t be done through typescript-client

Rodrigoplp-work:
> @Rodrigoplp-work Once we catch the error in tsp-typescript-client how do we report back to the trace-viewer? In order to change the status LED we can either use DOM manipulation, fire a signal, or add a custom action to Theia. But all these approaches would have to be performed in the trace-extension repo and can’t be done through typescript-client

Good point, Hriday. I think the right way to do this would have to start in `rest-client.js` in the tsp-typescript-client project. 

Today, if the server is down the fetch from `const response = await fetch(url, { method, headers, body });` in `httpRequest()` simply fails. It would need to have a `try... catch` around it, so that it could return a 503 or a response containing `{success: false, statusText: "Server down"}` to the front-end.

Then, in theia-trace-extension, we would need to detect that 503 and trigger the red icon. But I'm not sure there is a single point where we could do that. It may be necessary to do it everywhere we use `tsp-typescript-client/lib/protocol/tsp-client` (it's being called from 10 places now, like in `fetchAnnotationCategories()` in trace-viewer.tsx).

hriday-panchasara:
@PatrickTasse @Rodrigoplp-work updated the branch to use ConnectionStatusListener from tsp-typescript-client

bhufmann (approved):
Thanks!

PR #675, opened 2022-02-09
Split X-axis views and non-X-axis views

### LINKED ISSUE #287: Why a single axis for all time-based charts?
Having only one axis for time-based charts has widespread implications for UX. Because of this, I'd like to know the design criteria that lead to this decision.

* **Motivation for single time axis?** To maximize vertical space by avoiding repeating redundant UI elements?
* **How to integrate views that aren't time-based?** There is precedent for a single time axis when all data is time-based (see Google's Perfetto below), but because not all our views are time-based (and can be in mixed order with time-based views) this is a UX pain that needs careful consideration.

Example: Perfetto uses a single time axis
image

### LINKED ISSUE #618: Not possible to close single Events Table
If the Events Table is the only open view, its 'X' button does not work so the view can't be closed.

https://user-images.githubusercontent.com/28311615/149538529-bf90ac86-b1d5-4113-9d88-583fcf69d0c2.mp4

### LINKED ISSUE #648: Don’t automatically resize manually resized views when a new view is opened
Users were surprised (and some frustrated) when they opened a new views and manually resized views were reset back to the default size.

hriday-panchasara:
Implement different visual containers for views that have an X axis and those that don't

Changes Made:
- View is arranged into charts that use the X-axis (TimeAxis & TimeNavigator components) and those that don’t.
- Sizes and orders of charts are preserved when views are added or removed
- The close button is accessible when only a single view that does not spawn the time-axis is open

Notes:
- I've added some commenting for now to explain my implementation
- I feel that the runtime is a bit high, would like some suggestions on how I could optimize

fixes #287
fixes #648
fixes #618

Screen Shot 2022-02-17 at 11 39 32 PM

### DIFF
--- a/packages/react-components/src/components/trace-context-component.tsx
+++ b/packages/react-components/src/components/trace-context-component.tsx
@@ -26,6 +26,7 @@ import { TooltipComponent } from './tooltip-component';
 import { TooltipXYComponent } from './tooltip-xy-component';
 import { BIMath } from 'timeline-chart/lib/bigint-utils';
 import { DataTreeOutputComponent } from './datatree-output-component';
+import { cloneDeep } from 'lodash';
 
 const ResponsiveGridLayout = WidthProvider(Responsive);
 
@@ -69,6 +70,8 @@ export class TraceContextComponent extends React.Component<TraceContextProps, Tr
     private tooltipComponent: React.RefObject<TooltipComponent>;
     private tooltipXYComponent: React.RefObject<TooltipXYComponent>;
     private traceContextContainer: React.RefObject<HTMLDivElement>;
+    private _storedTimescaleLayout: Layout[] = [];
+    private _storedNonTimescaleLayout: Layout[] = [];
 
     protected widgetResizeHandlers: (() => void)[] = [];
     protected readonly addWidgetResizeHandler = (h: () => void): void => {
@@ -134,6 +137,7 @@ export class TraceContextComponent extends React.Component<TraceContextProps, Tr
         this.traceContextContainer = React.createRef();
         this.onResize = this.onResize.bind(this);
         this.setChartOffset = this.setChartOffset.bind(this);
+        this.onLayoutChange = this.onLayoutChange.bind(this);
         this.props.addResizeHandler(this.onResize);
         this.initialize();
         this.subscribeToEvents();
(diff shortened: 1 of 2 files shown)


PatrickTasse:
At first glance it does not behave how I would have expected, but we should consider its pros and cons.

The cons I see with two containers is that it creates two scrollable areas, so once they are created a user cannot see two initial-sized graphs at the same time just by scrolling, the sash needs to be resized. When there are a lot of outputs open, to find a specific one that is hidden, the user has to know to scroll the right area. Also the horizontal scroll bar moves around as outputs are created and removed, and could be hard to find.

What I expected was simply that the order of outputs would be restricted so that timescale outputs are on top and non-timescale outputs on the bottom. But the con of that is that the horizontal scroll bar is at the bottom of the tab where it could be far from the timescale outputs. It is like that in the original implementation, but I am not sure if it was a goal to change that.

A useful new feature is that it is now possible to fix an output while the other outputs can be scrolled. But in this implementation, the outputs that can be scrolled are limited to the outputs of the 'other' (timescale/non-timescale) category. You can't fix a graph while you scroll through other graphs, which would be a nice improvement.

An idea I had to achieve that was to add a pin feature to the output components. This could be a new button added to the output handle. "Pin to Top" for timescale outputs, "Pin to Bottom" for non-timescale outputs. The pinned outputs would then be moved out of the scrollable area, but they would keep their resize handle. Then all the non-pinned outputs would remain part of a scrollable area, as today.

Rodrigoplp-work:
Thank you for calling a meeting about this, Hriday. My perspective from that discussion is to have, if possible:

- Charts and tables get added like we currently have on the master branch, but time-line charts are automatically moved to be on the top, tables at the bottom;
- The horizontal bar that indicates zoom appears below the last time-line chart, above the first table;
- Time-line charts and tables can be moved up and down but, after the user ends dragging, the tables move to be below the time-line charts.

Even better if this could be achieved without adding the react-split library.

I like this solution because it would reduce the amount of scroll bars on screen, remove the white blank spaces created by the "split" solution, and eliminate the problem of hidden charts in the middle of the screen.

Rodrigoplp-work:
Amazing! Just like we discussed. Thank you Hriday.

Rodrigoplp-work (dismissed):

PatrickTasse (changes requested):
I think the timeNavigator should always be visible. So if there are more time scale outputs than fits in the window height, the timeNavigator should float and snap to the bottom of the window.

hriday-panchasara:
> I think the timeNavigator should always be visible. So if there are more time scale outputs than fits in the window height, the timeNavigator should float and snap to the bottom of the window.

@PatrickTasse I tired but I was unable to stick the timeNavigator on overflow.

To make the timeNavigator 'fixed' when its scrolled out of view, we can attach the following class to the component:
Screen Shot 2022-02-23 at 13 52 17 PM

But this would only work when the parent component does not have an overflow value. So removing 'overflow-y: scroll' from the RGL container would work but it would mean that our view is not scrollable anymore. Also, to make 'position' property work for RGL children we have to add the following prop to the RGL container: 'useCSSTransforms: false'. But by doing so the charts become static and I'm unable to interact with them. Looking into this css issue the recommended fix I found was to put the 'sticky' element in a separate container. By doing so the TimeNavigator won't be a part of the resizable Grid, hence we wouldn't have a visible divider between the two types of charts:
Screen Shot 2022-02-23 at 14 02 33 PM

hriday-panchasara:
@PatrickTasse @Rodrigoplp-work updated the PR to do all of the above as well as keep the timeNavigator visible on overflow:
Screen Shot 2022-02-25 at 09 19 26 AM

PatrickTasse (approved):

PR #677, opened 2022-02-09
GraphQL ADR: prototype resolvers for tsp-client.ts

marco-miller:
This PR is described in its commit message.

### DIFF
--- a/doc/adr/0002/graphql-server-prototype/test
+++ b/doc/adr/0002/graphql-server-prototype/test
@@ -1,0 +1,16 @@
+#!/bin/bash
+# Assumes [1] below, which assumes a running trace-server; based on [2].
+# POST is used for the curl request, but GET [3] can be used although less readable.
+# Any combinations of the supported fields can be queried, which is why GraphQL.
+# This example is a simple one that barely shows how far such queries can get.
+#
+# [1] https://www.apollographql.com/docs/apollo-server/getting-started/#step-7-start-the-server
+# [2] https://www.apollographql.com/docs/apollo-server/getting-started/#step-8-execute-your-first-query
+# [3] https://www.apollographql.com/docs/apollo-server/requests/#get-requests
+
+curl --request POST \
+  --header 'content-type: application/json' \
+  --url http://localhost:4000/ \
+  --data '{"query": "{status, traces}"}'
+ #--data '{"query": "{status}"}'
+ #--data '{"query": "{traces}"}'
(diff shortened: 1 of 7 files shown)


bhufmann (commented):
When looking at the proposal, I understand that the proposed GraphQL server, that is implemented using Apollo server, translates GraphQL queries into TSP API REST queries towards the trace server.

My understanding of the GraphQL for our project was, that the trace server application would be the GraphQL server, serving GraphQL query requests, and on the client side (tsp-typescript-client) a GraphQL client would implemented, that connects to the GraphQL server backend (though HTTP). So, the tsp-typescript-client (and tsp-python-client respectively) would implement the client side, maybe using Apollo Client.

Or do I misunderstand something?

marco-miller:
> When looking at the proposal, I understand that the proposed GraphQL server, that is implemented using Apollo server, translates GraphQL queries into TSP API REST queries towards the trace server.

As an initial or first proposal, yes.

> My understanding of the GraphQL for our project was, that the trace server application would be the GraphQL server, serving GraphQL query requests, and on the client side (tsp-typescript-client) a GraphQL client would implemented, that connects to the GraphQL server backend (though HTTP). So, the tsp-typescript-client (and tsp-python-client respectively) would implement the client side, maybe using Apollo Client.

Apollo Client is not necessary yet, but yes, the existing tsp-*-clients would send graph requests towards the GraphQL server, either directly or through an app using them such as this extension repo. Now, we may decide to make the GraphQL server rely directly on trace-server internals, but this proposal brings the idea of proving clients more quickly, by means of taping onto TSP REST as initial iterations.

doc/adr/0002/graphql-server-prototype/test
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+# Assumes [1] below, which assumes a running trace-server; based on [2].
+# POST is used for the curl request, but GET [3] can be used although less readable.
+# Any combinations of the supported fields can be queried, which is why GraphQL.
```
Rodrigoplp-work, line 4: Did something get cut out from the end of the sentence? Was it meant to be something like "... which is why GraphQL was suggested as a solution."?

Rodrigoplp-work (changes requested):

doc/adr/0002/graphql-server-prototype/test
marco-miller: No, not really; this way of saying does exist in English, AFAIK.

Rodrigoplp-work (approved):

PR #684, opened 2022-02-15
Adr d3

Rodrigoplp-work:
This ADR considers the benefits and difficulties of adopting D3 to
render the charts of Theia Trace Extension, in comparison with the
current solution that uses ChartJS.

The ADR contains an example of implementation in the file
`xy-output-component.tsx`.


bhufmann (changes requested):
This PR includes multiple unrelated commits that have been merged to master branch. Could you please rebase your PR so that it only has your D3 ADR commit?

PR #687, opened 2022-02-22
Change cursor on context clicks on XY-charts

### LINKED ISSUE #636: Change cursor when doing right-mouse drag zoom in XY-charts
In comparison to the Time Graph charts, the cursor is not changed  in xy-charts when the user is doing zoom operation using `right-mouse click drag`.

missing-right-mouse-drag-cursor

Rodrigoplp-work:
Change the cursor when dragging with the left button (crosshair cursor)
and with the right button (ew-resize cursor).

The behavior of the xy-charts now is the same as on time-graph charts.

Fixes #636.

### DIFF
--- a/packages/react-components/src/components/xy-output-component.tsx
+++ b/packages/react-components/src/components/xy-output-component.tsx
@@ -30,6 +30,8 @@ type XYOuputState = AbstractOutputState & {
     columns: ColumnHeader[];
     allMax: number;
     allMin: number;
+    cursor?: string;
+    shiftKey: boolean;
 };
 const RIGHT_CLICK_NUMBER = 2;
 const ZOOM_IN = true;
@@ -71,6 +73,9 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
            this.applySelectionZoom();
         }
         this.mouseIsDown = false;
+        if (!this.state.shiftKey) {
+            this.setState({ cursor: 'default'});
+        }
         document.removeEventListener('mouseup', this.endSelection);
     };
 
@@ -90,7 +95,9 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
             xyData: {},
             columns: [{title: 'Name', sortable: true}],
             allMax: 0,
-            allMin: 0
+            allMin: 0,
+            cursor: 'default',
+            shiftKey: false
         };
 
         this.afterChartDraw = this.afterChartDraw.bind(this);
@@ -329,12 +336,13 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
                     className='xy-main'
                     tabIndex={0}
                     onKeyDown={event => this.onKeyDown(event)}
+                    onKeyUp={event => this.onKeyUp(event)}
                     onWheel={event => this.onWheel(event)}
                     onMouseMove={event => this.onMouseMove(event)}
                     onContextMenu={event => event.preventDefault()}
                     onMouseLeave={event => this.onMouseLeave(event)}
                     onMouseDown={event => this.onMouseDown(event)}
-                    style={{ height: this.props.style.height, position: 'relative' }}
+                    style={{ height: this.props.style.height, position: 'relative', cursor: this.state.cursor }}
                     ref={this.divRef}
                 >
                     {this.chooseChart()}
@@ -476,9 +484,11 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
         const startTime = this.getTimeForX(event.nativeEvent.offsetX);
         if (event.button === RIGHT_CLICK_NUMBER) {
             this.isRightClick = true;
+            this.setState({ cursor: 'col-resize' });
             this.startPositionMouseRightClick = startTime;
         } else {
             this.isRightClick = false;
+            this.setState({ cursor: 'crosshair' });
             if (event.shiftKey && this.props.unitController.selectionRange) {
                 this.props.unitController.selectionRange = {
                     start: this.props.unitController.selectionRange.start,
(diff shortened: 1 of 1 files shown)


PatrickTasse (changes requested):
Not sure if these comments are in the scope of this PR or if they need a new PR:

There are some small differences in behavior compared with time-graph charts:

- If you start dragging in one mode (zoom or select) then click the other button, it starts a new action with the new mode. The other buttons should be ignored while an action is already in progress.
- In time-graph chart, holding the Shift button will change to the crosshair cursor to show that a click will extend the selection instead of starting a new selection. Note that pressing any other modifier key will cancel this mode.

Rodrigoplp-work:
- Shift button: implemented;
- Avoid click while dragging: I made a few attempts but didn't succeed. The problem is not only on the xy-chart, it also happens on others, like in the Resources Status chart. If we really want to do this, I suggest we open a new ticket to address it in the future.

PatrickTasse (changes requested):
If you hold Shift, then click to extend selection, if the Shift is still being pressed then the cursor should still be the crosshair.

About the click while dragging, I'm not sure if it's the same use case that was tried because it doesn't happen with time graphs:
- Left mouse down (starts time selection)
- Mouse move (extends time selection)
- Right click (ignored, nothing happens) <- but xy-chart different
- Mouse move (continues to extend time selection)
- Left mouse up (finishes time selection)
And same thing when swapping left and right buttons.
But it's OK to address this separately, as it's not only related to the cursor but to the action handling itself also.

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -797,10 +803,20 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
+                    this.setState({ cursor: 'crosshair' });
+                    break;
+                }
             }
         }
     }
 
+    private onKeyUp(key: React.KeyboardEvent) {
+        if (key.key === 'Shift') {
+            this.setState({ cursor: 'default' });
```
PatrickTasse, line 816: It should not reset the cursor is selection is in progress with the left mouse button being pressed.

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -797,10 +803,20 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
                     this.pan(PAN_RIGHT);
                     break;
                 }
+                case 'Shift': {
+                    this.setState({ cursor: 'crosshair' });
```
PatrickTasse, line 807: It should not change the cursor if zoom selection is in progress with the right mouse button being pressed.

packages/react-components/src/components/xy-output-component.tsx

packages/react-components/src/components/xy-output-component.tsx

PatrickTasse (dismissed):

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -71,6 +73,9 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
            this.applySelectionZoom();
         }
         this.mouseIsDown = false;
+        if (!this.state.shiftKey) {
+            this.setState({ cursor: 'default '});
```
marco-miller, line 77: Extra space right after "default".

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -90,7 +95,9 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
             xyData: {},
             columns: [{title: 'Name', sortable: true}],
             allMax: 0,
-            allMin: 0
+            allMin: 0,
+            cursor: 'default',
```
marco-miller, line 99: Should a constant be defined and reused for this repeating 'default' value?

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -797,7 +807,23 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
                     this.pan(PAN_RIGHT);
                     break;
                 }
+                case 'Shift': {
+                    if (!this.isRightClick) {
+                        this.setState({ cursor: 'crosshair', shiftKey: true });
```
marco-miller, line 812: Same question for this other repeated value.

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -797,7 +807,23 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
+                    }
+                    break;
+                }
+            }
+        }
+    }
+
+    private onKeyUp(key: React.KeyboardEvent) {
+        if (key.key === 'Shift') {
```
marco-miller, line 821: Same question again, here for this key string value. Not sure if such duplication is done elsewhere or treated as normal (or expected).

marco-miller (changes requested):

packages/react-components/src/components/xy-output-component.tsx
Rodrigoplp-work: Fixed.

packages/react-components/src/components/xy-output-component.tsx

packages/react-components/src/components/xy-output-component.tsx
Rodrigoplp-work: I would rather leave as is, as explained above.

packages/react-components/src/components/xy-output-component.tsx
marco-miller: How about CURSOR_DEFAULT and so on, or one constant per reused cursor string value?

marco-miller (changes requested):
Another proposal in case it could fit in; thanks.

PatrickTasse:
I can understand Marco's point coming from Java where pretty much every code literal should be defined as a constant, as a good coding practice, especially when it is a fixed string defined in an API. If this was a non-sucking language it would provide API-defined constants for these CSS strings so that the client could use them directly...

But it's not just cursor constants in the same situation, there's also mouse event types 'mouseup', keyboard event key 'Shift', etc. We seem to be using literals for these all over the project (and other HTML projects?), so I'm not sure if we should start doing it differently here?

I'm fine either way.

PatrickTasse (approved):

marco-miller (approved):

PR #693, opened 2022-02-28
Support displaying statistics for a region selection in the chart

hriday-panchasara:
Show statistics for selected region in datatree-output-component

In the datatree-output-component, add a subtree labeled 'Selection' whenever there is a region selected. When no region is selected, hide the row:
Screen Shot 2022-02-28 at 10 50 26 AM
Screen Shot 2022-02-28 at 10 53 06 AM

### DIFF
--- a/.markdownlint.json
+++ b/.markdownlint.json
@@ -1,0 +1,6 @@
+{
+  "default": true,
+  "MD010": { "code_blocks": false },
+  "MD013": false,
+  "MD033": false
+}


mirsky-work:
Thanks for the contribution!
Tested on Gitpod and it looks like the selection always returns the same values as total - is backend implementation missing?
image

From the UI perspective, there seems to be an issue when starting the range selection from right to left.
https://user-images.githubusercontent.com/67920463/156031627-b3ae7e9b-c0cf-4a9d-9a96-4006c887670f.mp4

PR #694, opened 2022-03-02
Implement statistics for a selected area in the chart

hriday-panchasara:
Show statistics for selected region in datatree-output-component

In the datatree-output-component, a subtree labeled 'Selection' is shown whenever there is a region selected. When no region is selected, hide the row:
Screen Shot 2022-03-02 at 16 50 57 PM
Screen Shot 2022-03-02 at 16 51 27 PM

### DIFF
--- a/packages/react-components/src/components/datatree-output-component.tsx
+++ b/packages/react-components/src/components/datatree-output-component.tsx
@@ -8,6 +8,8 @@ import { EntryTree } from './utils/filter-tree/entry-tree';
 import { getAllExpandedNodeIds } from './utils/filter-tree/utils';
 import { TreeNode } from './utils/filter-tree/tree-node';
 import ColumnHeader from './utils/filter-tree/column-header';
+import { cloneDeep } from 'lodash';
+import debounce from 'lodash.debounce';
 
 type DataTreeOutputProps = AbstractOutputProps & {
 };
@@ -22,6 +24,8 @@ type DataTreeOuputState = AbstractOutputState & {
 
 export class DataTreeOutputComponent extends AbstractOutputComponent<AbstractOutputProps, DataTreeOuputState> {
     treeRef: React.RefObject<HTMLDivElement> = React.createRef();
+
+    private _debouncedFetchSelectionData = debounce(() => this.fetchSelectionData(), 500);
 
     constructor(props: AbstractOutputProps) {
         super(props);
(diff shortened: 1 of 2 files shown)


mirsky-work:
Verified on Gitpod - works great!
Thanks for that very useful contribution!

packages/react-components/src/components/datatree-output-component.tsx
```diff
@@ -154,4 +160,45 @@ export class DataTreeOutputComponent extends AbstractOutputComponent<AbstractOut
...
+                        xyTree: treeResponse.model.entries,
+                    });
+                }
+            }
+        }
+    }
+
+    async componentDidUpdate(prevProps: DataTreeOutputProps): Promise<void> {
+        if (this.props.selectionRange && this.props.selectionRange !== prevProps.selectionRange) {
+            if (Math.abs(Number(this.props.selectionRange.getDuration())) > 0) {
```
PatrickTasse, line 194: If the selection is changed to a single timestamp during the debounce period, the request will be sent with this single selection (start == end). So I initially thought that the duration check should be done at the end, inside the debounced method.

However, the resulting data is not empty statistics for the timestamp selection. After talking with Matthew, the stats include any examined segment that intersects with the selection. In Trace Compass the Selection statistics are shown even with a single timestamp selection. Once a selection has been done once, the Selection statistics will never disappear (because it's not possible for the user to unselect time). So I believe the duration check should just be removed and selection statistics should be fetched from the server regardless of whether the selection is a timestamp or a time range.

packages/react-components/src/components/datatree-output-component.tsx
```diff
@@ -154,4 +160,45 @@ export class DataTreeOutputComponent extends AbstractOutputComponent<AbstractOut
...
+                this._debouncedFetchSelectionData();
+            }
+            else {
+                this.setState({
+                    xyTree: this._totalTree
```
PatrickTasse, line 199: In that case this _totalTree may no longer be needed.

PatrickTasse (changes requested):

packages/react-components/src/components/datatree-output-component.tsx
hriday-panchasara: @PatrickTasse Updated componentDidUpdate() implementation

packages/react-components/src/components/datatree-output-component.tsx
```diff
@@ -60,6 +65,7 @@ export class DataTreeOutputComponent extends AbstractOutputComponent<AbstractOut
                     xyTree: treeResponse.model.entries,
                     columns
                 });
+                this._totalTree = cloneDeep(this.state.xyTree);
```
PatrickTasse, line 68: This field is no longer used and should be removed.

packages/react-components/src/components/datatree-output-component.tsx
```diff
@@ -154,4 +160,38 @@ export class DataTreeOutputComponent extends AbstractOutputComponent<AbstractOut
         // fix Warning: Can't perform a React state update on an unmounted component
         this.setState = (_state, _callback) => undefined;
     }
+
+    protected async fetchSelectionData(): Promise<void> {
+        if (this.props.selectionRange) {
+            const origXYTree = cloneDeep(this.state.xyTree);
+            origXYTree.sort((a,b)=> (a.id > b.id) ? 1 : -1);
```
PatrickTasse, line 167: This local variable is not used.

packages/react-components/src/components/datatree-output-component.tsx
```diff
@@ -154,4 +160,38 @@ export class DataTreeOutputComponent extends AbstractOutputComponent<AbstractOut
...
+
+            let payload: any;
+            if (Number(this.props.selectionRange.getStart()) < Number(this.props.selectionRange.getEnd())) {
```
PatrickTasse, line 170: I think this might give the wrong result if both values are very large and very close, due to the loss of precision. You can compare the bigint values directly?

PatrickTasse (changes requested):

packages/react-components/src/components/datatree-output-component.tsx
hriday-panchasara: > I think this might give the wrong result if both values are very large and very close, due to the loss of precision. You can compare the bigint values directly?

Didn't know that, I'll make the change

PatrickTasse (approved):

Rodrigoplp-work (approved):

PR #695, opened 2022-03-02
Prevent unnecessary xy requests while analysis is running

### LINKED ISSUE #691: Unnecessary xy requests to the backend while analysis is running.
When opening an XY view, the fetchTree() is repeatedly called until the requests returns COMPLETED. If the back-end call returns RUNNING, the a new requests is send after 500 ms. However, the XyOutputComponent.fetchTree() that calls the back-end also updates the component state, the componentDidUpdate() message is called, which triggers the update xy data and hence a back-end call to fetchXY() is performed. This causes many unnecessary back-end fetchXY() calls. The result is not visualized anyways, until the analysis is completed. 

This can be observed, by opening the CPU usage view of a "large" LTTng kernel trace while having the network tab of the chrome developer open.

These unnecessary back-end call causes unnecessary network traffic, back-end and front-end processing.

The fetchXY() calls should only be done when analysis finished or using a debouncer.

Fixing this will increase the responsiveness of the front-end implementation.

PatrickTasse:
In componentDidUpdate(), do not trigger a xy request if the previous
outputStatus was running, only trigger it if the outputStatus changed.
This will trigger it only once, when the analysis is completed.

Fixes #691


bhufmann (approved):
Looks good to me. Thanks for fixing this performance issue.

Rodrigoplp-work (approved):

PR #696, opened 2022-03-03
Bar chart code base for xy charts

Rodrigoplp-work:
Provides the structure to plot bar charts when this type of chart becomes available.

To test it, set

```js
private isBarPlot = true;
```

in `xy-output-component.tsx` file. Then open an xy plot like "CPU usage".

bar-chart

### DIFF
--- a/packages/react-components/src/components/xy-output-component.tsx
+++ b/packages/react-components/src/components/xy-output-component.tsx
@@ -65,8 +65,9 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
     private startPositionMouseRightClick = BigInt(0);
     private margin = { top: 15, right: 0, bottom: 5, left: this.getYAxisWidth() };
     private isScatterPlot: boolean = this.props.outputDescriptor.id.includes('scatter');
+    private isBarPlot = false;
     private plugin = {
-        afterDraw: (chartInstance: Chart, _easing: Chart.Easing, _options?: any) => { this.afterChartDraw(chartInstance); }
+        afterDraw: (chartInstance: Chart, _easing: Chart.Easing, _options?: any) => { this.afterChartDraw(chartInstance.ctx, chartInstance.chartArea); }
     };
 
     private endSelection = () => {
@@ -181,7 +182,11 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
                 };
                 this.divRef.current.addEventListener('wheel', this.preventDefaultHandler);
             }
-            this.chartRef.current.chartInstance.render();
+            if (this.isBarPlot) {
+                this.renderChart();
+            } else {
+                this.chartRef.current.chartInstance.render();
+            }
         }
     }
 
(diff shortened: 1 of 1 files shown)


PatrickTasse:
> 1. Fixed;
> 2. Fixed;
> 3. Best to decide this when we have the data for the bar chart. Then we can evaluate if overlap is really an issue, and make changes accordingly;
> 4. Not sure I follow. For the given example, should the bottom of the y-chart display 0 or 198? Either way, the bottom value of the y-chart can be easily changed when we have the data;
> 5. Also can be easily changed when we have the data, if this is confirmed to be an issue;
> 6. Fixed.

1 and 2 look OK now.

For 4, yes I meant that bottom should be 0, but maybe that needs to be discussed and agreed. Maybe it depends on what we are showing. I was just expecting that when you have a bar chart you always see full bars and never partial bars.

6 doesn't seem to be fixed, at high zoom level with a window range of 2000ns, with 1130 data points returned from the server, I only see 9 bars. The returned y-Values are 453 x 100.0, 1 x 150.0, 676 x 200.0.

Rodrigoplp-work:
About item 6:

When the zoom level is high (very zoomed in) the backend returns repeated x values, as shown in the image below (click to expand).

Screenshot 2022-03-28 at 11 38 45

So the bar chart plots all points, but they all happen at the same few points along the x-axis.

The bar chart did not change this. This happens also on master, with the line chart, but the line chart connects the points, while the bar chart comes back to zero between them.

PatrickTasse:
> About item 6:
> 
> When the zoom level is high (very zoomed in) the backend returns repeated x values, as shown in the image below (click to expand).

The x values are not repeated, it's a bug in Chrome DevTools Network tab, it does not display bigint values properly, they are shown as Number with loss of precision.

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -321,6 +326,75 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
+
+                // Bar chart
+                if (ctx) {
+                    ctx.clearRect(0, 0, chartWidth, chartHeight);
+                    ctx.save();
+                    data.forEach((row, i) => {
+                        ctx.fillStyle = this.state.xyData.datasets[i].borderColor;
+                        row.forEach((tupple: any) => {
+                            ctx.beginPath();
+                            ctx.fillRect(xScale(Number(tupple.xValue)), chartHeight, 2, - chartHeight + yScale(+tupple.yValue));
```
PatrickTasse, line 379: This is where the problem happens, Number(tupple.xValues) creates loss of precision from the bigint value.

Not sure, maybe this can be fixed by using getXForTime() ?

Also, is it intentional or a typo the + in: yScale(+tupple.yValue) ?

PatrickTasse (changes requested):

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -321,6 +326,75 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
+            });
+
+            const yScale = scaleLinear()
+                .domain([this.state.allMin, Math.max(this.state.allMax, 1)])
+                .range([chartHeight - this.margin.bottom, this.margin.top]);
+
+            const xDomain = this.state.xyData.labels.length - 1;
+
+            const xScale = scaleLinear()
+                .domain([this.state.xyData.labels[0], this.state.xyData.labels[xDomain]].map(Number))
```
PatrickTasse, line 357: The xScale is probably also incorrect due to the map(Number) of the bigint value causing loss of precision.

packages/react-components/src/components/xy-output-component.tsx
Rodrigoplp-work: Using `getXForTime()` now.
The `+` is not a typo, it's a unary plus operator.

packages/react-components/src/components/xy-output-component.tsx
Rodrigoplp-work: Indeed. Using `getXForTime()` now.

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -576,24 +690,49 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
     }
 
     private getTimeForX(x: number): bigint {
```
PatrickTasse, line 692: It seems getTimeForX and getXForTime could be simplified if we always use this.getChartWidth(). You wouldn't even need to do a null check on the chart then.

When I debug, this.getChartWidth() and this.chartRef.current.chartInstance.width are almost always the same.

Sometimes they are not for a short time because the latter gets delayed update after a draw, I'm not even sure which width we should be using. And sometimes the chartInstance is undefined for a short time, perhaps during chart.js internal update. Also if you quickly resize the sash so that there is no room for the chart, this.getChartWidth() returns 0 but  this.chartRef.current.chartInstance.width gets stuck at the last visible non-zero width.

PatrickTasse (changes requested):

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -401,22 +478,30 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
             if (this.props.selectionRange) {
                 const startPixel = this.getXForTime(this.props.selectionRange.getStart());
                 const endPixel = this.getXForTime(this.props.selectionRange.getEnd());
                 ctx.strokeStyle = '#259fd8';
                 ctx.fillStyle = '#259fd8';
-                this.drawSelection(chart, startPixel, endPixel);
+                if (this.isBarPlot) {
+                    this.drawD3Selection(startPixel, endPixel);
+                } else {
+                    this.drawSelection(chart, startPixel, endPixel);
```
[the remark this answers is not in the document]
Rodrigoplp-work: Changed to ctx.

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -576,23 +654,37 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
-        if (!this.chartRef.current?.chartInstance) {
-            return BigInt(0);
+        if (this.isBarPlot) {
+            if (!this.chartRef.current) {
+                return BigInt(0);
+            }
+        } else {
+            if (!this.chartRef.current?.chartInstance) {
+                return BigInt(0);
+            }
```
PatrickTasse, line 664: Since we're not using the chart anymore, maybe we can skip these checks altogether?

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -576,23 +654,37 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
         }
         const offset = this.props.viewRange.getOffset() ?? BigInt(0);
         const duration = this.props.viewRange.getDuration();
+        const chartWidth = this.getChartWidth();
```
PatrickTasse, line 668: To be theoretically robust, we should check for 0 to avoid div by zero below.
But in practice I think this method is only called from mouse events which can't happen if chart width is 0...

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -576,23 +654,37 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
-        if (!this.chartRef.current?.chartInstance) {
-            return 0;
+        if (this.isBarPlot) {
+            if (!this.chartRef.current) {
+                return 0;
+            }
+        } else {
+            if (!this.chartRef.current?.chartInstance) {
+                return 0;
+            }
```
PatrickTasse, line 682: Since we're not using the chart anymore, maybe we can skip these checks altogether?

PatrickTasse (changes requested):

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -65,8 +65,9 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
     private startPositionMouseRightClick = BigInt(0);
     private margin = { top: 15, right: 0, bottom: 5, left: this.getYAxisWidth() };
     private isScatterPlot: boolean = this.props.outputDescriptor.id.includes('scatter');
+    private isBarPlot = true;
```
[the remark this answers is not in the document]

packages/react-components/src/components/xy-output-component.tsx

packages/react-components/src/components/xy-output-component.tsx
Rodrigoplp-work: Added the check.

packages/react-components/src/components/xy-output-component.tsx

PatrickTasse (approved):

bhufmann (approved):
LGTM. Looking forward to have d3 charts used for all xy charts. Thanks for providing this.

PR #697, opened 2022-03-07
Cursor is a cross when graph selection is possible

### LINKED ISSUE #162: Cursor is a cross when graph selection is possible
The cursor being a cross when selection is possible supports intuitive understanding by following UI standards set by other tools (ex. Grafana, Google Slides):

cursor-cross

grafana-cross-cursor

hriday-panchasara:
timeline-charts and xy-charts show crosshair on hover to indicate that selection is possible

fixes #162

### DIFF
--- a/packages/react-components/style/output-components-style.css
+++ b/packages/react-components/style/output-components-style.css
@@ -84,6 +84,7 @@
 .xy-main {
     width: 100%;
     display: flex;
+    cursor: crosshair;
 }
 
 #main {
@@ -247,4 +248,8 @@ canvas {
 
 .hoverClass:hover {
     border: 1px solid black;
+}
+
+#main-timegraph-content {
+    cursor: crosshair;
 }


bhufmann (commented):
Thanks for the patch. I don't think the cursor should be changed to crosshair all the time. I prefer the current implementation, where the cursor only changes when clicking with left mouse button until released. 
cross-hair

The default cursor makes also sense because user can point to a state or line to get tooltip information, as well as it's possible to select a state in timeline-charts.

This current behavior is consistent with the right mouse-click and drag where the cursor changes upon right mouse click.
right-zoom

hriday-panchasara:
> Thanks for the patch. I don't think the cursor should be changed to crosshair all the time. I prefer the current implementation, where the cursor only changes when clicking with left mouse button until released. cross-hair
> 
> The default cursor makes also sense because user can point to a state or line to get tooltip information, as well as it's possible to select a state in timeline-charts.
> 
> This current behavior is consistent with the right mouse-click and drag where the cursor changes upon right mouse click. right-zoom

I agree with your statements. I'll close this PR.

PR #698, opened 2022-03-07
Add ADR for Hotjar investigation

hoangphamEclipse:
The trace extension would like to obtain user feedback to improve its UI/UX.
This commit adds an ADR about using Hotjar investigation to track user happiness
for the Theia trace extension. It is a thorough analysis of Hotjar, its limitations
and the constraints of the trace extension. The conclusion is that we cannot use
Hotjar because the extension cannot have any commercial dependencies.


doc/adr/0004-Using-Hotjar-to-track-user-happiness.md
```diff
@@ -0,0 +1,66 @@
+# 3. Using hotjar to track user happiness
+
+Date: 2022-02-22
```
Rodrigoplp-work, line 3: Better if updated to today.

doc/adr/0004-Using-Hotjar-to-track-user-happiness.md
```diff
@@ -0,0 +1,66 @@
...
+    2. We need to add some [tracking code][tracking] to the ```<head>``` element of the web application. This code is provided by Hotjar.
+    3. After adding the code, we need to login to our Hotjar account and trigger the validation process. 
+* If we decide to support both internal and external domains, this might create possible issues in the future with the ever growing number of domains, unless we are sure that the number of domains do not change.
+* Each domain must create its own UI Widget, thus the results are not centralized.
+* Also, since the Trace Extension is part of Theia, we need to inject the Hotjar code into Theia ```<head>``` element. This brings up the question of how can we inject the code gracefully.
+* Hotjar has multiple pricing options. The free version allows the creation of 3 incoming feedback and surveys at a time. If we need more, then we cannot use Hotjar as it creates commercial dependencies.
+
+## Decision
+
+This proposed is declined since Hotjar creates commercial dependency for the Theia trace extension. We need access to [APIs](apis) that are only offered to paid subscriptions.
```
Rodrigoplp-work, line 54: Might be better phrased with something like:

> This proposal was declined since Hotjar would add a commercial dependency to Theia Trace Extension, which goes against the constrains.
> Also, in order to achieve all the requirements, it would be necessary to purchase a paid subscription from Hotjar, which is unfeasible in an open source project as Theia Trace Extension.

Rodrigoplp-work (changes requested):

Rodrigoplp-work (approved):

MatthewKhouzam (approved):

PR #701, opened 2022-03-09
Persist opened views on page refresh

williamsyang-work:
Implement StatefulWidget in the TraceViewer class and hook into the StoreState and RestoreState methods, which preserve data in LocalStorage.  This works with browser and electron.

Currently only storing outputDescriptors (opened views), but can built upon for future features (persisting view-range, selection-range, and grid-layout).

Handled an edge-case bug where timeline-chart components crashed when state.style.width = 0 in TraceContextComponent.  This also stops unopened experiment tab views from loading
on startup.


theia-extensions/viewer-prototype/src/browser/trace-viewer/trace-viewer.tsx
```diff
@@ -241,6 +241,14 @@ export class TraceViewerWidget extends ReactWidget {
             });
     }
 
+    storeState(): OutputDescriptor[] {
+        return this.outputDescriptors;
+    }
+
+    restoreState(state: OutputDescriptor[]) {
```
bhufmann, line 248: add return type `void`. I get a linter warning in VsCode.

theia-extensions/viewer-prototype/src/browser/trace-viewer/trace-viewer.tsx
```diff
@@ -241,6 +241,14 @@ export class TraceViewerWidget extends ReactWidget {
             });
     }
 
+    storeState(): OutputDescriptor[] {
+        return this.outputDescriptors;
```
bhufmann, line 245: Just to let you know, the calling method is using JSON.stringify(). The data structure OutputDescriptor defines a 2 optional fields `start` and `end` which are of type `bigint`. The JSON.stringify() will throw an exception because it can't serialize bigints. 

With the current Trace Compass server back-end `start` and `end` is never sent and hence there are not there when storeState() is called. Hence you won't see this exception. 

I'm currently debating with myself, whether we merge this PR as is and address the serialization issue later on, or we handle it right away. I'll let you know. Please still fix the comment on line 248 below right way.

bhufmann (changes requested):

bhufmann (approved):
Thanks for this improvement. Looks good to me.

PatrickTasse (changes requested):
Could you please rebase your branch on latest master commit instead of including a merge commit in the branch?

williamsyang-work:
Something got messed up with reverting the changes.  Going to re-open this PR.

PR #702, opened 2022-03-10
Move xy-chart's data with CTRL + mouse drag

### LINKED ISSUE #637: Support window dragging wiht CTRL+mouse-left and mouse-middle drag  in xy charts
In comparison to Time Graph charts, the xy-chart hasn't implemented the window drag with `CTRL+mouse left button drag` nor with `middle mouse button drag`. This shall be added to the XY-charts. The cursor shall be update to the hand icon during this operation.

missing-middle-mouse-drag

Rodrigoplp-work:
Zoom in into the xy-chart, press the control key and move the mouse with the left mouse button pressed to move the data of the xy-chart to the left or the right.

Lifting the control key while dragging does not stop the drag.

Fixes #637.

### DIFF
--- a/packages/react-components/src/components/xy-output-component.tsx
+++ b/packages/react-components/src/components/xy-output-component.tsx
@@ -31,7 +31,6 @@ type XYOuputState = AbstractOutputState & {
     allMax: number;
     allMin: number;
     cursor?: string;
-    shiftKey: boolean;
 };
 const RIGHT_CLICK_NUMBER = 2;
 const ZOOM_IN = true;
@@ -55,12 +54,16 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
     private divRef: any;
     private yAxisRef: any;
     private mouseIsDown = false;
+    private isPanning = false;
+    private isSelecting = false;
     private positionXMove = 0;
     private positionYMove = 0;
     private isRightClick = false;
     private isMouseMove = false;
     private posPixelSelect = 0;
     private isMouseLeave = false;
+    private mousePanningStart = BigInt(0);
+    private resolution = 0;
     private startPositionMouseRightClick = BigInt(0);
     private margin = { top: 15, right: 0, bottom: 5, left: this.getYAxisWidth() };
     private isScatterPlot: boolean = this.props.outputDescriptor.id.includes('scatter');
@@ -68,14 +71,23 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
         afterDraw: (chartInstance: Chart, _easing: Chart.Easing, _options?: any) => { this.afterChartDraw(chartInstance); }
     };
 
-    private endSelection = () => {
+    private endSelection = (event: MouseEvent) => {
         if (this.isRightClick) {
            this.applySelectionZoom();
         }
         this.mouseIsDown = false;
-        if (!this.state.shiftKey) {
-            this.setState({ cursor: 'default'});
+        this.isSelecting = false;
+        this.isPanning = false;
+        if (!event.shiftKey && !event.ctrlKey) {
+            this.setState({ cursor: 'default' });
+        } else if (!event.shiftKey && event.ctrlKey) {
+            this.setState({ cursor: 'grabbing' });
+        } else if (event.shiftKey && !event.ctrlKey) {
+            this.setState({ cursor: 'crosshair' });
+        } else {
+            this.setState({ cursor: 'default' });
         }
+
         document.removeEventListener('mouseup', this.endSelection);
     };
 
@@ -96,8 +108,7 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
             columns: [{title: 'Name', sortable: true}],
             allMax: 0,
             allMin: 0,
-            cursor: 'default',
-            shiftKey: false
+            cursor: 'default'
         };
 
         this.afterChartDraw = this.afterChartDraw.bind(this);
@@ -485,17 +496,26 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
         const startTime = this.getTimeForX(event.nativeEvent.offsetX);
         if (event.button === RIGHT_CLICK_NUMBER) {
             this.isRightClick = true;
+            this.isSelecting = false;
             this.setState({ cursor: 'col-resize' });
             this.startPositionMouseRightClick = startTime;
         } else {
             this.isRightClick = false;
-            this.setState({ cursor: 'crosshair' });
-            if (event.shiftKey && this.props.unitController.selectionRange) {
+            if (event.shiftKey && !event.ctrlKey && this.props.unitController.selectionRange) {
+                this.isSelecting = true;
+                this.setState({ cursor: 'crosshair' });
                 this.props.unitController.selectionRange = {
                     start: this.props.unitController.selectionRange.start,
                     end: startTime
                 };
-            } else {
+            } else if (event.ctrlKey && !event.shiftKey) {
+                this.resolution = this.getChartWidth() / Number(this.props.unitController.viewRangeLength);
+                this.mousePanningStart = this.props.unitController.viewRange.start + BIMath.round(event.nativeEvent.x / this.resolution);
+                this.isPanning = true;
+                this.setState({ cursor: 'grabbing' });
+            } else if (!(event.shiftKey && event.ctrlKey)){
+                this.isSelecting = true;
+                this.setState({ cursor: 'crosshair' });
                 this.props.unitController.selectionRange = {
                     start: startTime,
                     end: startTime
@@ -504,6 +524,19 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
             this.onMouseMove(event);
         }
         document.addEventListener('mouseup', this.endSelection);
+    }
+
+    private panHorizontally(event: React.MouseEvent) {
+        const delta = event.nativeEvent.x;
+        const change = Number(this.mousePanningStart) - (delta / this.resolution);
+        const min = BigInt(0);
+        const max = this.props.unitController.absoluteRange - this.props.unitController.viewRangeLength;
+        const start = BIMath.clamp(change, min, max);
+        const end = start + this.props.unitController.viewRangeLength;
+        this.props.unitController.viewRange = {
+            start,
+            end
+        };
     }
 
     private updateRange(rangeStart: bigint, rangeEnd: bigint) {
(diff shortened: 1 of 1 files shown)


PatrickTasse (changes requested):
- When you press Ctrl (and only Ctrl) the cursor should change to grabbing.
- Once you start panning or selecting with mouse button down, the current operation continues until the mouse button is released, so the cursor should remain crosshair/grabbing regardless of whether modifyier keys are pressed or released.
- Hold Shift down (crosshair), and tap Ctrl, the cursor returns to default on key down, but it should return to crosshair on key up because only Shift is down. And inversely should happen if you hold Ctrl and tap Shift.
- Perhaps could be another PR, but for consistency with time graph, Middle-button drag should also behave the same as Ctrl+Left-button drag.
- Definitely another PR: The temporary disconnect between selection range and the underlying xy graph while panning is quite off-putting... I'll write an issue if there isn't one.

Rodrigoplp-work:
Implemented the first 3 changes of the list above. The last 2 will be left for other PRs.

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -810,7 +840,19 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
+                            this.setState({ cursor: 'crosshair', shiftKey: true });
+                        }
+                    }
+                    break;
+                }
+                case 'Control': {
+                    if (key.shiftKey) {
+                        this.setState({ cursor: 'default' });
+                    } else if (!this.state.shiftKey) {
+                        this.setState({ cursor: 'grabbing', ctrlKey: true });
```
PatrickTasse, line 855: These keys interfere with the cursor when we are already panning or selecting. They could be ignored when isPanning is true, but we have to find another condition for selection, unless we also add a isSelecting state?

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -745,12 +771,16 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
         this.positionYMove = event.nativeEvent.offsetY;
         this.isMouseLeave = false;
 
-        if (this.mouseIsDown && !this.isRightClick) {
-            this.updateSelection();
-        }
-        if (this.mouseIsDown && this.isRightClick) {
-            this.forceUpdate();
+        if (this.mouseIsDown) {
+            if (this.state.isPanning && !(event.shiftKey && event.ctrlKey)) {
```
PatrickTasse, line 775: It should not matter which keys are pressed, if you start panning it should keep panning until mouse is released.

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -745,12 +771,16 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
+                this.panHorizontally(event);
+            } else if (!this.isRightClick && !event.ctrlKey) {
```
PatrickTasse, line 777: It should not matter which keys are pressed, if you start selecting it should keep selecting until mouse is released.

PatrickTasse (changes requested):

packages/react-components/src/components/xy-output-component.tsx
PatrickTasse: Maybe could use this.mouseIsDown...

packages/react-components/src/components/xy-output-component.tsx

packages/react-components/src/components/xy-output-component.tsx
Rodrigoplp-work: Fixed.

packages/react-components/src/components/xy-output-component.tsx
Rodrigoplp-work: Yeah, I used the mouseIsDown check in `private onMouseMove()`, so here we can leave as it is.

PatrickTasse: The cursor changes if you press and release the modifier keys during an operation (selecting or panning). I think the cursor should be fixed to the current operation in progress.

While selecting, holding the Ctrl key will temporarily stop the selection from extending to the mouse position.

PatrickTasse (changes requested):
Please rebase on top of latest master so the unrelated commits are not part of this branch.

packages/react-components/src/components/xy-output-component.tsx
Rodrigoplp-work: You are right. I used your suggestion of `isSelection`, as `mouseIsDown` was not enough to solve this.

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -819,11 +868,30 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
-            if (!this.mouseIsDown) {
-                change.cursor = 'default';
+        if (key.key === 'Shift' || key.key === 'Control') {
+            const change: { cursor: string | undefined, shiftKey: boolean, ctrlKey: boolean } = {
+                cursor: this.state.cursor ?? 'default',
+                shiftKey: this.state.shiftKey ?? false,
+                ctrlKey: this.state.ctrlKey ?? false
+            };
+
+            if (key.key === 'Shift') {
```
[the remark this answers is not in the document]
Rodrigoplp-work: Fixed.

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -809,8 +844,22 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
+                    }
+                    break;
+                }
+                case 'Control': {
+                    if (!this.isSelecting && !this.isPanning) {
+                        if (key.shiftKey) {
+                            this.setState({ cursor: 'default' });
+                        } else if (!this.state.shiftKey) {
+                            this.setState({ cursor: 'grabbing', ctrlKey: true });
+                        }
```
PatrickTasse, line 862: The shiftKey and ctrlKey state is not updated properly because it depends on other conditions than just the key being pressed.

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -73,8 +78,12 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
         }
         this.mouseIsDown = false;
-        if (!this.state.shiftKey) {
-            this.setState({ cursor: 'default'});
+        this.isSelecting = false;
+        if (!this.state.shiftKey && !this.state.ctrlKey) {
+            this.isPanning = false;
+            this.setState({ cursor: 'default' });
+        } else if (this.state.ctrlKey) {
+            this.setState({ cursor: 'grabbing' });
```
PatrickTasse, line 86: When you release the mouse then you should neither be isPanning or isSelecting, regardless of keys being pressed (the keys should only affect the cursor).

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -73,8 +78,12 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
            this.applySelectionZoom();
         }
         this.mouseIsDown = false;
-        if (!this.state.shiftKey) {
-            this.setState({ cursor: 'default'});
+        this.isSelecting = false;
+        if (!this.state.shiftKey && !this.state.ctrlKey) {
+            this.isPanning = false;
+            this.setState({ cursor: 'default' });
+        } else if (this.state.ctrlKey) {
```
PatrickTasse, line 85: Cursor should go to grabbing if and only if ctrlKey is pressed.
Cursor should go to crosshair if and only if shiftKey is pressed.

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -819,11 +868,30 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
+            if (key.key === 'Shift') {
+                change.shiftKey = false;
+                if (key.ctrlKey && !this.isSelecting && !this.isPanning) {
+                    change.cursor = 'grabbing';
+                } else if (!this.mouseIsDown && !this.isSelecting) {
+                    change.cursor = 'default';
+                }
+            } else {
+                change.ctrlKey = false;
+                if (key.shiftKey && this.isSelecting && !this.isPanning) {
```
PatrickTasse, line 887: I think this should be !thisIsSelecting?

But anyways should be simpler: if either isSelecting or isPanning is true, then cursor should not change under any condition.

And I think shiftKey or ctrlKey state should be updated based on the key being released and no other condition?

PatrickTasse (changes requested):

packages/react-components/src/components/xy-output-component.tsx
PatrickTasse: Actually, I don't think these are needed to be added to the state because these modifier key states are part of every KeyEvent or MouseEvent? Better use that directly than rely on our own copy...

PatrickTasse: Use case: Hold mouse down, hold Shift down, release mouse, cursor should remain crosshair.
Use case: Hold Ctrl down, hold mouse down, release Ctrl, hold Shift, release mouse, cursor should change to crosshair.
Use case: Hold Ctrl down, hold mouse down, hold Ctrl+Shift down, release mouse, cursor should change to default.

Rodrigoplp-work: Fixed.

packages/react-components/src/components/xy-output-component.tsx
Rodrigoplp-work: Agreed.

packages/react-components/src/components/xy-output-component.tsx

packages/react-components/src/components/xy-output-component.tsx

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -55,27 +54,38 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
            this.applySelectionZoom();
         }
         this.mouseIsDown = false;
-        if (!this.state.shiftKey) {
-            this.setState({ cursor: 'default'});
+        this.isSelecting = false;
+        this.isPanning = false;
+        if (!event.shiftKey && !event.ctrlKey) {
+            this.setState({ cursor: 'default' });
+        } else if (event.ctrlKey) {
```
PatrickTasse, line 83: Both keys pressed leads to grabbing, should be default. I suggest:
if (shift !ctrl) crosshair else if (ctrl !shift) grabbing else default.

PatrickTasse (changes requested):

packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -55,27 +54,38 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
...
+    private resolution = 0;
     private startPositionMouseRightClick = BigInt(0);
     private margin = { top: 15, right: 0, bottom: 5, left: this.getYAxisWidth() };
     private isScatterPlot: boolean = this.props.outputDescriptor.id.includes('scatter');
     private plugin = {
         afterDraw: (chartInstance: Chart, _easing: Chart.Easing, _options?: any) => { this.afterChartDraw(chartInstance); }
     };
 
-    private endSelection = () => {
+    private endSelection = (event: any) => {
```
PatrickTasse, line 74: Perhaps we are sure it will be a MouseEvent instead of any?

Rodrigoplp-work: Ah, you are right. After our conversation I found the problem: I was trying to use `React.MouseEvent`. Using only `MouseEvent` works without errors.

PatrickTasse (approved):

bhufmann (approved):
Looks good to me. I agree to implement middle mouse drag in a separate PR. Thanks for the improvement.

PR #710, opened 2022-03-16
Indicate when data is still loading

### LINKED ISSUE #649: Views: Visually show when data is still loading
If data that should currently be visible to the user is still loading, show this visually so that users don’t start interpreting what they see.

Some users expressed frustration that they would start interpreting the data and then additional chart elements would suddenly appear.

e.g. Grey out the view

hriday-panchasara:
Users expressed frustration that they would start interpreting the data and then additional chart elements would suddenly appear. Therefore in timeline-chart and xy-charts when data is being fetched due to user interactions (toggle tree, zoom and pan) show a loading spinner next to the title bar.

Examples:
Screen Recording 2022-03-16 at 2 57 28 PM
Screen Recording 2022-03-16 at 3 03 33 PM
Screen Recording 2022-03-16 at 3 04 19 PM

Fixes #649

### DIFF
--- a/packages/react-components/src/components/abstract-output-component.tsx
+++ b/packages/react-components/src/components/abstract-output-component.tsx
@@ -92,6 +92,8 @@ export abstract class AbstractOutputComponent<P extends AbstractOutputProps, S e
             </button>
             <div className='title-bar-label' title={outputName} onClick={() => this.setFocus()}>
                 {outputName}
+                <i id={this.props.traceId + this.props.outputDescriptor.id + 'handleSpinner'} className='fa fa-refresh fa-spin'
+                    style={{ marginTop: '5px', visibility: 'hidden'}} />
             </div>
         </React.Fragment>;
     }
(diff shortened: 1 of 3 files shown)


hriday-panchasara:
@PatrickTasse removed animation from loading icon

PatrickTasse (dismissed):

bhufmann:
> @PatrickTasse removed animation from loading icon

@hriday-panchasara why was the animation removed? I would expect a spinning wheel instead. Without, it looks like a refresh button which the user could press. Please let me know.

hriday-panchasara:
> > @PatrickTasse removed animation from loading icon
> 
> @hriday-panchasara why was the animation removed? I would expect a spinning wheel instead. Without, it looks like a refresh button which the user could press. Just curious. Please let me know.

Since the icon appears everytime new data is being rendered, I thought removing the animation would bring less attention to it. But I agree with your comment as well, and I've added the animation back

PatrickTasse (approved):

bhufmann (approved):
Looks good to me. Thanks!

PR #711, opened 2022-03-16
Update instructions to set up trace server

hoangphamEclipse:
The current instructions to set up the trace server for the trace extension does not explicitly indicate steps that only apply to Linux users. In addition, the lack of clear instructions for manually downloading and setting up the trace server makes it difficult for developers, who uses other OS, to run the trace server. This commit improves the documentation by providing clear instructions for developers on how to set up the server either by (1) using yarn, (2) downloading the server manually, or (3) building server manually. This will help contributors speed up their installation process, and reduce support effort.


README.md
```diff
@@ -97,13 +97,31 @@ You can find those example applications under `examples/`.
...
-In order to open traces, you need a trace server running on the same machine as the trace extension. You can download the [Eclipse Trace Compass server](https://download.eclipse.org/tracecompass.incubator/trace-server/rcp/?d) or let `yarn` download and run it:
+In order to open traces, you need a trace server running on the same machine as the trace extension. There are 3 options to run the trace server.
+
+#### Using yarn (Linux only)
+
+If you are working on Linux, you can let `yarn` download and run it:
 
 ```bash
 yarn download:server
 yarn start:server
```
bhufmann, line 108: First of all, if someone start the server externally, then that user has to stop by itself. Starting it manually, has the advantage, that the open-traces are populated with traces on the server, or when starting the server in debug mode from the server development environment to debug the server.

Secondly, the theia-trace-extension has a feature to automatically start the trace server from the application. There is a command or when trying to open a trace it will start it. That started server will be stopped by the applicaton as well.  By default, it uses the location where `yarn download:server` stores the server for both the browser backend and electron app.

If a user wants to use a different server build, the there is a preference in the application where the user can change the path to the custom server which will be used when starting it automatically. 

Could you please describe in this PR the trace viewer preference? Also describe the Start/stop trace server command (from the command menu opened using F1 or CTRL+SHIFT+P)

README.md
```diff
@@ -97,13 +97,31 @@ You can find those example applications under `examples/`.
...
 ```
 
+#### Manual download (Windows, Linux and MacOS)
+
+Alternatively, you can download the [Eclipse Trace Compass server](https://download.eclipse.org/tracecompass.incubator/trace-server/rcp/?d). Select the version of the trace server that best matches your working environment. There are different versions of the trace server for Linux, Windows and MacOS.
+
+After you have successfully downloaded the Trace Server, extract the compressed files, rename the extracted folder to `trace-compass-server`, then copy it into the root of the repo.
```
bhufmann, line 115: After extracting the server, the user can update the preference I mentioned earlier to allow the application to start the server automatically. Then you don't need to rename it.

README.md
```diff
@@ -97,13 +97,31 @@ You can find those example applications under `examples/`.
...
+
+Then, you can start the server with:
+
+```bash
+yarn start:server
```
bhufmann, line 120: This is only if a user wants to start it manually.

bhufmann (changes requested):

README.md
hoangphamEclipse: > If a user wants to use a different server build, the there is a preference in the application where the user can change the path to the custom server which will be used when starting it automatically.

By this, do you mean what is described [here](https://github.com/theia-ide/theia-trace-extension#change-the-trace-server-url)?

hoangphamEclipse:
@bhufmann Just letting you know that this PR is updated.

README.md
```diff
@@ -96,14 +96,45 @@ You can find those example applications under `examples/`.
...
 
 ```bash
-yarn download:server
 yarn start:server
 ```
 
-You can also build the trace-server yourself using Trace Compass and the Incubator. Take a look at the [instructions here][tc-server-build].
+Please note that you need a Trace Server Build before starting the server. See [Getting a Trace Server build][tc-server-download-guide] for more details. You will need to stop the server manually as well.
+
+#### Starting the Trace Server from the Theia application
```
bhufmann, line 118: In this chapter you should describe the preference setting for the trace server. User can change it to point to point to a custom trace server.

bhufmann (changes requested):
I just noticed your updates. Please describe the trace server related preferences and that it will be used when starting the server from inside the application.

PR #713, opened 2022-03-22
Throttle xy-chart data update

### LINKED ISSUE #453: Misalignment issues between the timeline and XY charts.
Thanks for this nice improvement!
I did notice some misalignment issues between the timeline and XY charts. For example when panning the timeline chart at a high zoom level using 'a' or 'd' shortcuts, the selection range on the XY chart dances around and sometimes ends up misaligned with the selection range on the timeline chart. There are no issues the other way around, and the way to correct the misalignment is to pan the XY chart. See the gif attached.

xy-chart-nav

_Originally posted by @mirsky-work in https://github.com/theia-ide/theia-trace-extension/issues/419#issuecomment-899810661_

### LINKED ISSUE #470: Throttling of XY chart data update
Currently the XY chart requests a new data model from the trace-server every time the window range is updated.

When navigation with the mouse or keyboard, this can lead to many requests within a short amount of time.

Moreover these requests are processed asynchronously in parallel, and the XY chart is updated from the returned model as soon as each response is received. This means that if the last request is faster than a previous one, the final XY chart data could be for a different window range than the current one, leading to a mismatch and a XY chart window range that is out of sync with the other charts. When panning continuously, we can also see the XY chart bouncing left and right in a somewhat unpredictable manner as the updates are applied out of order.

A throttling mechanism should be introduced to prevent flooding the trace-server with requests, and the algorithm should ensure that the XY chart is updated with the proper (most recent) window range.

hriday-panchasara:
Using throttling to prevent flooding the trace-server with requests when xy chart is updated. Fixes the issue of delayed movement during panning, as well as the misalignment between charts. I decided to use throttle instead of debounce since debounce would trigger only when the user stops panning, whereas throttle invokes the update once every 500ms. I've attached a video showing the performance of a panning operation, you can compare it to the video in #470 that shows a 37 second delay. Please suggest if there's any other performance optimizations for this scenario or tricks to achieve smoother pan animation.

https://user-images.githubusercontent.com/92893187/159570352-fd96001c-3810-482a-ae68-0f68d982273e.mp4

fixes #470
fixes #453

### DIFF
--- a/packages/react-components/src/components/xy-output-component.tsx
+++ b/packages/react-components/src/components/xy-output-component.tsx
@@ -18,6 +18,7 @@ import { BIMath } from 'timeline-chart/lib/bigint-utils';
 import { scaleLinear } from 'd3-scale';
 import { axisLeft } from 'd3-axis';
 import { select } from 'd3-selection';
+import { throttle } from 'lodash';
 
 type XYOuputState = AbstractOutputState & {
     selectedSeriesId: number[];
@@ -82,6 +83,8 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
     private preventDefaultHandler: ((event: WheelEvent) => void) | undefined;
 
     private onSelectionChanged = (payload: { [key: string]: string; }) => this.doHandleSelectionChangedSignal(payload);
+
+    private _throttledUpdateXY = throttle(() => this.updateXY(), 500);
 
     constructor(props: AbstractOutputProps) {
         super(props);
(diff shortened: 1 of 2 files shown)


packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -83,6 +84,8 @@ export class XYOutputComponent extends AbstractTreeOutputComponent<AbstractOutpu
 
     private onSelectionChanged = (payload: { [key: string]: string; }) => this.doHandleSelectionChangedSignal(payload);
 
+    private _throttledUpdateXY = throttle(() => this.updateXY(), 500);
```
PatrickTasse, line 87: I'm not sure, but if I understand correctly, if you don't put the trailing option then you might miss the last update if it happens within 500ms of the last throttled update?

PatrickTasse (changes requested):

packages/react-components/src/components/xy-output-component.tsx
hriday-panchasara: I thought default throttle has settings {leading: true, trailing: true}. Should I explicitly mention {leading: true, trailing: true} or {leading: false, trailing: true} ?

PatrickTasse: OK I see those defaults in the web documentation, but not in the inline Javadoc. So if the web documentation is correct, it's fine as it is.

bhufmann (approved):
LGTM. Thanks.

PR #715, opened 2022-03-23
Persist opened views on page refresh

williamsyang-work:
Implement StatefulWidget in the TraceViewer class and hook into the StoreState and RestoreState methods, which preserve data in LocalStorage.  This works with browser and electron.

Currently only storing outputDescriptors (opened views), but can built upon for future features.

Handled an edge-case bug where timeline-chart components crashed when state.style.width = 0 in TraceContextComponent.  This also stops unopened experiment tab views from loading
on startup.

### DIFF
--- a/packages/react-components/src/components/trace-context-component.tsx
+++ b/packages/react-components/src/components/trace-context-component.tsx
@@ -307,13 +307,14 @@ export class TraceContextComponent extends React.Component<TraceContextProps, Tr
     }
 
     render(): JSX.Element {
+        const shouldRenderOutputs = (this.state.style.width > 0 && this.props.outputs.length);
         return <div className='trace-context-container'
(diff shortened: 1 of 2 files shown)


packages/react-components/src/components/trace-context-component.tsx
```diff
@@ -307,13 +307,14 @@ export class TraceContextComponent extends React.Component<TraceContextProps, Tr
     }
 
     render(): JSX.Element {
+        const shouldRenderOutputs = (this.state.style.width > 0 && this.props.outputs.length);
```
williamsyang-work, line 310: @PatrickTasse removed state variable and added simple check.

PatrickTasse (approved):

bhufmann (approved):
LGTM. Thanks!

PatrickTasse (approved):

PR #718, opened 2022-03-24
Expanded persisted state

williamsyang-work:
Now storing currentViewRange, currentTimeSelection, and layouts.

PersistedState is pulled from trace-context-component using a React.refObj and accessing a new getter fxn for persistedState.

BigInt handling implemented for TimeRanges, but still needed for Outputs.start & end.


williamsyang-work:
Something I noticed is that the selectedViewRange does not render on refresh for Timegraph components:
screen-recorder-fri-mar-25-2022-12-35-37

PatrickTasse (changes requested):

packages/base/src/utils/time-range.ts
```diff
@@ -1,18 +1,42 @@
...
-     * @param offset Time offset, if this is defined the start and end time should be relative to this value
-     */
-    constructor(start: bigint, end: bigint, offset?: bigint) {
-        this.start = start;
-        this.end = end;
-        this.offset = offset;
+    constructor();
+    constructor(timeRangeString: TimeRangeString);
+    constructor(start: bigint, end: bigint, offset?: bigint);
+    constructor(a?: TimeRangeString | bigint, b?: bigint, c?: bigint) {
```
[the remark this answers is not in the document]
williamsyang-work: This does work.  Apparently this is the common pattern for "overloading" the constructor in TypeScript / JavaScript, because proper overloading is not allowed.

We can define a number of constructors in our class but we have to give one common implementation for all the constructors defined inside this constructor we can write our own logic in TypeScript.

packages/base/src/utils/time-range.ts
```diff
@@ -1,18 +1,47 @@
...
      */
-    constructor(start: bigint, end: bigint, offset?: bigint) {
-        this.start = start;
-        this.end = end;
-        this.offset = offset;
+    constructor(timeRangeString: TimeRangeString);
+    /**
+     * Constructor.
+     * @param timeRangeString string object returned by this.toString()
+     */
```
PatrickTasse, line 27: Need to swap the last two Javadocs

PatrickTasse (changes requested):

packages/base/src/utils/time-range.ts
PatrickTasse: Or swap the constructors. Probably makes more sense to have the one with (start, end, offset) before the one with (TimeRangeString).

williamsyang-work: Oops, sorry!

williamsyang-work: @PatrickTasse I put the JavaDocs in the correct spot, and moved the (start, end offset) constructor to the first position.

PatrickTasse (approved):

bhufmann:
> Something I noticed is that the selectedViewRange does not render on refresh for Timegraph components:
> 
> screen-recorder-fri-mar-25-2022-12-39-42

The selection in the timegraph compenents is still not restored, as well as the selection when the events table is open. What is the plan for supporting these 2 components as well?

PatrickTasse (changes requested):
I can reproduce the issue of selection range lost in time graphs after refresh.

bhufmann (approved):
Looks good to me. Thanks for the enhancement.

packages/react-components/src/components/table-output-component.tsx
```diff
@@ -382,6 +382,10 @@ export class TableOutputComponent extends AbstractOutputComponent<TableOutputPro
                 this.timestampCol = timestampHeader.getColDef().field;
             }
         }
+
+        if (this.props.unitController.selectionRange) {
+            this.handleTimeSelectionChange(this.props.unitController.selectionRange);
```
PatrickTasse, line 387: The selection of the index range is OK but ensuring the start index is visible does not always work.

I suspect it depends on timing, if fetching the selection index from the trace-server returns before or after the initial fetching of table lines.

I'm OK to fix this as a separate PR (or not fixed). The vertical scroll position of time graph rows and xy-chart checkbox tree is not restored either ;) The selected elements in xy-chart checkbox are also not restored. But we can expect some things to be reset when refreshing the page...

PatrickTasse (approved):

PR #733, opened 2022-04-14
Undo / Redo Functionality

williamsyang-work:
Adds the ability for the user to undo / redo changes made to the ViewRange and SelectionRange.  Adds buttons to the toolbar for these actions.

### DIFF
--- a/packages/react-components/src/components/trace-context-component.tsx
+++ b/packages/react-components/src/components/trace-context-component.tsx
@@ -27,10 +27,11 @@ import { TooltipXYComponent } from './tooltip-xy-component';
 import { BIMath } from 'timeline-chart/lib/bigint-utils';
 import { DataTreeOutputComponent } from './datatree-output-component';
 import { cloneDeep } from 'lodash';
+import { UnitControllerHistoryHandler } from './utils/unit-controller-history-handler';
 
 const ResponsiveGridLayout = WidthProvider(Responsive);
 
-interface TraceContextProps {
+export interface TraceContextProps {
     tspClient: TspClient;
     experiment: Experiment;
     outputs: OutputDescriptor[];
@@ -45,7 +46,7 @@ interface TraceContextProps {
     persistedState?: PersistedState;
 }
 
-interface TraceContextState {
+export interface TraceContextState {
     timeOffset: bigint;
     currentRange: TimeRange;
     currentViewRange: TimeRange;
@@ -78,6 +79,7 @@ export class TraceContextComponent extends React.Component<TraceContextProps, Tr
     private readonly DEFAULT_CHART_OFFSET = 200;
 
     private unitController: TimeGraphUnitController;
+    private historyHandler: UnitControllerHistoryHandler;
     private tooltipComponent: React.RefObject<TooltipComponent>;
     private tooltipXYComponent: React.RefObject<TooltipXYComponent>;
     private traceContextContainer: React.RefObject<HTMLDivElement>;
@@ -158,6 +160,7 @@ export class TraceContextComponent extends React.Component<TraceContextProps, Tr
             const nanos = zeroPad(theNumber % BigInt(1000));
             return seconds + '.' + millis + ' ' + micros + ' ' + nanos;
         };
+        this.historyHandler = new UnitControllerHistoryHandler(this.unitController);
         if (this.props.persistedState?.currentTimeSelection) {
             const { start, end } = this.props.persistedState.currentTimeSelection;
             this.unitController.selectionRange = { start: BigInt(start), end: BigInt(end) };
@@ -197,6 +200,8 @@ export class TraceContextComponent extends React.Component<TraceContextProps, Tr
         } else {
             this.unitController.viewRange = { start: BigInt(0), end: this.state.experiment.end - this.state.timeOffset };
         }
+        this.historyHandler.clear();
+        this.historyHandler.addCurrentState();
     }
 
     private async updateTrace() {
@@ -247,12 +252,16 @@ export class TraceContextComponent extends React.Component<TraceContextProps, Tr
         signalManager().on(Signals.THEME_CHANGED, this.onBackgroundThemeChange);
         signalManager().on(Signals.UPDATE_ZOOM, this.onUpdateZoom);
         signalManager().on(Signals.RESET_ZOOM, this.onResetZoom);
+        signalManager().on(Signals.UNDO, this.undoHistory);
+        signalManager().on(Signals.REDO, this.redoHistory);
     }
 
     private unsubscribeToEvents() {
         signalManager().off(Signals.THEME_CHANGED, this.onBackgroundThemeChange);
         signalManager().off(Signals.UPDATE_ZOOM, this.onUpdateZoom);
         signalManager().off(Signals.RESET_ZOOM, this.onResetZoom);
+        signalManager().off(Signals.UNDO, this.undoHistory);
+        signalManager().off(Signals.REDO, this.redoHistory);
     }
 
     async componentDidUpdate(prevProps: TraceContextProps): Promise<void> {
@@ -318,14 +327,14 @@ export class TraceContextComponent extends React.Component<TraceContextProps, Tr
             });
             this.setState(prevState => ({
                 currentTimeSelection: new TimeRange(range.start, range.end, prevState.timeOffset)
-            }));
+            }), () => this.updateHistory());
         }
     }
 
     private handleViewRangeChange(viewRange: TimelineChart.TimeGraphRange) {
         this.setState(prevState => ({
             currentViewRange: new TimeRange(viewRange.start, viewRange.end, prevState.timeOffset)
-        }));
+        }), () => this.updateHistory());
     }
 
     private onContextMenu(event: React.MouseEvent) {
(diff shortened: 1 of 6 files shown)


packages/react-components/src/components/utils/unit-controller-history-handler.ts
```diff
@@ -0,0 +1,84 @@
...
+
+    constructor(uC: TimeGraphUnitController) {
+        this.stack = [];
+        this.unitController = uC;
+    }
+
+    public addCurrentState(): void {
+        const { selectionRange, viewRange } = this.unitController;
+        this.waitToAdd({ selectionRange, viewRange });
+    }
```
williamsyang-work, line 25: Currently only storing ViewRange and SelectionRange.  If we want anything else added please let me know!  However, it will require refactoring.

packages/react-components/src/components/utils/unit-controller-history-handler.ts
```diff
@@ -0,0 +1,84 @@
...
+        this.stack[this.i] = item;
+    }
+
+    private restore(): void {
+        this.restoring = true;
+        const { selectionRange, viewRange } = this.stack[this.i];
+        this.unitController.selectionRange = selectionRange;
+        this.unitController.viewRange = viewRange;
+        this.restoring = false;
+    }
```
williamsyang-work, line 74: The restore function directly modifies the unit controller.  When changes are made to the unit controller, TraceContextComponent State is automatically updated via onChange handlers for ViewRange and SelectionRange.

theia-extensions/viewer-prototype/src/browser/trace-viewer/trace-viewer-toolbar-commands.ts
```diff
@@ -14,6 +14,18 @@ export namespace TraceViewerToolbarCommands {
...
+        label: 'Undo',
+        iconClass: 'fa fa-undo fa-lg',
+    };
+
+    export const REDO: Command = {
+        id: 'trace.viewer.toolbar.redo',
+        label: 'Redo',
+        iconClass: 'fa fa-repeat fa-lg',
+    };
+
```
williamsyang-work, line 28: Let me know if the icons look OK to you.

packages/react-components/src/components/utils/unit-controller-history-handler.ts
bhufmann: I think it's a good start for the initial implementation. We can have subsequent PRs for more things if needed.

packages/react-components/src/components/utils/unit-controller-history-handler.ts
```diff
@@ -0,0 +1,84 @@
...
+            this.restore();
+        }
+    }
+
+    public clear(): void {
+        this.i = 0;
+        this.max = 0;
+    }
+
+    private waitToAdd(item: HistoryItem): void {
```
bhufmann, line 46: maybe call it `enqueueItem`  or `queueItem`

theia-extensions/viewer-prototype/src/browser/trace-explorer/trace-explorer-sub-widgets/trace-explorer-keyboard-shortcuts/zoom-pan-shortcuts-table.tsx
```diff
@@ -58,6 +58,20 @@ export class ZoomPanShortcutsTable extends React.Component {
...
+                                    <span className='monaco-keybinding-key'>Z</span>
+                                </td>
+                            </tr>
+                            <tr>
+                                <td><i className='fa fa-repeat fa-lg' /> Redo</td>
+                                <td className='monaco-keybinding shortcuts-table-keybinding'>
+                                    <span className='monaco-keybinding-key'>CTRL</span>
+                                    <span className='monaco-keybinding-key'>Y</span>
+                                </td>
+                            </tr>
```
bhufmann, line 74: Does this bind the command to the main menu `Edit->Undo` and `Edit->Redo`. It would be good if the main menu is also triggering the command if the trace viewer is in focus.

To be consistent the redo should CTRL+SHIFT+Z.

bhufmann (changes requested):
Thanks for this useful feature.

theia-extensions/viewer-prototype/src/browser/trace-viewer/trace-viewer-toolbar-commands.ts
williamsyang-work: I like those icons, but the spacing looks weird with them.  The FontAwesome icons for some reason float to the left, while the VSCode icons are properly centered.  They look off when next to each other, in my opinion.

image

williamsyang-work:
@bhufmann 

I'm working on Undo / Redo, and trying to implement CTRL + SHIFT + Z / Y for the keyboard shortcuts.  It appears that CTRL + Z and CTRL + SHIFT + Z keyboard events don't propagate down to trace-context-component.

To explain: I can console log CTRL, CTRL + SHIFT, and even CTRL + SHIFT + [1, Q, E, 3, etc...]in the onKeyDown function in TraceContextComponent.  However, CTRL + SHIFT + Z/Y does not get logged.

Not sure how to move forward.  Could we use a different keyboard shortcut, or only use the buttons?  My guess is that we'd have to make changes in a different repo to enable  CTRL + SHIFT + X/Y.

Code for console logging.
image

bhufmann:
> @bhufmann
> 
> I'm trying to implement `CTRL + SHIFT + Z/Y` for the keyboard shortcuts. It appears that `CTRL + Z` and `CTRL + SHIFT + Z` keyboard events don't propagate down to trace-context-component.
> 
> To explain: I can console log `CTRL`, `CTRL + SHIFT`, and even `CTRL + SHIFT + [1, Q, E, 3, etc...]` in the onKeyDown function in TraceContextComponent. However, `CTRL + SHIFT + Z/Y` does not get logged.
> 
> Not sure how to move forward. Could we use a different keyboard shortcut, or only use the buttons? My guess is that we'd have to make changes in a different repo to enable `CTRL + SHIFT + X/Y`.
> 
@williamsyang-work I think we should stick to standard keyboard shortcuts for undo/redo. I tried to use `onKeyUp` instead and the event is triggered for any key that is released after key down. Unfortunately, this is a slight different behavior when using other keys, like on the timeline-chart itself (e.g. shift + arrow key to select states). The timeline-chart adds the events to the canvas of the chart.

bhufmann:
> > @bhufmann
> > I'm trying to implement `CTRL + SHIFT + Z/Y` for the keyboard shortcuts. It appears that `CTRL + Z` and `CTRL + SHIFT + Z` keyboard events don't propagate down to trace-context-component.
> > To explain: I can console log `CTRL`, `CTRL + SHIFT`, and even `CTRL + SHIFT + [1, Q, E, 3, etc...]` in the onKeyDown function in TraceContextComponent. However, `CTRL + SHIFT + Z/Y` does not get logged.
> > Not sure how to move forward. Could we use a different keyboard shortcut, or only use the buttons? My guess is that we'd have to make changes in a different repo to enable `CTRL + SHIFT + X/Y`.
> 
> @williamsyang-work I think we should stick to standard keyboard shortcuts for undo/redo. I tried to use `onKeyUp` instead and the event is triggered for any key that is released after key down. Unfortunately, this is a slight different behavior when using other keys, like on the timeline-chart itself (e.g. shift + arrow key to select states). The timeline-chart adds the events to the canvas of the chart.

As per our conversation let's go with the following shortcuts for the undo/redo feature, which are consistent with the Theia/VsCode default shortcuts for undo/redo in code editors:
- undo: `CTRL+Z`
- redo: `CTRL+SHIFT+Z` and `CTRL+Y`

bhufmann (approved):
Thanks a lot for this enhancement.

packages/react-components/src/components/trace-context-component.tsx
```diff
@@ -348,15 +357,16 @@ export class TraceContextComponent extends React.Component<TraceContextProps, Tr
         return <div className='trace-context-container'
             onContextMenu={event => this.onContextMenu(event)}
             onKeyDown={event => this.onKeyDown(event)}
+            onKeyUp={event => this.onKeyUp(event)}
```
PatrickTasse, line 360: Found an issue with Events Table:

- select a row near the top of the table
- use the scroll bar to navigate to the bottom of the table
- select a row near the bottom of the table
- press Ctrl+Z to undo

This causes a loading of the rows. After loading, the correct row is selected. But Ctrl+Shift+Z to redo does not work. It appears it is a keyboard focus issue, the onKeyDown() isn't being called. If the Events Table title bar is clicked, it appears to give back keyboard focus and then redo works.

If this is found to be an issue not related to this patch, it can be fixed in a separate commit/PR.

packages/react-components/src/components/utils/unit-controller-history-handler.ts
```diff
@@ -0,0 +1,84 @@
...
+    protected unitController: TimeGraphUnitController;
+    protected restoring = false;
+
+    constructor(uC: TimeGraphUnitController) {
+        this.history = [];
+        this.unitController = uC;
+    }
+
+    public addCurrentState(): void {
+        const { selectionRange, viewRange } = this.unitController;
```
PatrickTasse, line 23: In some cases (e.g. making time selection by clicking) the user can make the same selection many times. Each of these create their own history entry. Maybe it shouldn't be allowed to have (consecutive) duplicates in the history?

Not sure if it's unit controller's fault to trigger a selection/view range change event when it hasn't really changed, though...

PatrickTasse (changes requested):

PatrickTasse (dismissed):

PatrickTasse (approved):

PR #735, opened 2022-04-18
Export table views to csv

hriday-panchasara:
- Added an optional 'export button' to AbstractOutputComponent. This is used by statistics tables and events table to export them to csv:

Screen Shot 2022-04-18 at 12 23 39 PM

Screen Shot 2022-04-18 at 12 24 47 PM

- For events table, we have to fetch all the table lines and then parse the response to csv format. For traces that have too many table lines we are splitting the work into smaller jobs. At the moment I'm incrementally fetching every 100000 lines. @PatrickTasse has suggested that in the future we should show a progress bar for such long jobs that allow the user to cancel. Also, should there be a limit set to how big of a table we can export? Some experiments (groups of traces) might have an events table with more than a million lines. For such tables should we show an alert message asking the user to make a smaller selection?

### DIFF
--- a/packages/react-components/src/components/table-output-component.tsx
+++ b/packages/react-components/src/components/table-output-component.tsx
@@ -67,7 +67,8 @@ export class TableOutputComponent extends AbstractOutputComponent<TableOutputPro
             outputStatus: ResponseStatus.RUNNING,
             tableColumns: [],
             optionsDropdownOpen: false,
-            showToggleColumns: false
+            showToggleColumns: false,
+            additionalOptions: true
         };
 
         this.frameworkComponents = {
@@ -609,6 +610,7 @@ export class TableOutputComponent extends AbstractOutputComponent<TableOutputPro
                 if (currRowIndexFound && !isFound && rowNode.data && rowNode.data['isMatched']) {
                     this.gridApi?.ensureIndexVisible(rowNode.rowIndex);
                     this.selectStartIndex = this.selectEndIndex = rowNode.rowIndex;
+                    this.enableIndexSelection = true;
                     this.handleRowSelectionChange();
                     rowNode.setSelected(true);
                     isFound = true;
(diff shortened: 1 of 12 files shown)


packages/react-components/src/components/table-output-component.tsx
```diff
@@ -653,4 +654,116 @@ export class TableOutputComponent extends AbstractOutputComponent<TableOutputPro
             });
         }
     }
+
+    async shareOutput(option: string): Promise<void> {
```
PatrickTasse, line 658: Would be clearer as exportOutput()

packages/react-components/src/components/table-output-component.tsx
```diff
@@ -653,4 +654,116 @@ export class TableOutputComponent extends AbstractOutputComponent<TableOutputPro
...
+        const colHeaderRow: string[] = [];
+        this.columnArray.forEach(columnHeader => {
+            colHeaderRow.push(columnHeader.headerName);
+        });
+        csv_data.push(colHeaderRow.join(','));
+
+        if (option === 'selection') {
+            if (this.startTimestamp && this.endTimestamp) {
+                const startIndex = await this.fetchTableIndex(this.startTimestamp < this.endTimestamp ? this.startTimestamp : this.endTimestamp);
+                const endIndex = await this.fetchTableIndex(this.startTimestamp > this.endTimestamp ? this.startTimestamp : this.endTimestamp);
```
PatrickTasse, line 690: Not working properly. The export should be index-based not time-based. The user could have made a specific index selection and when multiple events share the same timestamp this does not export the correct rows.

PatrickTasse (changes requested):

packages/react-components/src/components/table-output-component.tsx
hriday-panchasara: @PatrickTasse Updated PR to support index-based selections as well

packages/react-components/src/components/table-output-component.tsx
```diff
@@ -607,6 +607,7 @@ export class TableOutputComponent extends AbstractOutputComponent<TableOutputPro
                 if (currRowIndexFound && !isFound && rowNode.data && rowNode.data['isMatched']) {
                     this.gridApi?.ensureIndexVisible(rowNode.rowIndex);
                     this.selectStartIndex = this.selectEndIndex = rowNode.rowIndex;
+                    this.enableIndexSelection = true;
```
PatrickTasse, line 610: I'm not sure this is needed. You can just check if selectStartIndex and selectEndIndex are set?

packages/react-components/src/components/table-output-component.tsx
```diff
@@ -685,13 +686,16 @@ export class TableOutputComponent extends AbstractOutputComponent<TableOutputPro
         csv_data.push(colHeaderRow.join(','));
 
         if (option === 'selection') {
-            if (this.startTimestamp && this.endTimestamp) {
+            if (this.enableIndexSelection && this.selectStartIndex !== -1 && this.selectEndIndex !==-1) {
+                fetchIndex = Math.min(this.selectStartIndex, this.selectEndIndex);
+                totalLinesToFetch = Math.abs(this.selectStartIndex - this.selectEndIndex) + 1;
+            } else if (!this.enableIndexSelection && this.startTimestamp && this.endTimestamp) {
```
PatrickTasse, line 692: I don't see any need at all to have time-based export from the table?

packages/react-components/src/components/table-output-component.tsx
PatrickTasse: Should it not be index-based immediately? Not sure if you are going to merge commits of this PR?

packages/react-components/src/components/table-output-component.tsx
```diff
@@ -653,4 +655,126 @@ export class TableOutputComponent extends AbstractOutputComponent<TableOutputPro
...
+            let curLinesToFetch = bufferSize;
+
+            if (fetchLinesRemainder > bufferSize) {
+                fetchLinesRemainder-=bufferSize;
+            } else {
+                curLinesToFetch = fetchLinesRemainder;
+                fetchLinesRemainder = 0;
+            }
+
+            const tspClientResponse = await tspClient.fetchTableLines(traceUUID, outputId, QueryHelper.tableQuery(this.columnIds, fetchIndex, curLinesToFetch,
```
PatrickTasse, line 715: The fetchIndex needs to be incremented in the loop! It's always fetching the first 10000 lines at every iteration...

If you have fetchIndex properly used, you might not need fetchLinesRemainder.

packages/react-components/src/components/table-output-component.tsx
```diff
@@ -653,4 +655,126 @@ export class TableOutputComponent extends AbstractOutputComponent<TableOutputPro
...
+        link.style.display = 'none';
+        document.body.appendChild(link);
+        link.click();
+        document.body.removeChild(link);
+    }
+
+    protected showExportMenu(): React.ReactNode {
+        return <React.Fragment>
+            <ul>
+                <li className='drop-down-list-item' key={0} onClick={() => this.exportOutput('table')}>Export table to csv</li>
```
PatrickTasse, line 773: The menu remains open after being clicked while the slow operation is ongoing. This can be really bad because there is no feedback to the user, if the user clicks again, there are 2, 3, .... loops in parallel exporting the full trace, it can bring the UI to a halt...

At the very least the menu should be closed. But even so, the user might open and click it again while the previous operation is still ongoing. I would have expected a progress dialog, maybe with a Cancel button.

packages/react-components/src/components/table-output-component.tsx
```diff
@@ -653,4 +655,126 @@ export class TableOutputComponent extends AbstractOutputComponent<TableOutputPro
...
+                csv_data.push(csvrow.join(','));
+            }
+        }
+
+        // combine each row data with new line character
+        const tableString = csv_data.join('\n');
+
+        const link = document.createElement('a');
+        link.setAttribute('href', `data:text/csv;charset=utf-8,${encodeURIComponent(tableString)}`);
+        link.setAttribute('download', 'export');
```
PatrickTasse, line 763: I would have expected a file chooser dialog for the user to decide where to store it and what to call it. Maybe for future improvement. But otherwise perhaps the name should be computed from the trace name? Now it's export (1), export (2), etc...

PatrickTasse (changes requested):

packages/react-components/src/components/trace-context-component.tsx
```diff
@@ -435,7 +435,7 @@ export class TraceContextComponent extends React.Component<TraceContextProps, Tr
                     unitController: this.unitController,
                     outputWidth: this.state.style.width,
                     backgroundTheme: this.state.backgroundTheme,
-                    setChartOffset: this.setChartOffset
+                    setChartOffset: this.setChartOffset,
```
PatrickTasse, line 438: Were you trying to add experimentName here? Would it be better instead of adding only for specific outputs?

packages/react-components/src/components/table-output-component.tsx
```diff
@@ -653,4 +655,143 @@ export class TableOutputComponent extends AbstractOutputComponent<TableOutputPro
...
+
+        this.closeExportMenu();
+
+        const traceUUID = this.props.traceId;
+        const tspClient = this.props.tspClient;
+        const outputId = this.props.outputDescriptor.id;
+        let fetchIndex = 0;
+        let totalLinesToFetch = this.gridApi.getDisplayedRowCount();
+        const linesArray = new Array<Line>();
+        const bufferSize = 100000;
```
PatrickTasse, line 672: I suggest 50,000 so that it's closer to 5 seconds per iteration, but of course that depends on many factors, e.g. parsing speed in trace-server, network bandwidth, number of columns and length of the cell data...

packages/react-components/src/components/table-output-component.tsx
```diff
@@ -653,4 +655,143 @@ export class TableOutputComponent extends AbstractOutputComponent<TableOutputPro
...
+                    // a row and push it to csvrow
+                    const cellContent = lines[i].cells[j].content;
+                    if (cellContent.includes(',')) {
+                        csvrow.push('"' + cellContent + '"');
+                    } else {
+                        csvrow.push(cellContent);
+                    }
+                }
+                // Combine each column value with comma
+                csv_data.push(csvrow.join(','));
```
PatrickTasse, line 766: It's not acceptable to accumulate all rows in memory, it can easily crash the browser with a large trace. The partial data should be appended to the target file at each iteration and discarded from memory.

PatrickTasse (changes requested):

PR #746, opened 2022-05-06
Select chart components from the events table

Rodrigoplp-work:
Select corresponding chart components from the events table.

Clicking on a line in the events table will cause a chart like "Thread
Status" to scroll vertically to the row which has the element that
corresponds to the clicked event.

Only works if the events contains matching information about start time,
PID, TID or unique process name.

This PR requires the following pull requests:

(Link to Incubator PR, TBD)
https://github.com/eclipse-cdt-cloud/tsp-typescript-client/pull/62

https://user-images.githubusercontent.com/88502808/167218470-4d17f54a-1fc8-436a-b5bc-aeea0a34cbde.mp4


packages/react-components/src/components/timegraph-output-component.tsx
```diff
@@ -288,6 +298,59 @@ export class TimegraphOutputComponent extends AbstractTreeOutputComponent<Timegr
         }
     }
 
+    private findElement(payload: { [key: string]: string | { [key: string]: string }}): TimeGraphEntry | number | undefined {
+        const startTimestampString: string = payload['startTimestamp'].toString();
+        const found = this.state.timegraphTree.findIndex(e => e.start === BigInt(startTimestampString));
```
bhufmann, line 303: The entry start time in not relevant in most cases not matching the selected event time. So, you can remove it.

packages/react-components/src/components/timegraph-output-component.tsx
```diff
@@ -288,6 +298,59 @@ export class TimegraphOutputComponent extends AbstractTreeOutputComponent<Timegr
...
+        if (timeFilter.length === 1) {
+            // Found unique timestamp match
+            const row = this.state.dataRows.findIndex(e => e.id === timeFilter[0].id);
+            if (row) {
+                return row;
+            }
+        }
+        else if (found === -1 && typeof(payload.load) !== 'string') {
+            // Parse identifiers from "Contents" field
+            const contents = payload.load.Contents.split(', ');
```
bhufmann, line 315: So, your trying to match the Content column with the metadata. However, the idea of the this feature was to match the metadata with the columns of the event in the table, where the key is the header name and value the content of the cell. So, what you need to do is find the most intersecting row, i.e. most number of matching key/value pairs from metadata and event header/cell pairs. To test I tried to implement it with the following code:
```javascript
    private doHandleSelectionChangedSignal(payload: { [key: string]: string }) {
        const offset = this.props.viewRange.getOffset() || BigInt(0);
        const startTimestamp = payload['startTimestamp'];
        const endTimestamp = payload['endTimestamp'];
        if (startTimestamp !== undefined && endTimestamp !== undefined) {
            const selectionRangeStart = BigInt(startTimestamp) - offset;
            const selectionRangeEnd = BigInt(endTimestamp) - offset;
            const foundElement = this.findElement(payload);
            // Scroll vertically
            if (foundElement) {
                const rowIndex = this.state.dataRows.findIndex(d => d.id === foundElement.id);
                if (this.timeGraphTreeRef.current) {
                    this.timeGraphTreeRef.current.scrollTop = 20 * rowIndex;
                }
                // Select row
                const row = this.state.dataRows.find(d => d.id === foundElement.id);
                if (row) {
                    this.chartLayer.selectRow(row);
                    // TODO select state at row/starttime
                }
            }
            // Scroll horizontally
            this.props.unitController.selectionRange = {
                start: selectionRangeStart,
                end: selectionRangeEnd
            };
            this.chartCursors.maybeCenterCursor();
        }
    }

    private findElement(payload: { [key: string]: string | { [key: string]: string }}): TimeGraphEntry | undefined {
        let element: TimeGraphEntry | undefined = undefined;
        let max = 0;
        if (payload && payload.load) {
            this.state.timegraphTree.forEach(el => {
                if (el.metadata) {
                    let cnt = 0;
                    Object.entries(el.metadata).some(([key, value]) => {
                        if (typeof (payload.load) !== 'string') {
                            const val = payload.load[key];
                            if (val !== undefined) {
                                const num = Number(val);
                                if ((num !== undefined && num === value[0]) || (val === value[0])) {
                                    cnt++;
                                }
                            }
                        }
                    });
                    if (cnt > max) {
                        max = cnt;
                        element = el;
                    }
                }
            });
        }
        return element;
    }
```
The is probably a better way to implement the search algorithm `findElement` with some stream and lambdas and some javascript magic which you are more familiar that me. Feel free to reimplement the ```findElement``` method.

I tested my code with the `Thread Status` and `Resources` view.

bhufmann (changes requested):
Thanks for the contribution. It was fun to see this useful feature getting shape. I have request for change in the search algorithm for the row.

packages/react-components/src/components/timegraph-output-component.tsx
Rodrigoplp-work: Thank you for the clarification and suggestions!
I tried a reducer in place of the `forEach` in `findElement`, but the solution was not much shorter than yours, just more abstract. So I preferred to kept your solution with a few minor adjustments, and cleaned up unused code.

Rodrigoplp-work: Thank you for the clarification and suggestions!
I tried a reducer in place of the `forEach` in `findElement`, but the solution was not much shorter than yours, just more abstract. So I preferred to kept your solution with a few minor adjustments, and cleaned up unused code.

packages/react-components/src/components/timegraph-output-component.tsx
Rodrigoplp-work: Done.

Rodrigoplp-work: Done.

bhufmann (changes requested):

packages/react-components/src/components/timegraph-output-component.tsx
```diff
@@ -27,7 +27,6 @@ import hash from 'traceviewer-base/lib/utils/value-hash';
 import ColumnHeader from './utils/filter-tree/column-header';
 import { TimeGraphAnnotationComponent } from 'timeline-chart/lib/components/time-graph-annotation';
 import { Entry } from 'tsp-typescript-client';
-import { isEqual } from 'lodash';
```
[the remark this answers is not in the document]
Rodrigoplp-work: True. Fixed.

packages/react-components/src/components/timegraph-output-component.tsx
```diff
@@ -293,6 +307,35 @@ export class TimegraphOutputComponent extends AbstractTreeOutputComponent<Timegr
         }
     }
 
+    private findElement(payload: { [key: string]: string | { [key: string]: string }}): TimeGraphEntry | undefined {
+        let element: TimeGraphEntry | undefined = undefined;
```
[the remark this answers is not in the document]
Rodrigoplp-work: Good idea! Done.

packages/react-components/src/components/timegraph-output-component.tsx
```diff
@@ -285,6 +286,19 @@ export class TimegraphOutputComponent extends AbstractTreeOutputComponent<Timegr
         if (startTimestamp !== undefined && endTimestamp !== undefined) {
             const selectionRangeStart = BigInt(startTimestamp) - offset;
             const selectionRangeEnd = BigInt(endTimestamp) - offset;
+            const foundElement = this.findElement(payload);
+            // Scroll vertically
+            if (foundElement) {
+                const rowIndex = this.state.dataRows.findIndex(d => d.id === foundElement.id);
```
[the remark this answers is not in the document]
Rodrigoplp-work: Ok.

bhufmann:
This PR is moved to #760.

PR #747, opened 2022-05-09
trace-viewer: Remove onAfterShow causing event loop

### LINKED ISSUE #736: Endless change of focus of trace viewer tabs when tabs are beside each other
After opening 2 traces, the trace viewer tabs are stacked behind each other. When moving one of the traces beside the other so that both are visible, then each tab gets focus and loses focus endlessly (see below). Because of this endless change of focus it's hard to  interact with one tab. The focus on the tab should be set by selection to one tab.

bouncing-trace-container

This needs to be fixed as per-requisite for #660.

marco-miller:
Before this change, opening two traces then moving one in its own panel
caused an infinite focus loop between these two. That event loop seemed
to have been caused by onAfterShow firing an extra event during that
move action.

Removing that overriding method in TraceViewerWidget fixes that event
loop, lending the moved trace tab in focus as expected (in its own
panel). No evidence of requiring that method in other cases was found.

Fixes #736.

### DIFF
--- a/theia-extensions/viewer-prototype/src/browser/trace-viewer/trace-viewer.tsx
+++ b/theia-extensions/viewer-prototype/src/browser/trace-viewer/trace-viewer.tsx
@@ -284,13 +284,6 @@ export class TraceViewerWidget extends ReactWidget implements StatefulWidget {
         }
     }
 
-    onAfterShow(msg: Message): void {
-        super.onAfterShow(msg);
-        if (this.openedExperiment) {
-            signalManager().fireTraceViewerTabActivatedSignal(this.openedExperiment);
-        }
-    }
-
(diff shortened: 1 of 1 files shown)


marco-miller:
I've been trying to un-Assign this from me (did so by mistake, thinking of the Issue). But GH seems like not caching or setting this properly so far.

marco-miller:
> I've been trying to un-Assign this from me (did so by mistake, thinking of the Issue). But GH seems like not caching or setting this properly so far.

Fixed now it seems, sorry for the hassle.

bhufmann (approved):
Thanks for fixing this. The event loop doesn't happen anymore. I've tested that the viewer tabs get focus when a new trace is opened or selected.

PR #748, opened 2022-05-12
table: Fix time sync. when searching previous and next

### LINKED ISSUE #743: Moving through filtered events in events table does not change selection in xy-chart / timeline-chart.
To replicate:
1) Open an XY Chart, a Timeline Chart, and an Events Table view.
2) Select an event in the Events Table.
3) Filter a column in the events table (use the spyglass icon).
4) While focused on the filter textbox, press enter to move to the next instance that meets search criteria.

Expected behavior: When you press enter / move to next search item, the blue selection marker moves on xy-chart and timeline-chart components.

166261536-eb1b10fd-b99c-4622-9447-390890f4900d

bhufmann:
fixes #743

### DIFF
--- a/packages/react-components/src/components/table-output-component.tsx
+++ b/packages/react-components/src/components/table-output-component.tsx
@@ -8,7 +8,7 @@ import { cloneDeep } from 'lodash';
 import { signalManager } from 'traceviewer-base/lib/signals/signal-manager';
 import { TimelineChart } from 'timeline-chart/lib/time-graph-model';
 import { CellKeyDownEvent } from 'ag-grid-community/dist/lib/events';
-import { Line } from 'tsp-typescript-client/lib/models/table';
+import { Line, TableModel } from 'tsp-typescript-client/lib/models/table';
 import { SearchFilterRenderer, CellRenderer, LoadingRenderer } from './table-renderer-components';
 import { ResponseStatus } from 'tsp-typescript-client';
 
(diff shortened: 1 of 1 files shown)


packages/react-components/src/components/table-output-component.tsx
```diff
@@ -609,6 +609,10 @@ export class TableOutputComponent extends AbstractOutputComponent<TableOutputPro
                 if (currRowIndexFound && !isFound && rowNode.data && rowNode.data['isMatched']) {
                     this.gridApi?.ensureIndexVisible(rowNode.rowIndex);
                     this.selectStartIndex = this.selectEndIndex = rowNode.rowIndex;
+                    if (this.timestampCol) {
+                        this.startTimestamp = BigInt(rowNode.data[this.timestampCol]);
+                        this.endTimestamp = BigInt(rowNode.data[this.timestampCol]);
+                    }
```
PatrickTasse, line 615: It only works if the match is found in the current cache. Otherwise it find the index below and ensures it's made visible. We probably need for the lines to be loaded before we can update the selected time and fire the selection changed signal.

PatrickTasse (changes requested):

packages/react-components/src/components/table-output-component.tsx
bhufmann: Thanks for finding this.

PatrickTasse (approved):

PR #750, opened 2022-05-17
Export datatree outputs to csv

hriday-panchasara:
Export datatree-output-component to csv
Screen Shot 2022-05-17 at 1 15 33 AM
- Expanded options menu to show 'Export to CSV' in datatree-output-component. Aim is to add 'Export to CSV' options to table-output-component's options menu as well as pin/unpin option to abstract-output-component's (Base class) option menu
- Trace name given to downloaded file

### DIFF
--- a/packages/react-components/src/components/abstract-output-component.tsx
+++ b/packages/react-components/src/components/abstract-output-component.tsx
@@ -16,6 +16,7 @@ export interface AbstractOutputProps {
     tooltipComponent: TooltipComponent | null;
     tooltipXYComponent: TooltipXYComponent | null;
     traceId: string;
+    traceName?: string;
     range: TimeRange;
     nbEvents: number;
     viewRange: TimeRange;
@@ -43,14 +44,15 @@ export interface AbstractOutputProps {
 export interface AbstractOutputState {
     outputStatus: string;
     styleModel?: OutputStyleModel;
-    optionsDropdownOpen?: boolean;
+    optionsDropdownOpen: boolean;
+    additionalOptions?: boolean;
 }
 
 export abstract class AbstractOutputComponent<P extends AbstractOutputProps, S extends AbstractOutputState> extends React.Component<P, S> {
 
     private readonly DEFAULT_HANDLE_WIDTH = 30;
 
-    private mainOutputContainer: React.RefObject<HTMLDivElement>;
+    protected mainOutputContainer: React.RefObject<HTMLDivElement>;
 
     private optionsMenuRef: React.RefObject<HTMLDivElement>;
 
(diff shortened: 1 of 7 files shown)


PatrickTasse (approved):

packages/react-components/src/components/datatree-output-component.tsx
```diff
@@ -188,4 +190,51 @@ export class DataTreeOutputComponent extends AbstractOutputComponent<AbstractOut
...
+                        }
+                        row.push(data);
+                    }
+                    csvArray.push(row.join(','));
+                }
+                const tableString = csvArray.join('\n');
+
+                const link = document.createElement('a');
+                link.setAttribute('href', `data:text/csv;charset=utf-8,${encodeURIComponent(tableString)}`);
+                link.setAttribute('download', this.props.traceName ?? 'export');
```
PatrickTasse, line 221: Might be useful to differentiate the outputs in the file name, e.g. 'traceName - outputName.csv'

PatrickTasse (approved):

packages/react-components/src/components/abstract-output-component.tsx
```diff
@@ -43,14 +44,15 @@ export interface AbstractOutputProps {
...
+    optionsDropdownOpen: boolean;
+    additionalOptions?: boolean;
 }
 
 export abstract class AbstractOutputComponent<P extends AbstractOutputProps, S extends AbstractOutputState> extends React.Component<P, S> {
 
     private readonly DEFAULT_HANDLE_WIDTH = 30;
 
-    private mainOutputContainer: React.RefObject<HTMLDivElement>;
+    protected mainOutputContainer: React.RefObject<HTMLDivElement>;
```
bhufmann, line 55: should we have a protected getter method?

packages/react-components/src/components/datatree-output-component.tsx
```diff
@@ -188,4 +190,51 @@ export class DataTreeOutputComponent extends AbstractOutputComponent<AbstractOut
...
+
+                document.body.removeChild(link);
+            }
+        }
+    }
+
+    protected showAdditionalOptions(): React.ReactNode {
+        return <React.Fragment>
+            <ul>
+                <li className='drop-down-list-item' key={0} onClick={() => this.exportOutput()}>Export to csv</li>
```
bhufmann, line 236: The menus in the output components are currently hard-coded because there are only a few options. But more and more will be added soon and we should have a generic way to provide menu and commands. The advantage will be to decouple the implementation from the output classes itself as well as it easier to customize or change commands. Please backlog this task.

The menu has a different look and feel than the rest of the application. It should be possible the css style it so that we can re-use the theia styles for it.

bhufmann (commented):
Thanks for this feature. It works well. I have some suggestion for follow-up PRs. Also some unit tests would be good in a follow-up PR.

bhufmann (approved):
Looks good to me.

Issue #662, opened 2022-01-28
Remember and communicate analysis results (annotation)


ebugden:
Allow annotation at a point in time

PR #759, opened 2022-06-02
Improve filter text box discoverability

hoangphamEclipse:
This commit makes the text box for filter in the filter-tree more noticeable and recognizable as a text box by adding rounded borders and a filter-icon.

Resolve: [Issue #662](https://github.com/eclipse-cdt-cloud/theia-trace-extension/issues/622)


marco-miller (changes requested):
Tests also fail locally for me, as expected [per above](https://github.com/eclipse-cdt-cloud/theia-trace-extension/runs/6710744169?check_suite_focus=true).

marco-miller (approved):
LGTM in both default themes (light/dark). Will add @PatrickTasse as co-reviewer just in case.

PatrickTasse (approved):

PR #767, opened 2022-06-17
Fix css positioning for filter text box icon

hoangphamEclipse:
Currently, the filter icon of the filter text box stays fixed at the top left of the filter tree, even when the filter text box is out of sight when users scroll down the tree. This commit fixes this issue by setting the position css property of the parent div of the filter text box to relative, so that the filter icon is positioned relatively to the parent container and stays attached to the filter text box.

This bug is related to [PR 759](https://github.com/eclipse-cdt-cloud/theia-trace-extension/pull/759).


bhufmann (approved):
Thanks for fixing this. Looks good to me.

PatrickTasse (approved):

PR #779, opened 2022-07-06
Fire TSP client requests in parallel in getData function

williamsyang-work:
Requests for States, Annotations, and Arrows are now fired simultaneously.

Annotation response data processing moved in front of states processing, since annotation response time is usually shorter.

### DIFF
--- a/packages/react-components/src/components/data-providers/tsp-data-provider.ts
+++ b/packages/react-components/src/components/data-providers/tsp-data-provider.ts
@@ -8,6 +8,7 @@ import { QueryHelper } from 'tsp-typescript-client/lib/models/query/query-helper
 import { OutputElementStyle } from 'tsp-typescript-client/lib/models/styles';
 import { Annotation, Type } from 'tsp-typescript-client/lib/models/annotation';
 import { TimeRange } from 'traceviewer-base/lib/utils/time-range';
+import { GenericResponse, TspClientResponse } from 'tsp-typescript-client';
 
 enum ElementType {
     STATE = 'state',
@@ -15,7 +16,6 @@ enum ElementType {
 }
 
 export class TspDataProvider {
-
     private client: TspClient;
     private outputId: string;
     private traceUUID: string;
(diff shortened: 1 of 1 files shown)


MatthewKhouzam (dismissed):
LGTM, but I asked for it.

packages/react-components/src/components/data-providers/tsp-data-provider.ts
```diff
@@ -48,106 +49,113 @@ export class TspDataProvider {
...
-                categoryArray.forEach(annotation => {
-                    if (annotation.type === Type.CHART) {
-                        if (annotation.entryId === -1) {
-                            rangeEvents.push(this.getAnnotation(category, annotation, rangeEvents.length, chartStart));
-                        } else {
-                            let entryArray = annotations.get(annotation.entryId);
-                            if (entryArray === undefined) {
-                                entryArray = [];
-                                annotations.set(annotation.entryId, entryArray);
+            const tspClientAnnotationsResponse = await annotationsPromise;
```
bhufmann, line 86: when executing this await, all 3 created promises (for annotations, states and arrows) are triggered and the back-end requests are done. With this all 3 are sent in parallel. However, we should make it more explicit and schedule all 3 executions using Promise.all() or Promise.allSettled().

bhufmann (changes requested):

bhufmann:
@williamsyang-work any news on this one?

packages/react-components/src/components/data-providers/tsp-data-provider.ts
```diff
@@ -60,48 +56,38 @@ export class TspDataProvider {
...
-        const rangeEvents: TimelineChart.TimeGraphAnnotation[] = [];
+        const annotationsPromise = this.client.fetchAnnotations(this.traceUUID, this.outputId, annotationParams);
+
+        const arrowStart = viewRange.start + this.timeGraphEntries[0].start;
+        const arrowEnd = viewRange.end + this.timeGraphEntries[0].start;
+        const fetchParameters = QueryHelper.timeRangeQuery(arrowStart, arrowEnd, nbTimes);
+        const arrowsPromise = this.client.fetchTimeGraphArrows(this.traceUUID, this.outputId, fetchParameters);
+
+        // Wait for responses
+        const [tspClientAnnotationsResponse, tspClientStatesResponse, tspClientArrowsResponse] = [await annotationsPromise, await statesPromise, await arrowsPromise];
```
bhufmann, line 83: Use the following Promise.all() to make sure that the 3 promises are triggered at the same time:
`const [tspClientAnnotationsResponse, tspClientStatesResponse, tspClientArrowsResponse] = await Promise.all([annotationsPromise, statesPromise, arrowsPromise]);`

bhufmann (changes requested):

bhufmann (approved):
Looks good to me. Thanks for this improvement.

MatthewKhouzam (approved):
Looks good to me!

PR #782, opened 2022-07-08
Persist state of view after pin/unpin

### LINKED ISSUE #762: Persist state of pinned view
When pinning/unpinning a view, the xy and timeline charts lose state of their tree selections. Would be nice to persist the state of these charts when they are re-rendered due to pin/unpin action

hriday-panchasara:
- This is an improvement of the feature in PR #739 
- When a view is pinned/unpinned it is re-rerendered onto the layout as it is moving to a different container. This re-rendering does not happen when dragging views since they are re-arranged in the same container. So when a view is pinned/unpinned, the charts lose their state and the user has to redo the selections.
- This patch tries to persist the state of the view that's being pinned and unpinned. Luckily, we are able to pass the state of the chart as a payload when firing the signal to pin/unpin a view.
- This is being used by the xy-charts to pass the checkedSeries and collapsedNodes of the tree, and the timegraph-charts to pass the collapsedNodes and collapsedMarkerNodes.
- By doing so we ensure that the chart state stays the same and the user does not have to re-do their selections
- Unfortunately I could not find a way to persist the scroll position of the tree or the order of tree columns of the xy chart. If there are any suggestions I'll try them out.
- The xy-output-component currently does not support data-provider styles and randomly assigns series colors. Therefore, sometimes after pin/unpin I've noticed minor changes in the overall color scheme. We'll be working on implementing data-provider styles for xy in the near future, so I'm assuming this issue would not pose as a blocker.

Fixes #762

### DIFF
--- a/packages/base/src/signals/signal-manager.ts
+++ b/packages/base/src/signals/signal-manager.ts
@@ -26,9 +26,11 @@ export declare interface SignalManager {
     fireTraceServerStartedSignal(): void;
     fireUndoSignal(): void;
     fireRedoSignal(): void;
-    firePinView(output: OutputDescriptor): void;
-    fireUnPinView(): void;
     fireOpenOverviewOutputSignal(): void;
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    firePinView(output: OutputDescriptor, payload?: any): void;
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    fireUnPinView(output: OutputDescriptor, payload?: any): void;
 }
 
 export const Signals = {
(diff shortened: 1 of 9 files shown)


packages/base/src/signals/signal-manager.ts
```diff
@@ -26,8 +26,8 @@ export declare interface SignalManager {
     fireTraceServerStartedSignal(): void;
     fireUndoSignal(): void;
     fireRedoSignal(): void;
-    firePinView(output: OutputDescriptor): void;
-    fireUnPinView(): void;
+    firePinView(payload: { output: OutputDescriptor, checkedSeries?: number[], collapsedNodes?: number[], collapsedMarkerNodes?: number[]}): void;
```
PatrickTasse, line 29: It feels like the persisted state of outputs should be opaque as it can be different for every output (e.g. there are no checkedSeries in timegraph). It doesn't feel right to have a merge of all possible state properties here.

I would perhaps keep the output and persistedState as separate parameters in the signal. I'm not sure if the persisted state can be an 'any' object or a choice of state classes defined in each specific output.

But I think signal-manager, abstract-output-component and trace-context-component should pass the persisted state as a 'black box' object.

PatrickTasse (changes requested):

packages/base/src/signals/signal-manager.ts
hriday-panchasara: Thanks for the suggestion Patrick, I've added another commit which passes the persisted state as a 'black-box' object. I was able to specify the persisted state as an 'any' object with the help of disable es-lint statements.

bhufmann (changes requested):
It needs a rebase. Heads-up that the merge of xy-output-component will be non-trivial due to the overview addition.

I tested it and it works fine. I wonder if you could persist the `selectedRow` of the timegraphs?

bhufmann (dismissed):
It looks good to me. Nice improvement. Thanks.

I'm going to open a follow-up issue to track other things that could be persisted as well, and would delay this PR to implement.

bhufmann:
Note that there is a network issue that prevents a successful  build. I'll re-trigger build once it's back to normal.

hriday-panchasara:
@bhufmann Thanks for the reviews! I've added another commit to persist the selected rows of the events table as well. For the selectedRow of timegraphs, it would be possible to persist once PR #788 is merged

bhufmann:
> @bhufmann Thanks for the reviews! I've added another commit to persist the selected rows of the events table. For the selectedRow of timegraphs, it would be possible to persist once PR #788 is merged

Sounds good. I'm ok to merge this PR and adding the selected row for time graphs afterwards.

bhufmann (approved):

PatrickTasse (approved):

PR #783, opened 2022-07-08
Fix issues with server status indicator

### LINKED ISSUE #763: Server status indicator always shows online when starting trace-extension
The server status indicator is supposed to show an error when the server fails to respond to a request. But when starting the trace-extension the server status icon always shows online regardless of server status. In the following example, I've started the trace-extension without running the trace-server:
Screen Shot 2022-06-14 at 11 54 41 AM

The network log indicates that the initial call to the /experiments endpoint fails:
Screen Shot 2022-06-14 at 11 55 06 AM

The expected behavior is for the server status to catch the above error and show offline on startup.

It looks like the component needs to re-render to show the error:
Screen Recording 2022-06-14 at 12 04 06 PM

hriday-panchasara:
I found the following issues with the current server status indicator:
1) Status indicator is incorrect when the trace-viewer is started without connection to the server:
Screen Shot 2022-07-08 at 2 52 35 PM

2) Status indicator is incorrect when toggling the trace-viewer left-hand panel (trace-explorer-widget):
Screen Recording 2022-07-08 at 2 53 50 PM

When looking into the code, the connectionStatusListener defined in PR #671 works as expected, but it is not always able to modify the indicator through JavaScript. This might be because the indicator is not an icon but a Theia Command. Registering a toolbar command allowed us to place the icon in Trace Explorer widget's title bar, but the statusListener fails to always update the icon. A better contribution point would be to place the icon as a child of the Trace Explorer widget. The following is how the status indicator looks like in this patch:
Screen Shot 2022-07-08 at 2 56 24 PM copy

I would like some suggestions on the UI and possible alternatives on how to present the status indicator

fixes #763

### DIFF
--- a/packages/react-components/style/trace-explorer.css
+++ b/packages/react-components/style/trace-explorer.css
@@ -286,7 +286,19 @@
     justify-content: flex-end !important;
 }
 
-#trace\.viewer\.serverCheck {
-    cursor: default;
-    color: green;
+.server-status-header {
+    background: var(--theia-sideBarSectionHeader-background);
+    line-height: var(--theia-view-container-title-height);
+    z-index: 10;
+    color: var(--theia-sideBarSectionHeader-foreground);
+    margin-bottom: 2px;
+    padding-left: 23px;
+    overflow: hidden;
+}
+
+#server-status-id {
+    float: right;
+    margin-top: 4px;
+    margin-bottom: 2px;
+    margin-right: 3px;
 }
--- a/theia-extensions/viewer-prototype/src/browser/trace-explorer/trace-explorer-sub-widgets/trace-explorer-placeholder-widget.tsx
+++ b/theia-extensions/viewer-prototype/src/browser/trace-explorer/trace-explorer-sub-widgets/trace-explorer-placeholder-widget.tsx
@@ -8,7 +8,7 @@ import { OpenTraceCommand } from '../../trace-viewer/trace-viewer-commands';
 export class TraceExplorerPlaceholderWidget extends ReactWidget {
 
     static ID = 'trace-explorer-placeholder-widget';
-    static LABEL = 'Trace Exploerer Placeholder Widget';
+    static LABEL = 'Trace Explorer Placeholder Widget';
 
     state = {
         loading: false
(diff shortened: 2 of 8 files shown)


bhufmann:
> This PR needs a rebase to latest in master branch, because there are merge conflicts.
> 
> I have 2 questions/suggestions:
> 
>     * Would it be possible to right align the icon? Maybe I'm just used to have it on the right side.
> 
>     * Would it be a good idea to allow for hiding the `Server Status` widget, like the other widgets in the explorer?
> 
> 
> Just some food for thoughts. Please let me know.

@hriday-panchasara any comments to these questions?

hriday-panchasara:
> > This PR needs a rebase to latest in master branch, because there are merge conflicts.
> > I have 2 questions/suggestions:
> > ```
> > * Would it be possible to right align the icon? Maybe I'm just used to have it on the right side.
> > 
> > * Would it be a good idea to allow for hiding the `Server Status` widget, like the other widgets in the explorer?
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Just some food for thoughts. Please let me know.
> 
> @hriday-panchasara any comments to these questions?
> 
> Also, I just noticed that the status indicator is still green after executing `Stop trace server` command.

@bhufmann Sorry about that, was not ready for re-review at the time.
- I fixed the indicator so that it responds correctly to the `Start trace server` and `Stop trace server` commands.
- The icon is right-aligned:
Screen Shot 2022-07-26 at 5 40 37 PM
- I was unable to find a good way to hide the Server Status Widget. I could add the widget to Trace Explorer's ViewContainer which allows to collapse/expand widget content, but it didn't look great when trying to toggle an empty widget:
[Screen Recording 2022-07-26 at 5.47.36 PM.webm](https://user-images.githubusercontent.com/92893187/181125390-acedba3e-bcfb-47e2-a957-357c878868bf.webm)

bhufmann:
> * I was unable to find a good way to hide the Server Status Widget. I could add the widget to Trace Explorer's ViewContainer which allows to collapse/expand widget content, but it didn't look great when trying to toggle an empty widget:
>       [Screen Recording 2022-07-26 at 5.47.36 PM.webm](https://user-images.githubusercontent.com/92893187/181125490-960f20cd-d425-4e0e-a1d6-0ba056deddeb.webm)

Thanks for looking into it. It looks like the current implementation works better, because the status is also visible when the placeholder is shown. So, let's keep it as in your patch.

bhufmann (approved):
It looks good to me. Thanks for this improvement!

hoangphamEclipse (approved):
I tested the change and went through the code. Looks good to me!

PR #784, opened 2022-07-11
Fix theia-trace-extension table vertical scrollbar

KyrollosBekhet:
During the developpment of the segment store table data provider this issue was found.
image
In fact the vertical scrollbar was not adjusted to show the real size of the table.
In this PR, the issue is resolved by setting the size of the table to be the same as the number of elements displayed in the table.
MicrosoftTeams-image (2)

### DIFF
--- a/packages/react-components/src/components/table-output-component.tsx
+++ b/packages/react-components/src/components/table-output-component.tsx
@@ -51,6 +51,7 @@ export class TableOutputComponent extends AbstractOutputComponent<TableOutputPro
     private selectEndIndex = -1;
     private filterModel: Map<string, string> = new Map<string, string>();
     private dataSource: IDatasource;
+    private tableSize = 0;
 
     static defaultProps: Partial<TableOutputProps> = {
         cacheBlockSize: 200,
@@ -83,13 +84,13 @@ export class TableOutputComponent extends AbstractOutputComponent<TableOutputPro
                     this.fetchColumns = false;
                     await this.init();
                 }
-                const rowsThisPage = await this.fetchSearchTableLines(params.startRow, params.endRow - params.startRow);
+                const rowsThisPage = await this.fetchTableLines(params.startRow, params.endRow - params.startRow);
                 for (let i = 0; i < rowsThisPage.length; i++) {
                     const item = rowsThisPage[i];
                     const itemCopy = cloneDeep(item);
                     rowsThisPage[i] = itemCopy;
                 }
-                params.successCallback(rowsThisPage, this.props.nbEvents);
+                params.successCallback(rowsThisPage, this.tableSize);
             }
         };
         this.onEventClick = this.onEventClick.bind(this);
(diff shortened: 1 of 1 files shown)


KyrollosBekhet:
> I can't review yet, but please test the events table with filters, if you didn't already. :)

packages/react-components/src/components/table-output-component.tsx
```diff
@@ -491,6 +492,7 @@ export class TableOutputComponent extends AbstractOutputComponent<TableOutputPro
     }
```
PatrickTasse, line 492: I think it would be better to set the tableSize in this method just before calling modelToRow(), as it doesn't really belong in that method.

PatrickTasse (changes requested):

PatrickTasse (approved):

packages/react-components/src/components/table-output-component.tsx
```diff
@@ -83,13 +84,13 @@ export class TableOutputComponent extends AbstractOutputComponent<TableOutputPro
...
                 }
-                const rowsThisPage = await this.fetchSearchTableLines(params.startRow, params.endRow - params.startRow);
+                const rowsThisPage = await this.fetchTableLines(params.startRow, params.endRow - params.startRow);
                 for (let i = 0; i < rowsThisPage.length; i++) {
                     const item = rowsThisPage[i];
                     const itemCopy = cloneDeep(item);
                     rowsThisPage[i] = itemCopy;
                 }
-                params.successCallback(rowsThisPage, this.props.nbEvents);
+                params.successCallback(rowsThisPage, this.tableSize);
```
hoangphamEclipse, line 93: Can we use rowThisPage.size instead of introducing the new property this.tableSize?

KyrollosBekhet: The fetchTableLines fetches by a chunck of 200 rows at a time so the rowsThisPage.length will be 200. However the table may have more lines than 200. So unfortunately, your change request is not applicable.

hoangphamEclipse: Sounds good

bhufmann, line 93: It took me a while to understand why it was working on with events table and not with the segment store table. For the events table there is an indexing loop that checks for number of events and updates the nbEvents which then triggers the function call componentDidUpdate() on line 158 below. The componentDidUpdate() sets the infiniteRowCount which then updates the vertical scrollbar. When opening the table with a segment store there is no corresponding indexing loop that updates the nbEvents and the scrollbar.

The fix that you suggest, will work, only when the fetchTableLines call to fill the table initially returns the total number of segments i.e. that the back-end has the value available at that time. 

I think there should be a generic way to determine the table size for any virtual table data providers. I'll open a separate ticket for that. I'll merge this PR because it works-around the bug for segment store tables and it still works with the events table.

bhufmann (approved):

Issue #775, opened 2022-06-21
Provide output selector for overview view


bhufmann:
With the addition of the Overview view (#745) it provides an overview of the trace using the histogram data provider. This ticket is to trace the addition of an dialog box so that user can select the XY data provider to use in the Overview view, because some user's are more interested in other characteristics, for example the CPU usage.

Add an button to the Overviews toolbar or open a dialog when pressing the `Show Overview` button and provide a list of data providers where user can select the data provider using a radio button.

MatthewKhouzam:
This is done right?

bhufmann:
Yes this one can be closed.

PR #789, opened 2022-07-14
Add dialog to select output for trace overview

hoangphamEclipse:
Currently, the trace overview only works with the Histogram output. This commit adds a dialog that allows users to select different outputs for the trace overview.

This resolves issue #775.

### DIFF
--- a/packages/base/src/signals/signal-manager.ts
+++ b/packages/base/src/signals/signal-manager.ts
@@ -31,6 +31,7 @@ export declare interface SignalManager {
     firePinView(output: OutputDescriptor, payload?: any): void;
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     fireUnPinView(output: OutputDescriptor, payload?: any): void;
+    fireOverviewOutputSelectedSignal(payload: { traceId: string, outputDescriptor: OutputDescriptor}): void;
 }
 
 export const Signals = {
@@ -58,7 +59,8 @@ export const Signals = {
     TRACE_SERVER_STARTED: 'trace server started',
     PIN_VIEW: 'view pinned',
     UNPIN_VIEW: 'view unpinned',
-    OPEN_OVERVIEW_OUTPUT: 'open overview output'
+    OPEN_OVERVIEW_OUTPUT: 'open overview output',
+    OVERVIEW_OUTPUT_SELECTED: 'overview output selected'
 };
 
 export class SignalManager extends EventEmitter implements SignalManager {
@@ -135,6 +137,9 @@ export class SignalManager extends EventEmitter implements SignalManager {
     }
     fireOpenOverviewOutputSignal(): void {
         this.emit(Signals.OPEN_OVERVIEW_OUTPUT);
+    }
+    fireOverviewOutputSelectedSignal(payload: { traceId: string, outputDescriptor: OutputDescriptor}): void {
+        this.emit(Signals.OVERVIEW_OUTPUT_SELECTED, payload);
     }
 }
 
--- a/packages/react-components/src/components/utils/available-views-component.tsx
+++ b/packages/react-components/src/components/utils/available-views-component.tsx
@@ -1,0 +1,122 @@
+import { ListRowProps, AutoSizer, List } from 'react-virtualized';
+import React from 'react';
+import { OutputDescriptor } from 'tsp-typescript-client/lib/models/output-descriptor';
+
+export interface AvailableViewsComponentProps {
+    availableViewListKey: number,
+    outputDescriptors: OutputDescriptor[],
+    onContextMenuEvent?: (event: React.MouseEvent<HTMLDivElement, MouseEvent>, output: OutputDescriptor | undefined) => void,
+    onOutputClicked: (selectedOutput: OutputDescriptor) => void,
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    listRowWidth?: string,
+    listRowPadding?: string
+}
+
+export interface AvailableViewsComponentState {
+    lastSelectedOutputIndex: number;
+}
+
+export class AvailableViewsComponent  extends React.Component<AvailableViewsComponentProps, AvailableViewsComponentState>{
+    static LIST_MARGIN = 2;
+    static LINE_HEIGHT = 16;
+    static ROW_HEIGHT = (2 * AvailableViewsComponent.LINE_HEIGHT) + AvailableViewsComponent.LIST_MARGIN;
+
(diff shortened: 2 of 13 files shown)


hoangphamEclipse:
Fixed issue with multiple dialogs are opened when the user has multiple traces opened.

packages/react-components/src/components/xy-output-overview-component.tsx
```diff
@@ -334,4 +342,15 @@ export class XYOutputOverviewComponent extends AbstractXYOutputComponent<Abstrac
     protected getTitleBarTooltip(): string {
         return this.props.outputDescriptor.name + ' overview';
     }
+
+    protected showOptionsMenu(): React.ReactNode {
+        return <React.Fragment>
+            {<div className='options-menu-container'>
+                <button title="Select overview output source" className='options-menu-button'
+                onClick={()=>{signalManager().fireSelectOverviewOutputSignal(this.props.experiment);}}>
```
bhufmann, line 350: Shouldn't the onClick handler open the dialog and when the use select an XY output a signal is sent with the new output descriptor. You can use the tspClient that is passed to the output to fetch the available outputs. No need to pass the experiment to the class.

packages/react-components/src/components/xy-output-overview-component.tsx
```diff
@@ -334,4 +342,15 @@ export class XYOutputOverviewComponent extends AbstractXYOutputComponent<Abstrac
...
+                    <FontAwesomeIcon icon={faCog} />
```
bhufmann, line 351: I wonder if we should have the "hamburger menu" instead and have the config as an entry in the menu. Not sure though. What do you think?

bhufmann (changes requested):
Nice feature to allow users to select the overview data. Thanks.

packages/react-components/src/components/utils/trace-overview-selection-dialog-component.tsx
```diff
@@ -0,0 +1,117 @@
+import { ListRowProps, AutoSizer, List } from 'react-virtualized';
+import {  DialogProps } from '@theia/core/lib/browser/dialogs';
```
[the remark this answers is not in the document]
hoangphamEclipse: By having my own implementation, are you suggesting that I should inherit the AbstractDialog instead of ReactDialog?

packages/react-components/src/components/xy-output-overview-component.tsx
hoangphamEclipse: The reason why I did this was because in the trace-viewer, there is a handler that listens to the SelectOverviewOutputSignal for each opened trace. When we have multiple traces open, multiple dialogs will be opened by this listener. Thus, I use the experiment UUID to distinguish which experiment is opening the selection dialog. I can take a look at the code to see if there is a better option.

packages/react-components/src/components/xy-output-overview-component.tsx
bhufmann: True and I was aware of that.  In the future we might have more options for all the views. I want to avoid to have multiple buttons in the views, because of limited space. Other views (xy or timegrah) have the hamburger menu with only the "Pin view" in it at the moment. I'll get back to you about this.

packages/base/src/signals/signal-manager.ts
```diff
@@ -58,7 +59,8 @@ export const Signals = {
     TRACE_SERVER_STARTED: 'trace server started',
     PIN_VIEW: 'view pinned',
     UNPIN_VIEW: 'view unpinned',
-    OPEN_OVERVIEW_OUTPUT: 'open overview output'
+    OPEN_OVERVIEW_OUTPUT: 'open overview output',
+    OVERVIEW_OUTPUT_SELECTED: 'select overview output'
```
bhufmann, line 63: `select overview output` should be called now `'overview output selected`

packages/react-components/src/components/utils/available-views-component.tsx
```diff
@@ -0,0 +1,107 @@
+import { ListRowProps, AutoSizer, List } from 'react-virtualized';
```
bhufmann, line 1: This is a modified copy of `trace-explorer-views-widget.tsx`. Do you plan to refactor the code in a separate PR to optimize code re-use and remove code duplication? Please let me know.

theia-extensions/viewer-prototype/src/browser/trace-viewer/trace-viewer.tsx
```diff
@@ -66,11 +65,13 @@ export class TraceViewerWidget extends ReactWidget implements StatefulWidget {
 
     private onOutputAdded = (payload: OutputAddedSignalPayload): Promise<void> => this.doHandleOutputAddedSignal(payload);
     private onTraceOverviewOpened = (): Promise<void> => this.doHandleTraceOverviewOpenedSignal();
-    private onExperimentSelected = (experiment: Experiment): void => this.doHandleExperimentSelectedSignal(experiment);
+    private onSelectTraceOverviewOutput = (payload: {traceId: string, outputDescriptor: OutputDescriptor}): Promise<void> => this.doHandleSelectTraceOverviewOutput(payload);
```
bhufmann, line 68: `onSelectTraceOverviewOutput` should be called `onTraceOverviewOutputSelected`, and `doHandleSelectTraceOverviewOutput` should be called `doHandleTraceOverviewOutputSelected`, right?

bhufmann (changes requested):
Some more small comments. Also, please make styling the same as for the main menu for example. The color of rows are not the same when hovering over the rows.

packages/react-components/src/components/utils/available-views-component.tsx
hoangphamEclipse: Yes there is a PR that is in draft state for this refactoring. Once this one is approve, I will put that as ready for review.

theia-extensions/viewer-prototype/src/browser/trace-viewer/trace-viewer.tsx
hoangphamEclipse: Yes, I will update the name.

bhufmann (changes requested):
It looks good to me. It has merge conflicts, though. Could you please rebase this PR? Thanks!

hoangphamEclipse:
I rebased the code

hoangphamEclipse:
This is the refactoring change #790  for the previously mentioned common code of the available views component.

bhufmann (approved):
Looks good to me. Thanks!

MatthewKhouzam (approved):
Thanks! Added bonus: scatters work.

PR #794, opened 2022-07-21
XY Chart refresh data model on resize

### LINKED ISSUE #726: XY chart doesn't refresh its data model on resize
The XY chart doesn't react to resize events. When the chart is opened, or when the view range changes, it refreshes its data model.

If, at that moment, the chart width is 0, then it doesn't get a successful response from the trace-server and displays an error message in place of the chart: "Select a checkbox to see analysis results". When the chart is resized, this error message remains.

If, at that moment, the chart width is 5, it receives a data model from the trace-server with only 5 data points. When the chart is resized (e.g. to width=1000), it still displays only 5 data points.

The XY chart should refresh its data model when the chart width is changed. However this should be done through a debouncer to avoid overloading the server with many close requests.

hriday-panchasara:
xy chart updates data model when size of parent component (trace-context-component) is changed

[Screen Recording 2022-07-21 at 11.05.33 AM.webm](https://user-images.githubusercontent.com/92893187/180261591-0270e757-adf0-47e1-ba8e-3edf91792912.webm)

fixes #726

### DIFF
--- a/packages/react-components/src/components/abstract-xy-output-component.tsx
+++ b/packages/react-components/src/components/abstract-xy-output-component.tsx
@@ -218,7 +218,8 @@ export abstract class AbstractXYOutputComponent<P extends AbstractOutputProps, S
         const viewRangeChanged = this.props.viewRange !== prevProps.viewRange;
         const checkedSeriesChanged = this.state.checkedSeries !== prevState.checkedSeries;
         const collapsedNodesChanged = this.state.collapsedNodes !== prevState.collapsedNodes;
-        const chartWidthChanged = this.props.style.width !== prevProps.style.width || this.props.style.chartOffset !== prevProps.style.chartOffset;
+        const chartWidthChanged = this.props.style.width !== prevProps.style.width || this.props.style.chartOffset !== prevProps.style.chartOffset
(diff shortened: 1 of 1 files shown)


hoangphamEclipse (approved):
Thank you for the bug fix. I tested the change and it worked. The code also looks good.

hoangphamEclipse:
@hriday-panchasara  Looks good to me :)

hoangphamEclipse (approved):
Looks good to me! :)

bhufmann (changes requested):
The fetching of the data should use a debouncer to avoid too many back-end requests.

hriday-panchasara:
> The fetching of the data should use a debouncer to avoid too many back-end requests.
Yup, throttling is already implemented thanks to PR #713

Rodrigoplp-work (approved):
Well done. And it seems PR #713 does in fact handle the throttling problem.

bhufmann:
> Well done. And it seems PR #713 does in fact handle the throttling problem.

I should have expanded the lines below in the diff view then I would have seen the throttling :-)

bhufmann (approved):
Thanks!

PR #798, opened 2022-07-26
Add select all and deselect all in marker menu

### LINKED ISSUE #770: Add select all and deselect all in marker menu
The trace view has a menu to select which marker categories to visualize. Some domain specific data providers can have a long list of markers. A `Select all` and `Deselect all` menu item shall be added so that the use doesn't have click each category one by one.

image

hriday-panchasara:
In markers menu add menu item to select/deselect all

fixes #770


bhufmann (approved):
It looks good to me. Thanks!

PatrickTasse (approved):

PatrickTasse:
Suggestion, could be done now or later: Add a menu separator or divider between Select All and the other items.

hriday-panchasara:
Looks like Theia does not provide the functionality to add classNames when registering menu actions - only label and iconClass are customizable. We could add an iconClass and modify its css, but not sure if it would be as effective:
Screen Shot 2022-08-02 at 5 55 27 PM

bhufmann:
> Looks like Theia does not provide the functionality to add classNames when registering menu actions - only label and iconClass are customizable. We could add an iconClass and modify its css, but not sure if it would be as effective: Screen Shot 2022-08-02 at 5 55 27 PM

Thanks for the update. Let's keep the current implementation. We can re-visit it later on if needed.

PR #799, opened 2022-08-02
Select all series in the left tree of Histogram output by default

### LINKED ISSUE #786: Select all series in the left tree of Histogram output by default
All series in the left tree of the Histogram output should be selected by default. This will improve the user experience for both the Histogram output and the trace overview.

hriday-panchasara:
fixes #786

### DIFF
--- a/packages/react-components/src/components/abstract-xy-output-component.tsx
+++ b/packages/react-components/src/components/abstract-xy-output-component.tsx
@@ -85,6 +85,7 @@ export abstract class AbstractXYOutputComponent<P extends AbstractOutputProps, S
     private preventDefaultHandler: ((event: WheelEvent) => void) | undefined;
     protected posPixelSelect = 0;
     protected positionYMove = 0;
+    protected isHistogram: boolean = this.props.outputDescriptor.id.includes('histogram');
 
     // Event flags
     protected isMouseLeave = false;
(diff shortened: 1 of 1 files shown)


packages/react-components/src/components/abstract-xy-output-component.tsx
```diff
@@ -85,6 +85,7 @@ export abstract class AbstractXYOutputComponent<P extends AbstractOutputProps, S
     private preventDefaultHandler: ((event: WheelEvent) => void) | undefined;
     protected posPixelSelect = 0;
     protected positionYMove = 0;
+    protected isHistogram: boolean = this.props.outputDescriptor.id.includes('histogram');
```
bhufmann, line 88: This is specific when running the trace viewer front-end with the Trace Compass server. Other servers might not have a histogram, or have a histogram where all series should not be selected. The solution of selecting default series needs to be independent to the trace server implementation. That means, the information which series to select by default needs to be provided by the back-end over the TSP.  Some protocol design is required for that.

bhufmann (changes requested):

packages/react-components/src/components/abstract-xy-output-component.tsx
hriday-panchasara: Thanks Bernd, I'll look into doing so over the TSP and close this PR. Before that I wanted to ask if we can still tell the front-end to select all series only if an Histogram overview view is opened, or whether it'll be better to let the TSP do so in this case as well.

PR #803, opened 2022-08-04
Refactor available views component

hoangphamEclipse:
Currently, the trace-overview-selection-dialog-component and the trace-explorer-views-widget have some common code. This
commit refactors the common code by creating the reusable available-view-component. The prior two components will then call the later component and thus use the same code.


bhufmann (approved):
Looks good to me. Thanks!

MatthewKhouzam (approved):
Keep an open eye, let's get more patches like this!

PR #804, opened 2022-08-10
Add Events Table Shortcuts to the Keyboard Shortcuts Menu

### LINKED ISSUE #742: Support keybord shortcuts in events table
Add keyboard shortcuts for navigation:

- Arrow up/down: Select next or previous event
- Page Up/Down: Select event one page (window size) up or down
- End: Select last event in trace
- Home: Select first event in trace

hriday-panchasara:
Helps fix #742

- The requested shortcuts described in #742 are implemented by default in all ag-grid tables: https://www.ag-grid.com/react-data-grid/keyboard-navigation/
- I added these shortcuts to the Keyboard Shortcuts menu to make them discoverable. I also added the following keyboard/mouse interactions:
        - Select Multiple rows: 'Shift + click' or 'Shift + arrow up/down'
        - Deselect rows: 'ctrl or meta key'
- In table-output-component I also noticed that 'space' is registered as a key interaction to replace multi-select with a single row, but was not sure how to describe it in the shortcuts menu as a user can single-click or use ctrl to remove multi-select
- Here's how the shortcuts menu looks now:
Screen Shot 2022-08-10 at 12 15 02 PM

### DIFF
--- a/theia-extensions/viewer-prototype/src/browser/trace-explorer/trace-explorer-sub-widgets/trace-explorer-keyboard-shortcuts/events-table-shortcuts-table.tsx
+++ b/theia-extensions/viewer-prototype/src/browser/trace-explorer/trace-explorer-sub-widgets/trace-explorer-keyboard-shortcuts/events-table-shortcuts-table.tsx
@@ -1,0 +1,84 @@
+import * as React from 'react';
+
+export class EventsTableShortcutsTable extends React.Component {
+
+    render(): React.ReactNode {
+        return <React.Fragment>
+            <span className='shortcuts-table-header'>TABLE NAVIGATION</span>
+            <div className='shortcuts-table-row'>
+                <div className='shortcuts-table-column'>
+                    <table>
+                        <tbody>
+                            <tr>
+                                <td>Cell Navigation</td>
+                                <td className='monaco-keybinding shortcuts-table-keybinding'>
+                                    <span className='monaco-keybinding-key'><i className='fa fa-arrow-up' /></span>
+                                    <span className='monaco-keybinding-key'><i className='fa fa-arrow-down' /></span>
+                                    <span className='monaco-keybinding-key'><i className='fa fa-arrow-left' /></span>
+                                    <span className='monaco-keybinding-key'><i className='fa fa-arrow-right' /></span>
+                                </td>
+                            </tr>
+                            <tr>
+                                <td>Page Up</td>
+                                <td className='monaco-keybinding shortcuts-table-keybinding'>
+                                    <span className='monaco-keybinding-key'>Page Up</span>
+                                </td>
+                            </tr>
+                            <tr>
+                                <td>Page Down</td>
+                                <td className='monaco-keybinding shortcuts-table-keybinding'>
+                                    <span className='monaco-keybinding-key'>Page Down</span>
+                                </td>
+                            </tr>
+                            <tr>
+                                <td>Multi-select</td>
+                                <td className='monaco-keybinding shortcuts-table-keybinding'>
+                                    <span className='monaco-keybinding-key'>Shift</span>
+                                    <span className='monaco-keybinding-key'>Click</span>
+                                    <span className='monaco-keybinding-seperator'>or</span>
+                                    <span className='monaco-keybinding-key'>Shift</span>
+                                    <span className='monaco-keybinding-key'><i className='fa fa-arrow-up' /></span>
+                                    <span className='monaco-keybinding-key'><i className='fa fa-arrow-down' /></span>
+                                </td>
+                            </tr>
+                        </tbody>
+                    </table>
+                </div>
+                <div className='shortcuts-table-column'>
+                    <table>
+                        <tbody>
+                            <tr>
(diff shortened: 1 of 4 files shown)


hoangphamEclipse:
Thank you for the PR. I tried the code out and it works. However, the navigation itself for the table is not very smooth.

[1]  A cell needs to be explicitly selected before the navigation with arrows key can work. When an event table is opened the first time, an event is selected, but the navigation does not work because no cells is selected. I don't know if this is the selected behavior. 

[2] When you deselect with Ctrl, you need to move to a new row before you can start selecting multiple row with Shift+Ctrl again.

> In table-output-component I also noticed that 'space' is registered as a key interaction to replace multi-select with a single row, but was not sure how to describe it in the shortcuts menu as a user can single-click or use ctrl to remove multi-select

Looks like the `Spacebar` is used to select a single row. So in addition to what you have mentioned, I can also select/highlight an entire row as well.

hriday-panchasara:
> Thank you for the PR. I tried the code out and it works. However, the navigation itself for the table is not very smooth.
> 
> [1] A cell needs to be explicitly selected before the navigation with arrows key can work. When an event table is opened the first time, an event is selected, but the navigation does not work because no cells is selected. I don't know if this is the expected behavior.
> 
> [2] When you deselect with `Ctrl`, you need to move to a new row before you can start selecting multiple row with `Shift+Ctrl` again.
> 
> > In table-output-component I also noticed that 'space' is registered as a key interaction to replace multi-select with a single row, but was not sure how to describe it in the shortcuts menu as a user can single-click or use ctrl to remove multi-select
> 
> Looks like the `Spacebar` is used to select a single row. So in addition to what you have mentioned, I can also select/highlight an entire row as well.

I believe [1] is expected behavior, in order to interact with a chart there needs to be tab focus which is indicated by a blue outline. This can be either achieved by clicking 'Tab' until you tab into the desired cell, or by clicking on it. For example on the xy-charts you need to click on the chart first or tab into it before attempting keyboard operations (zoom and pan through buttons).

For [2], you can click spacebar to select the current row rather than moving to another row for multi-select. In table-output-component, I noticed that only 'Shift + arrow up/down', 'Spacebar', and 'arrow up/down' result in row(s) being selected - any other key press would lead to gridapi.deselectAll(). This makes me believe that [2] is intentional behavior as well

hriday-panchasara:
Added the 'Spacebar' keybinding to the shortcuts menu:
Screen Shot 2022-08-12 at 2 58 07 PM

hoangphamEclipse:
> > Thank you for the PR. I tried the code out and it works. However, the navigation itself for the table is not very smooth.
> > [1] A cell needs to be explicitly selected before the navigation with arrows key can work. When an event table is opened the first time, an event is selected, but the navigation does not work because no cells is selected. I don't know if this is the expected behavior.
> > [2] When you deselect with `Ctrl`, you need to move to a new row before you can start selecting multiple row with `Shift+Ctrl` again.
> > > In table-output-component I also noticed that 'space' is registered as a key interaction to replace multi-select with a single row, but was not sure how to describe it in the shortcuts menu as a user can single-click or use ctrl to remove multi-select
> > 
> > 
> > Looks like the `Spacebar` is used to select a single row. So in addition to what you have mentioned, I can also select/highlight an entire row as well.
> 
> I believe [1] is expected behavior, in order to interact with a chart there needs to be tab focus which is indicated by a blue outline. This can be either achieved by clicking 'Tab' until you tab into the desired cell, or by clicking on it. For example on the xy-charts you need to click on the chart first or tab into it before attempting keyboard operations (zoom and pan through buttons).
> 
> For [2], you can click spacebar to select the current row rather than moving to another row for multi-select. In table-output-component, I noticed that only 'Shift + arrow up/down', 'Spacebar', and 'arrow up/down' result in row(s) being selected - any other key press would lead to gridapi.deselectAll(). This makes me believe that [2] is intentional behavior as well

Thank you @hriday-panchasara, this make sense. Also thank you for adding the `Spacebar` to the shortcut menu.

hoangphamEclipse (dismissed):

Rodrigoplp-work (dismissed):

theia-extensions/viewer-prototype/src/browser/trace-explorer/trace-explorer-sub-widgets/trace-explorer-keyboard-shortcuts/events-table-shortcuts-table.tsx
```diff
@@ -0,0 +1,84 @@
+import * as React from 'react';
+
+export class EventsTableShortcutsTable extends React.Component {
+
+    render(): React.ReactNode {
+        return <React.Fragment>
+            <span className='shortcuts-table-header'>EVENTS TABLE NAVIGATION</span>
```
PatrickTasse, line 7: Should this be more generically "TABLE NAVIGATION" ?

It applies to output type "TABLE" which covers the Events Table but also now some instances of Latency Table.

PatrickTasse (changes requested):

marco-miller (approved):
LGTM too; thanks!

PR #805, opened 2022-08-12
Selection of time range through marker axis

### LINKED ISSUE #768: Support selection of time range using marker axis (timegraph-output-components)
The time graph output components supports showing a marker axis that can be filled with trace annotations provided by specific data providers. For example, lost event is one generic example. Other data providers can add their own domain specific markers as well.

It would be good that when selection a marker, that the time range of the marker is selected in all views.

hriday-panchasara:
Clicking a marker on the marker axis would result in its time range to be selected in all views

fixes #768


bhufmann (approved):
Looks good to me. Thanks!

Rodrigoplp-work (approved):

PR #690, opened 2022-02-23
Upgrade to Theia 1.21.0

hoangphamEclipse:
This commit upgrades the Theia version of the trace extension to 1.21.0.

Signed-off-by: Hoang Thuan Pham <hoang.pham@calian.ca>

PR #813, opened 2022-08-25
Upgrade Theia dependency to 1.29.0

Rodrigoplp-work:
Currently theia-trace-extenison is using version 1.21.0 of `@theia/cli`.

This pull request upgrades this dependency to the lastest one available,
1.29.0.

The version of NodeJS used by the extension (12.14.1) was no longer
accepted by Theia, so it is being upgraded to 16.16.0.

Signed-off-by: Rodrigo Pinto (<rodrigo.pinto@calian.ca>)


.gitpod.dockerfile
```diff
@@ -16,7 +16,7 @@ RUN sudo apt-get update \
     # Clear package info (to free space)
     && sudo rm -rf /var/lib/apt/lists/*
 
-ENV NODE_VERSION="12.14.1"
+ENV NODE_VERSION="16.16.0"
```
marco-miller, line 19: 16.17.0 would be the current latest LTS version for v16.

marco-miller:
Addendum: the above CI failures are likely caused by `yarn.lock`, which needs to be updated locally first (then pushed herein).

package.json
```diff
@@ -39,7 +39,7 @@
   },
   "homepage": "https://github.com/theia-ide/theia-trace-extension",
   "devDependencies": {
-    "@theia/cli": "1.21.0",
+    "@theia/cli": "1.29.0",
     "jsonc-parser": "^3.0.0",
     "lerna": "^4.0.0",
     "typescript": "4.5.4"
```
marco-miller, line 45: Theia's own latest TS version says "~4.5.5"; amend here too?

marco-miller: One more thing: #690 and the earlier #461 had more updates upon bumping Theia's versions that way. Does this PR also require similar dependency upgrades, say, based on Theia's corresponding bumps (to align)?

marco-miller:
@hriday-panchasara, I cannot add you as Reviewer in here, but feel free to share your perspective; thanks.

.gitpod.dockerfile
Rodrigoplp-work: Agreed.

package.json
Rodrigoplp-work: Yes, I think it's a good idea to keep the TS dependency on par with Theia. Upgraded.
Thank you for bringing up those other PRs. I upgraded all the other related dependencies.

packages/react-components/jest-shim.ts
```diff
@@ -0,0 +1,6 @@
+import { TextEncoder } from 'util';
```
marco-miller, line 1: `"util"`, according to the existing convention in this repo, AFAIK.

packages/react-components/jest-shim.ts
```diff
@@ -0,0 +1,6 @@
...
+/*
+ * Add TextEncoder from NodeJS util as it is
+ * no longer avaiable for Jest tests after React 18.
+ */
+global.TextEncoder = TextEncoder;
```
marco-miller, line 6: EOF newline missing.

packages/react-components/jest-shim.ts
```diff
@@ -0,0 +1,6 @@
+import { TextEncoder } from 'util';
+/*
+ * Add TextEncoder from NodeJS util as it is
+ * no longer avaiable for Jest tests after React 18.
```
marco-miller, line 4: `available` (typo).

packages/react-components/src/components/__tests__/table-renderer-components.test.tsx
```diff
@@ -18,7 +18,7 @@ afterEach(() => {
 
 describe('Table output component tests', () => {
     
-    test('Renders AG-Grid table with provided props & state', async () => {
+    test.skip('Renders AG-Grid table with provided props & state', async () => {
```
marco-miller, line 21: Add a comment about why skipping, similar to the other ones added in?

packages/react-components/src/components/utils/filter-tree/__tests__/table-row.test.tsx
```diff
@@ -233,13 +220,9 @@ describe('Multiple labels', () => {
...
             onToggleCheck={mockOnChecked} isClosable={false} onClose={mockOnClose}        />)
             .toJSON();
         expect(nodeWithChildren).toMatchSnapshot();
     })
 
-});
-
-
-
+});
```
marco-miller, line 228: EOF newline missing.

marco-miller (changes requested):
Beside these latest yet minor comments, some views for me don't render as expected locally; e.g., Overview -as well as some others. Do they, for you?

packages/react-components/jest-shim.ts
Rodrigoplp-work: Thank you. Fixed.

packages/react-components/jest-shim.ts
Rodrigoplp-work: Thank you. Fixed.

packages/react-components/src/components/__tests__/table-renderer-components.test.tsx
Rodrigoplp-work: Indeed. Fixed.

packages/react-components/src/components/utils/filter-tree/__tests__/table-row.test.tsx
Rodrigoplp-work: Thank you. Fixed.

Rodrigoplp-work:
> > Beside these latest yet minor comments, some views for me don't render as expected locally; e.g., Overview -as well as some others. Do they, for you?
> 
> They do not. I believe they are failing due to the changes I did in the "Typescript adjustments for React 18" commit, which were made to fix new compiler's errors coming from `@types/react 18`. I'll be looking into different ways to resolve those, perhaps making use of suggestions from the team.

It took a tenuous investigation, but we found the culprit! @hriday-panchasara (brilliantly!) spotted a change in the way state was being updated in `trace-explorer-opened-traces-widget.tsx`. It's caused by React's new automatic batching feature, and we adapted accordingly.

Thank you, Hriday!

packages/react-components/src/trace-explorer/trace-explorer-opened-traces-widget.tsx
```diff
@@ -272,7 +273,10 @@ export class ReactOpenTracesWidget extends React.Component<ReactOpenTracesWidget
         });
         const selectedIndex = remoteExperiments.findIndex(experiment => this._selectedExperiment &&
             experiment.UUID === this._selectedExperiment.UUID);
-        this.setState({ openedExperiments: remoteExperiments, selectedExperimentIndex: selectedIndex });
+        // flushSync: force imediate state update instead of waiting for React 18's automatic batching
```
marco-miller, line 276: immediate

marco-miller:
> > > Beside these latest yet minor comments, some views for me don't render as expected locally; e.g., Overview -as well as some others. Do they, for you?
> > 
> > They do not. I believe they are failing due to the changes I did in the "Typescript adjustments for React 18" commit, which were made to fix new compiler's errors coming from `@types/react 18`. I'll be looking into different ways to resolve those, perhaps making use of suggestions from the team.
> 
> It took a tenuous investigation, but we found the culprit! @hriday-panchasara (brilliantly!) spotted a change in the way state was being updated in `trace-explorer-opened-traces-widget.tsx`. It's caused by React's new automatic batching feature, and we adapted accordingly.
> 
> Thank you, Hriday!

Cool. I tested again now and these views for me are empty with the kernel test trace:
1. Histogram
2. IRQ Analysis - Latency vs Time
3. System Call Latency (Latency vs Time)

Rodrigoplp-work:
> > > > Beside these latest yet minor comments, some views for me don't render as expected locally; e.g., Overview -as well as some others. Do they, for you?
> > > 
> > > 
> > > They do not. I believe they are failing due to the changes I did in the "Typescript adjustments for React 18" commit, which were made to fix new compiler's errors coming from `@types/react 18`. I'll be looking into different ways to resolve those, perhaps making use of suggestions from the team.
> > 
> > 
> > It took a tenuous investigation, but we found the culprit! @hriday-panchasara (brilliantly!) spotted a change in the way state was being updated in `trace-explorer-opened-traces-widget.tsx`. It's caused by React's new automatic batching feature, and we adapted accordingly.
> > Thank you, Hriday!
> 
> Cool. I tested again now and these views for me are empty with the kernel test trace:
> 
> 1. Histogram
> 2. IRQ Analysis - Latency vs Time
> 3. System Call Latency (Latency vs Time)

Great catch, Marco! They were all caused by the same issue. Fixed.

packages/react-components/src/trace-explorer/trace-explorer-opened-traces-widget.tsx
Rodrigoplp-work: Thank you! Fixed.

marco-miller (approved):
LGTM now, thanks and good work!

packages/react-components/src/components/table-output-component.tsx
```diff
@@ -408,8 +416,8 @@ export class TableOutputComponent extends AbstractOutputComponent<TableOutputPro
         });
 
         if (this.columnApi) {
-            const columns = this.columnApi.getAllColumns();
-            const timestampHeader = columns.find(column => column.getColDef().headerName === 'Timestamp ns');
+            const columns = this.columnApi.getColumns();
```
bhufmann, line 419: This seems not to work. columns is an empty array. Because of that the timestampHeader can't found and hence this.timestampCol is never set. This cause that time synchronization between view are not working because no signal is sent due to undefined this.timetampCol.

Set a breakpoint in method onEventClick() and when stepping through the code that no signal is sent.

bhufmann (changes requested):
Thanks for the upgrade PR!

I found a issues:
- View menu is not rendered (e.g. events table, data trees)
- Events table sync (see comment in table-output-component)

Due the menu bug, I was not able to test the commands of these menus

packages/react-components/src/components/table-output-component.tsx
Rodrigoplp-work: Good catch! `getColumns` is empty because state is empty. Resolved by applying `flushSync` to a `setState` function earlier in the class.

Rodrigoplp-work:
> Thanks for the upgrade PR!
> 
> I found a issues:
> 
> * View menu is not rendered (e.g. events table, data trees)
> * Events table sync (see comment in table-output-component)
> 
> Due the menu bug, I was not able to test the commands of these menus

Indeed.
- View menu: the event listeners were being triggered immediately upon creation, maybe due to the new way React updates the state. I had to add a timeout to avoid bubbling;
- `getColumns`: fixed with ` flushSync`, as explained above.

bhufmann:
> 

It's happening consistently. It's happening when the table shows events closer to the end of the trace. I set a breakpoint in onEventClick(), and when it fails it's not called. I'm not able to pin-point why the event handler is not called in this case. Any ideas?

Rodrigoplp-work:
> > 
> 
> It's happening consistently. It's happening when the table shows events closer to the end of the trace. I set a breakpoint in onEventClick(), and when it fails it's not called. I'm not able to pin-point why the event handler is not called in this case. Any ideas?

Apparently the new versions of AG Grid (I'm not sure when it started) have a new way to deal with the number of rows. When it's too big, scrolling, hovering and row selection stop working.

https://www.ag-grid.com/javascript-data-grid/massive-row-count/#:~:text=This%20means%20in%20theory%20only%20320%2C000%20rows%20can%20get%20displayed.

They have an example on that page, a table with 1 million rows. If you scroll to the bottom of the table (~ after row 500,000) you will see the same problems we have.

Back in theia-trace-extension, if you use `rowHeight={20}` as an option of `AgGridReact` in `renderMainArea`, you will see that all rows work well. But if we had more rows, the problem would eventually happen again.

Even pagination doesn't solve the problem. We already apply it by retrieving a few hundred rows per time. The problem is the index of the rows, the index used for scrolling the page. Even with version 28.2.0 of ag-grid, released 3 hours ago, the problem is still there.

Rodrigoplp-work:
Proposed a solution that uses Ag-grid's pagination.

Rodrigoplp-work:
> The time sync is working now as well as the menus and its features. Thanks! However I found an issue that the table sometimes looses the ability to select an event by clicking. It happens after clicking outside the current window in the histogram (for example) and the table has to load new data. It doesn't happen all the time, but often. Could you please look into it?
> 
> failing-click-table [ failing-click-table ](https://user-images.githubusercontent.com/413424/193304981-dbd60646-d496-46a7-a317-be2778e27e08.gif) [ ](https://user-images.githubusercontent.com/413424/193304981-dbd60646-d496-46a7-a317-be2778e27e08.gif)

It happens indeed, as described on the last comments. Added change of pages when clicking on the next / previous search button.

bhufmann (approved):
Looks good to me now! Thanks a lot for doing this challenging upgrade.

marco-miller (approved):
LGTM too again now, thanks and kudos @Rodrigoplp-work for this courageous effort. I tested these commits rebased on latest master locally. -Letting @bhufmann rebase/merge this PR at will if ready to do so at this stage.

PR #821, opened 2022-09-01
Debounce xy chart data update

hriday-panchasara:
Replace throttle with debounce when handling xy chart update requests


bhufmann:
Here is what it will look like:

new-hover-trace-explorer

bhufmann (approved):
Looks good to me. Thanks!

marco-miller (approved):
LGTM too.

PR #823, opened 2022-09-02
Frontend fixes

hriday-panchasara:
Add the following UI fixes:
- Fix issue with datatree csv export in electron
- Fix padding in options-menu list items
- Fix onRowClick not defined error with filter-tree component
- Fix padding in server status indicator
- Fix 'Start Trace Server' issue


bhufmann (changes requested):
This PR combines different fixes in one PR. Each fix should be a separated git commit. Since there are no issue tracker problem, each fix should describe what the issue was. You could add issues trackers and add `fixes #...` in the commit in the PR to be more explicit.

marco-miller:
Closing since all moved to the aforementioned split PRs.

PR #833, opened 2022-09-12
Support window dragging with mouse-middle drag in xy charts

### LINKED ISSUE #732: Support window dragging with mouse-middle drag in xy charts
In comparison to Time Graph charts, the xy-chart hasn't implemented the window drag with `middle mouse button` drag. This shall be added to the XY-charts. The cursor shall be update to the hand icon during this operation.

#702 added support for this operation using CTRL+mouse left button drag.

hriday-panchasara:
Fixes #732

### DIFF
--- a/packages/react-components/src/components/xy-output-component.tsx
+++ b/packages/react-components/src/components/xy-output-component.tsx
@@ -209,7 +209,7 @@ export class XYOutputComponent extends AbstractXYOutputComponent<AbstractOutputP
                     start: this.props.unitController.selectionRange.start,
                     end: startTime
                 };
-            } else if (event.ctrlKey && !event.shiftKey) {
+            } else if ((event.ctrlKey && !event.shiftKey) || (!(event.shiftKey && event.ctrlKey) && this.clickedMouseButton === MouseButton.MID)) {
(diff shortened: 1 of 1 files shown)


packages/react-components/src/components/xy-output-component.tsx
```diff
@@ -241,6 +241,7 @@ export class XYOutputComponent extends AbstractXYOutputComponent<AbstractOutputP
     }
 
     private onWheel(wheel: React.WheelEvent): void {
+        console.log('onWheel called');
```
marco-miller, line 244: Should this be kept?

marco-miller (changes requested):
Works well for me locally. Note, this PR fixes both `XY` and `TimeGraph` cases it seems. Without this patch, both cases are lacking mouse-middle button drag support, it seems. At least from my local point of view. And, I do see the hand pointer upon clicking that way.

packages/react-components/src/components/xy-output-component.tsx
hriday-panchasara: Oops, forgot to take it out

marco-miller (approved):
LGTM, with that note I posted above / earlier. Thanks!

bhufmann (approved):
LGTM. Thanks!

PR #843, opened 2022-09-29
Add ADR for programmatic columns UI/UX

hoangphamEclipse:
This commit adds an ADR that contains the proposal for a UI/UX desgin for programmatic columns in the Theia Trace Extension.


jonahgraham:
This looks really interesting and a welcome addition. I have some questions, but they aren't mean't to holdup the first steps as many of them may simply be considered longer term considerations.

- Can you reference previous/next state in the calculation? To expand on the example in the proposal, if I have a column of timestamps, can I have a calculated column showing duration that subtracts the previous row's timestamp from this row's one?
- Does the calculation happen on the client side or the server side?
- Can you visualize (graph) the new columns?

_As I found the diff version a little hard to read (And others may too) here is a [link back to the originating branch](https://github.com/hoangphamEclipse/theia-trace-extension/blob/programmatic-columns-adr/doc/adr/0007-programmatic-columns.md) where the markdown is well rendered with images and all. If there is a GitHub trick to see that from the PR directly let me know._

hoangphamEclipse:
Hi @jonahgraham, thank you for your feedback. I will get @MatthewKhouzam or @bhufmann to join in this conversation because some questions are better answered by them.

> * Can you reference previous/next state in the calculation? To expand on the example in the proposal, if I have a column of timestamps, can I have a calculated column showing duration that subtracts the previous row's timestamp from this row's one?

> * Can you visualize (graph) the new columns?

These were not original requirements for this ADR, but it is interesting. I will let Bernd or Matthew give their opinion on this.

> * Does the calculation happen on the client side or the server side?

The calculation will be on the server side.

> As I found the diff version a little hard to read (And others may too) here is a [link back to the originating branch](https://github.com/hoangphamEclipse/theia-trace-extension/blob/programmatic-columns-adr/doc/adr/0007-programmatic-columns.md) where the markdown is well rendered with images and all. If there is a GitHub trick to see that from the PR directly let me know.

Thank you for noticing this, I will put the link to the well rendered ADR in the PR description so that it will be easier for others to view it.

MatthewKhouzam:
Hi @jonahgraham, thank you for your feedback. I will get @MatthewKhouzam or @bhufmann to join in this conversation because some questions are better answered by them.

>        Can you reference previous/next state in the calculation? To expand on the example in the proposal, if I have a column of timestamps, can I have a calculated column showing duration that subtracts the previous row's timestamp from this row's one?

We had this in mind with the caveat that filtering messes this up. So we put it off for "later"
I really like (time - time') as a column

>        Can you visualize (graph) the new columns?

That is the next logical and soon to be published step! :)

bhufmann:
What is the plan of this one?

PR #852, opened 2022-10-13
Add ADR about custom actions through queryable properties

bhufmann:
This ADR describes how to add domain specific queries for data provider without the front-end know the back-end domain. It enables features like `Go to max execution time`, `Go to crash time`, `Run critical path execution` and more. See ADR for details. 

### DIFF
--- a/doc/adr/0007-queryable-properties.md
+++ b/doc/adr/0007-queryable-properties.md
@@ -1,6 +1,6 @@
 # 7. Queryable properties
 
-Date: 2022-10-13
+Date: 2023-01-30
 
 ## Status
 
@@ -10,17 +10,18 @@ New
 
 User of the trace-viewer application are often interested in additional information from the data provider. For example, the Trace Compass latency statistics views has columns about minimum and maximum duration and user would like to navigate to the time range where this maximum occurred. The front-end that uses the Trace Server Protocol doesn't know about what data it displaying. It only knows about different graph types. For the latency views it's a table tree with columns, but it doesn't know that there is a minimum column or maximum column, nor it knows where in the trace it occurred.
 
-The Trace Server Protocol (TSP) needs to be augmented to provide such type of data, so that the front-end can provide controls for the user to navigate.
+The Trace Server Protocol (TSP) needs to be augmented to provide such type of data, so that the front-end can provide controls for the user to apply actions in the front-end like select and/or navigate to a given time range, open source code or show a new view using a new custom data provider. These properties are meant to be optional and on-demand, for specific select elements in the views. They should be quick queries and the returned data structures should have limited purpose. They are not meant to query all elements in the view. If that is required, new data providers should be created.
 
-**Use cases:**
+This feature is complementary to the tooltip endpoint for data providers which provides some additional, static details on an UI element. 
 
-- GoTo min/max per entry or all
-- Get longest executions per entry or all
-- Get the N longest execution per thread
-- Get Crash details (time and other details like CPU)
-- GoTo most utilized frame
-- Get the N most utilised spans in order
-- Run critical path analysis (returns a data provider descriptor)
+### Use cases
+
+- Get min/max time range per entry
+- Get the N longest executions per thread
+- Get crash time (other details like CPU, process, thread)
+- Get most utilized frame (time range)
+- Get source location and open source code file
+- Run critical path analysis and create a data provider descriptor
 
 The challange is to handshake these properties or capabilites between front-end (clients) and server over an domain-agnostic protocol.
 
@@ -30,35 +31,76 @@ Augment the `DataProviderDescriptor` data structure with information about query
 
 ```java
     /*
-     * QuerablePropertyDescriptor {
-     *  name:        string
-     *  description: string
-     *  id:          string
-     *  actionType:  apply, display
-     *  inputType:   entry, custom (e.g. thread name, or filter string?)
-     *  queryParams: {
-     *                 requested_items:          optional, array of entry ids if inputType is entry, if omitted all
-     *                 requested_time_range:     optional, of not passed full trace
-     *                 requested_custom_strings: optional, provide custom strings (filter string?)
-     *               }
-     *  returnType:  one of:
-     *                   - array of { id: entryId, if inputType=entry
-     *                              name: string used if inputType=custom
-     *                              one of:
-     *                                time_range_list: [{start1, end1}, ..., {startN, endN}]
-     *                                timestamp_list: [time1, time2..., timeN]
-     *                                // Note if actionType=apply only use the first one, if list.size > 0
-     *                            }
-     *                   - new data provider descriptor
-     */
+        PropertyDescriptor 
+        {
+            name:        string
+            description: string
+            id:          string
+            queryParams: {
+                            requested_items:          optional, single entry ID of interest
+                            requested_element:        optional, single element of interest (state or arrow or link, state, duration)
+                            requested_time_range:     optional, time range the query is for, or full trace if ommitted
+                            requested_table_index:    optional, index of event
+                        }
+            propertyModelType: 
+                        enum {
+                            time_ranges,
+                            dp_descriptors,
+                            callsites
+                        }
+        }                
 
-    /*
-     * Notes: for actionType displayed, the data should be shown and the user should
-     * have a menu either select (highlight) or (zoom into) or (move to).
+        Where:
+            queryParams can be any query paramerter that is allowed for a given data provider type.
+
+            Returned PropertyModel determined by propertyModelType:
+
+                time_ranges:
+                        {   
+                            array of {
+                                time_range: {start, end}, where end is optional for single a timestamp
+                                metadata: optional, map of key-value pairs to identify cpu, thread etc.
+                            }
+                        }
+                dp_descriptors:
+                        {   
+                            array of {
+                                dp_descriptor: OutputDescritpor
+                            }
+                        }
+                callsites:
+                        {   
+                            array of {
+                                callsite: {file, line}
+                        }
      */
 ```
 
-### Example Sequence Diagram (for GoTo Max)
+### Suggested default behaviours
+
+`Single time ranges (or time)`
+
+- "Select" or "Select and Reveal" time
+
+`Multiple time ranges`
+
+- Display list of time ranges (e.g. dialog, overlay list)
+- User can click on time range
+  - "Select" or "Select and Reveal"
+
+`List of data providers`
+
+- Refresh `Available Views` view with new descriptors
+- User can select new descripter to open view
+
+`List of callsites`
+
+- Display list of callsites (e.g. dialog, overlay list)
+- User can click on a callsite and source code file will be opened using IDE's capabilities
+
+### Example Sequence Diagram
+
+Prerequiste `Open Trace and fetched property descriptors` for all diagrams:
 
 ```mermaid
 sequenceDiagram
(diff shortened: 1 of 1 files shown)


bhufmann:
Thanks @MatthewKhouzam for the typo fixes. I'll update my branch with the fixes.

doc/adr/0007-queryable-properties.md
```diff
@@ -0,0 +1,96 @@
...
+
+Augment the `DataProviderDescriptor` data structure with information about queryable properties.
+
+```java
+    /*
+     * QuerablePropertyDescriptor {
+     *  name:        string
+     *  description: string
+     *  id:          string
+     *  actionType:  apply, display
```
frallax, line 37: Can we add some more details on the meaning of "actionType"? By reading this ADR it is not very clear to me when the server is supposed to return one type or the other. Maybe also adding some info to line 73 below could help to clarify.

Same for the input type.

hriday-panchasara:
Demo: https://github.com/eclipse-cdt-cloud/theia-trace-extension/pull/882

doc/adr/0007-queryable-properties.md
```diff
@@ -0,0 +1,286 @@
...
+    Note over client,server: Prerequisite Open Trace and fetched property descriptors
+    User->>client: open Latency Table
+    client->>server: POST /experiment/{expUuid}/outputs/{outputId}/data/tree
+    server->>client: 200: TreeModel
+    client->>client: populate view
+    User->>client: open view menu
+    loop while descriptor in propertyDescriptors
+        client->>client: create menu entry (using descriptor.name, descriptor.description, descriptor.propertyId)
+    end
+    User->>client: select "Go To Max" menu item
```
frallax, line 171: I think here "Go to max" is a copy/paste error

doc/adr/0007-queryable-properties.md
```diff
@@ -0,0 +1,286 @@
...
+
+Notes: * This doesn't exists yet and needs to be implemented. Tooltip endpoint can be added for that.
+
+## Decision
+
+Implement this feature for the `Go to min/max` for Latency views first which represents an example time navigation use case. Make it the TSP definition extendible so that the other use cases can be added gradually.
+
+## Consequences
+
+If this proposal is implemented, the tool becomes more interactive and it will provide a better UX. Users can retrieve more details from the backend in order to navigate inside the trace and find relevant, interesting time ranges, to look-up source code or create custom data providers.
```
frallax, line 284: Can you clarify what you mean by "the tool"? You mean the frontend (i.e. theia-trace-extension) only?

doc/adr/0007-queryable-properties.md
```diff
@@ -0,0 +1,286 @@
...
+Augment the `DataProviderDescriptor` data structure with information about queryable properties.
+
+```java
+    /*
+        PropertyDescriptor 
+        {
+            name:        string
+            description: string
+            id:          string
+            queryParams: {
```
frallax, line 39: How will the "queryParams" and "metadatakeys" be regulated in the TSP? Will each data provider need to add a description of the available queryParams / properties? In other words, can you provide an example of how the TSP will change too?

doc/adr/0007-queryable-properties.md
```diff
@@ -0,0 +1,286 @@
+# 7. Queryable properties
```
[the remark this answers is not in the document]
bhufmann: > General comment: I think this is a very clever proposal, but there are some aspects of it that I still do not fully understand. The main "fear" I feel with this proposal is that the trace-server will have to embed too much logic that is specific to only one specific client, and that results useless/redundant/confusing to many other clients. But probably it is just because I do not fully understand the logic, so pick this statement with a grain of salt.
> 
I take the `clever proposal` :-)

I glad that you raise your concerns. We use ADR to be able to discuss the proposals to come up with reasonable design. 

I understand that it feels complicated. I was wondering myself if it's over-engineered and if it could be made simpler. 

> In general I feel that the client should have a minimal intelligence too, to avoid overloading the server. For example, if the client asks data from the end point that represents a data provider that generates the latency statistics tables, I would expect that the client knows what it is actually requesting (i.e. a table that contains "_the statistics of the latencies. These show the minimum, maximum, average, standard deviation, count and Total of the latencies when applicable_" as documented [here](https://archive.eclipse.org/tracecompass/doc/stable/org.eclipse.tracecompass.doc.user/Latency-Analyses.html)). I mean, as we expect that a user reads the documentation before clicking randomly on trace compass views to get what he/she wants...

> 
> Any thoughts? Or am I diverging too much from the fact that the front-end should be a pure "data visualization tool without domain specific knowledge" ?

Not sure how we could convey some domain specific content to the client that. We would have to define some "Concepts" which instructs the clients with additional behaviour:
- If a column is tagged (or has metadata) 'maximum', then client can fetch that time range.  
- If an element has a string that is a timestamp or time range, user should be able to select and navigate to
- If an element has a string that corresponds to a callsite, the user should be able to open that callsite

The latter 2 could use the tooltip endpoint, where we have key-value pairs to be displayed and where the key is the known metadata string.

> 
> If I'd like to give a counter-proposal, should I create a separate ADR, or should we try to keep the discussion on this one?

Maybe a dedicated proposal would be easier to follow. You can use markdown formatting, provide sequence diagrams etc. to illustrate your proposal. We can have a reference in this ADR.

doc/adr/0007-queryable-properties.md
bhufmann: yeah, I added the metadata part in the last proposal for callsites, because I realized that certain information is not available when constructing the `PropertyDescriptor` and it is dependent on the instance of trace event or state etc. I want to avoid to show or enable a menu if certain information can't be retrieved. For example for callsites, only certain time graph states or trace events can provide that info.

I used metadata because we already use that. The TreeModel for time graphs, xy or data tree can add a metadata map  to each row. For example, the resource view has a metadata map entry for key=cpu and a value. The key is common across views and that's how we are able to select the corresponding CPU row when another view sends a selection signal with such known key.

I agree, it's a bit clunky and I wouldn't have put it in the TSP initially, and only when implementing the solution I would revisit that.  

Hopefully with you counter proposal, it will get also easier to transport callsite information and provide source code look-up for the users in various views. (BTW, in the in the properties view we look for key source to find the source code strin. This was just a prototype and needs to be revisited)

doc/adr/0007-queryable-properties.md
```diff
@@ -0,0 +1,286 @@
...
+    User->>client: open Latency Table
+    client->>server: POST /experiment/{expUuid}/outputs/{outputId}/data/tree
+    server->>client: 200: TreeModel
+    client->>client: populate view
+    User->>client: open view menu
+    loop while descriptor in propertyDescriptors
+        client->>client: create menu entry (using descriptor.name, descriptor.description, descriptor.propertyId)
+    end
+    User->>client: select "Go To Max" menu item
+    client->>server: POST /experiment/{expUuid}/outputs/{outputId}/data/properties/{propertyId}
```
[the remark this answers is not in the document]
bhufmann: It should be described in the TSP. In the queryParams of property descriptor only query strings are allowed, that already exists for given data provider type. These allowed strings have a predefined meaning like in regular queries (e.g. query state). For example, requested_items` would be used for entry ids (row ids), `requested_time_range` for time ranges etc.

I thought about that for a while and I think it make sense to do it that way.

doc/adr/0007-queryable-properties.md
bhufmann: yep, copy paste error.

doc/adr/0007-queryable-properties.md
bhufmann: yes, it's the front-end, but not limited to this client. I think it's ok to be specific here because the theia-trace-extension fornt-end will be reference implementation for it.

bhufmann (commented):
@frallax thanks for the feedback. I'm looking forward to your counter proposal.

doc/adr/0007-queryable-properties.md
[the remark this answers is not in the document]
frallax: Ok, I think I understand, but I think I understood the idea just by reverse engineering [this code]. I think that in order to make the ADR reviews related to these kind of complex functionalities it would be needed a clear link between ADR and proposed TSP changes.

I.e. I find multiple ADRs on the topic of queryableparams, go to min max, etc, but I have not found a TSP change proposal. That is what created confusion for me.

In other words, we probably need a bit more "structured" way of discussing changes with ADRs, but we are learning :)

[this code]:https://git.eclipse.org/r/c/tracecompass/org.eclipse.tracecompass/+/196970/1/analysis/org.eclipse.tracecompass.analysis.profiling.core/src/org/eclipse/tracecompass/internal/analysis/profiling/core/callgraph/CallGraphStatisticsDataProviderFactory.java#38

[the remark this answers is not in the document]
bhufmann: This could was a proof of concept to see if the proposal in the ADR is working and it supposed to help with the ADR.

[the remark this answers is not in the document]
frallax: Yes, and actually it helped when I found it. What I was proposing here is just to try to collect all info in the same place for the next time (e.g. giving links in this ADR).
Otherwise it can be difficult to understand that your proposal here covers what is also described in [this other ADR].

In addition, I still not find the TSP update, i.e. the proposal on how to change the [trace-server-protocol] based on your ADR.

I am not criticizing, I am just trying to give some suggestion so that we can make this cool ADR process work as good as possible, I really like it!

[this other ADR]: https://github.com/eclipse-cdt-cloud/theia-trace-extension/pull/882
[trace-server-protocol]: https://github.com/eclipse-cdt-cloud/trace-server-protocol

doc/adr/0007-queryable-properties.md
frallax: For some reasons I was imagining the [tsp-python-client] to be the reference implementation. I think it is much more intuitive, easy to understand and portable compared to the theia-trace-extension UI. However, independently on this, I think it will improve the understanding if you can update the word "tool" to "theia-trace-extension" in case you'll do a follow up commit.

[tsp-python-client]: https://github.com/eclipse-cdt-cloud/tsp-python-client

doc/adr/0007-queryable-properties.md
frallax: Unfortunately I am not fully sure that I have the full understanding of all use cases. I'll try my best to come up with a alternative proposal, but it can take some time, and I do not want to block this proposal if other contributors understand it better than me / feel confident about it.

My greatest concern is that the TSP would become less intuitive to understand, and that possible collaborators would prefer other solutions (e.g. SQL based [perfetto-trace-analysis]) instead of taking advantage of the trace-server features just because the API is too "clunky". That is why I am asking for a pull request in [trace-server-protocol] (connected to this ADR) just to get a better understanding of which endpoints it will impact and how the API will look like if we go this way.

I also would like to make sure that the API we propose will take in consideration:
- if we want to use what you propose here to replace [time-graph-filtering] PR (i.e. use "filter" as one of the possible queryparameters)
- if it makes sense to create our own custom "chart / graph / table query logic", or if we can use something similar to [graphQL]
- that the API we propose is contained in what the trace-server is actually providing, i.e. "data representing charts/graph/tables", and that the API does not diverge too much from that by adding too exotic concepts

[perfetto-trace-analysis]: https://perfetto.dev/docs/quickstart/trace-analysis
[trace-server-protocol]: https://github.com/eclipse-cdt-cloud/trace-server-protocol
[time-graph-filtering]: https://github.com/eclipse-cdt-cloud/theia-trace-extension/pull/909
[graphQL]: https://github.com/eclipse-cdt-cloud/theia-trace-extension/blob/master/doc/adr/0002-GraphQL.md

MatthewKhouzam (commented):
My lightweight review. This covers all of our needs, going SQL / GraphQL may be good for other protocols as a side-band to the TSP. @frallax  I am looking forward to seeing a counter, but I would argue that this direction is a very good and clear one. Maybe I'm biased. :)

doc/adr/0007-queryable-properties.md
bhufmann: > Unfortunately I am not fully sure that I have the full understanding of all use cases. I'll try my best to come up with a alternative proposal, but it can take some time, and I do not want to block this proposal if other contributors understand it better than me / feel confident about it.
> 
I tried to come up with something so that we don't have to come-up with new APIs and endpoints for the different use cases. Maybe it's trying to cover too many cases. Maybe we should just look at the go to min and max and disregard the others. We can put a time range it the tooltip and make it click-able. It could have a hard-carded key value or pattern. 

So, I'm all for simplifying the proposal. I'm looking forward to seeing your alternative proposal. 

> My greatest concern is that the TSP would become less intuitive to understand, and that possible collaborators would prefer other solutions (e.g. SQL based [perfetto-trace-analysis](https://perfetto.dev/docs/quickstart/trace-analysis)) instead of taking advantage of the trace-server features just because the API is too "clunky". That is why I am asking for a pull request in [trace-server-protocol](https://github.com/eclipse-cdt-cloud/trace-server-protocol) (connected to this ADR) just to get a better understanding of which endpoints it will impact and how the API will look like if we go this way.
> 

The ADR has already that information. The sequence diagrams and the data structure definition shows what is added and affected.

> I also would like to make sure that the API we propose will take in consideration:
> 
>     * if we want to use what you propose here to replace [time-graph-filtering](https://github.com/eclipse-cdt-cloud/theia-trace-extension/pull/909) PR (i.e. use "filter" as one of the possible queryparameters)
> 

This ADR is not hand;ing the same thing than the other ADR and hence it doesn't replace the other ADR.
>     * if it makes sense to create our own custom "chart / graph / table query logic", or if we can use something similar to [graphQL](https://github.com/eclipse-cdt-cloud/theia-trace-extension/blob/master/doc/adr/0002-GraphQL.md) (and note that here I am talking about embedding graphQL in the trace-server, not to create a "parallel" implementation as the ADR [graphQL](https://github.com/eclipse-cdt-cloud/theia-trace-extension/blob/master/doc/adr/0002-GraphQL.md) proposes)
> 

I'm not clear on how to use GraphQL for this. Please provide a proposal for that so that I can understand.
>     * that the API we propose is contained in what the trace-server is actually providing, i.e. "data representing charts/graph/tables", and that the API does not diverge too much from that by adding too exotic concepts

Still it needs to provide a way to customize for getting additional information.

bhufmann:
Closing PR for this ADR since no agreement has been achieved on the API. It can be reopened if necessary.

Aew ADR will be provided with smaller scope to achieve `goto min and max` duration from latency views.

PR #855, opened 2022-10-17
Add configurable default view for the trace overview

### LINKED ISSUE #840: Default data provider of Overview View is hard-coded
The default data provider ID of the `Overview View` is hard-coded to the histogram XY chart of the Trace Compass trace server. The default data provider ID should not be hard-coded, but should be configurable using a preference.

Note: The end-user has now way to know what the IDs of the available XY views are. So, the ID is probably not the best thing to have a preference for.

hoangphamEclipse:
Currently, the trace overview default view is hard-coded in the trace viewer. This commit allows users/developers to set an XY view as the default view for the trace overview (using its name).

Users can set the default view for the trace overview using the Preference UI (File > Preferences > Open Settings or Ctrl + ,).

Developers can also set the default view for a specific Theia build by adding the "trace Viewer.trace Overview.defaultView" property to the "preferences" object in the package.json file (see snippet below - similar to `trace-viewer.port`) of the build and indicate the name of the default view.

https://github.com/eclipse-cdt-cloud/theia-trace-extension/blob/38d4e175718746f368e2ba2e3669e76bdcd40ec9/examples/browser/package.json#L5-L15

Fixes #840.

### DIFF
--- a/theia-extensions/viewer-prototype/src/browser/trace-overview-preference.ts
+++ b/theia-extensions/viewer-prototype/src/browser/trace-overview-preference.ts
@@ -1,0 +1,35 @@
+import { PreferenceProxy, PreferenceSchema, PreferenceScope } from '@theia/core/lib/browser';
+
+export const TRACE_OVERVIEW_DEFAULT_VIEW_KEY = 'trace Viewer.trace Overview.defaultView';
+export const DEFAULT_OVERVIEW_OUTPUT_NAME = 'Histogram';
+
+export function getSwitchToDefaultViewErrorMessage(preferredName: string, defaultName: string): string {
+    return `The ${preferredName} view cannot be opened as the trace overview. ` +
+    `Opening ${defaultName} instead. ` +
+    'Please set the Default View preference (Ctrl + ,) of the Trace Overview to an XY view, ' +
+    'or make sure the name is spelled correctly.';
+}
+
+export function getOpenTraceOverviewFailErrorMessage(): string {
+    return 'An error has occurred while opening the trace overview.';
+}
+
+export const OverviewSchema: PreferenceSchema = {
+    scope: PreferenceScope.Folder,
+    type: 'object',
+    properties: {
+        [TRACE_OVERVIEW_DEFAULT_VIEW_KEY]: {
+            type: 'string',
+            default: DEFAULT_OVERVIEW_OUTPUT_NAME,
+            description: 'Specify the name of the view that will be used as the default view for the Trace Overview. ' +
+            'Use the same name displayed in the Available Views list. E.g: For the Histogram view, enter Histogram.'
+        }
+    },
+};
+
+interface OverviewContribution {
+    [TRACE_OVERVIEW_DEFAULT_VIEW_KEY]: string
+}
+
+export const OverviewPreferences = Symbol('OverviewPreferences');
+export type OverviewPreferences = PreferenceProxy<OverviewContribution>;
--- a/theia-extensions/viewer-prototype/src/browser/trace-viewer/trace-viewer.tsx
+++ b/theia-extensions/viewer-prototype/src/browser/trace-viewer/trace-viewer.tsx
@@ -21,6 +21,8 @@ import { BackendFileService } from '../../common/backend-file-service';
 import { CancellationTokenSource } from '@theia/core';
 import * as React from 'react';
 import 'animate.css';
+import { DEFAULT_OVERVIEW_OUTPUT_NAME, TRACE_OVERVIEW_DEFAULT_VIEW_KEY,
+    getOpenTraceOverviewFailErrorMessage, getSwitchToDefaultViewErrorMessage, OverviewPreferences } from '../trace-overview-preference';
 
 export const TraceViewerWidgetOptions = Symbol('TraceViewerWidgetOptions');
 export interface TraceViewerWidgetOptions {
@@ -32,7 +34,6 @@ export interface TraceViewerWidgetOptions {
 export class TraceViewerWidget extends ReactWidget implements StatefulWidget {
     static ID = 'trace-viewer';
     static LABEL = 'Trace Viewer';
-    static DEFAULT_OVERVIEW_DATA_PROVIDER_ID = 'org.eclipse.tracecompass.internal.tmf.core.histogram.HistogramDataProvider';
 
     protected uri: Path;
     protected openedExperiment: Experiment | undefined;
@@ -83,6 +84,7 @@ export class TraceViewerWidget extends ReactWidget implements StatefulWidget {
     @inject(TraceExplorerContribution) protected readonly traceExplorerContribution: TraceExplorerContribution;
     @inject(WidgetManager) protected readonly widgetManager!: WidgetManager;
     @inject(ThemeService) protected readonly themeService: ThemeService;
+    @inject(OverviewPreferences) protected overviewPreferences: OverviewPreferences;
 
     @postConstruct()
     async init(): Promise<void> {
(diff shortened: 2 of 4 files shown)


hoangphamEclipse:
This commit add the cross-env npm package to set the environment variable OVERVIEW_DEFAULT that acts as the default view name if no user preference is set. Accidentally (in a good way), that this allows the TRACE_SERVER_PATH variable to be set in Windows. This means users on Windows does not have to edit the package.json of the browser example to start the application anymore.

examples/browser/package.json
```diff
@@ -36,7 +36,7 @@
   },
   "scripts": {
     "prepare": "theia build --mode development",
-    "start": "TRACE_SERVER_PATH=../../trace-compass-server/tracecompass-server theia start --plugins=local-dir:../plugins",
+    "start": "cross-env TRACE_SERVER_PATH=../../trace-compass-server/tracecompass-server OVERVIEW_DEFAULT=Histogram theia start --plugins=local-dir:../plugins",
```
bhufmann, line 39: We don't need to use an environment variable for OVERVIEW_DEFAULT. A preference is sufficient.

Btw, the environment variable for the trace server path exists to be able to package the trace server application when building the electron app. There is no need for an environment variable for the default overview.

theia-extensions/viewer-prototype/src/browser/trace-overview-preference.ts
```diff
@@ -0,0 +1,33 @@
...
+    return 'An error has occurred while opening the trace overview.';
+}
+
+export const OverviewSchema: PreferenceSchema = {
+    scope: PreferenceScope.Folder,
+    type: 'object',
+    properties: {
+        [TRACE_OVERVIEW_DEFAULT_VIEW]: {
+            type: 'string',
+            description: 'Specify the name of the view that will be used as the default view for the Trace Overview. ' +
```
bhufmann, line 22: set the `default` value to histogram here:

bhufmann (dismissed):
Thanks for the PR. Initial review comments.

theia-extensions/viewer-prototype/src/browser/trace-overview-preference.ts
hoangphamEclipse: Wouldn't this make the default value a hard coded value? Would it still work with different server builds that do not have the Histogram view?

bhufmann: Not really. When building a Theia base application, you can specify configure the default in the package.json of that application.

Rodrigoplp-work (approved):

theia-extensions/viewer-prototype/src/browser/trace-overview-preference.ts
```diff
@@ -0,0 +1,35 @@
+import { PreferenceProxy, PreferenceSchema, PreferenceScope } from '@theia/core/lib/browser';
+
+export const TRACE_OVERVIEW_DEFAULT_VIEW_KEY = 'trace Viewer.trace Overview.defaultView';
+export const DEFAULT_OVERVIEW_OUTPUT_NAME = 'Histogram';
+
+export function getSwitchToDefaultViewErrorMessage(preferredName: string, defaultName: string): string {
+    return `The ${preferredName} view cannot be opened as the trace overview. ` +
+    `Opening ${defaultName} instead. ` +
+    'Please set the Default View Preference (Ctrl + ,) of the Trace Overview to an XY view, ' +
```
marco-miller, line 9: Default View preference (would lower-case the `p`).

theia-extensions/viewer-prototype/src/browser/trace-overview-preference.ts
```diff
@@ -0,0 +1,35 @@
...
+    'or make sure the name is spelled correctly. ';
```
marco-miller, line 10: `correctly.';` (Extra trailing space, would remove it.)

theia-extensions/viewer-prototype/src/browser/trace-overview-preference.ts
```diff
@@ -0,0 +1,35 @@
...
+
+export const OverviewSchema: PreferenceSchema = {
+    scope: PreferenceScope.Folder,
+    type: 'object',
+    properties: {
+        [TRACE_OVERVIEW_DEFAULT_VIEW_KEY]: {
+            type: 'string',
+            default: DEFAULT_OVERVIEW_OUTPUT_NAME,
+            description: 'Specify the name of the view that will be used as the default view for the Trace Overview. ' +
+            ' Use the same name displayed in the Available Views list. E.g: For the Histogram view, enter Histogram.'
```
marco-miller, line 25: `'Use` (As there is already a space added to the previous line, above.)

theia-extensions/viewer-prototype/src/browser/trace-viewer/trace-viewer.tsx
```diff
@@ -583,7 +585,26 @@ export class TraceViewerWidget extends ReactWidget implements StatefulWidget {
...
+        const preferredViewName: string | undefined = this.overviewPreferences[TRACE_OVERVIEW_DEFAULT_VIEW_KEY];
+        let descriptor: OutputDescriptor | undefined = availableDescriptors?.find(output => output.name.toLowerCase() === preferredViewName?.toLowerCase());
+
+        // Failed to find an output that matches the user preference, fallback to default
+        if (!descriptor) {
+            const defaultViewName = DEFAULT_OVERVIEW_OUTPUT_NAME;
+            descriptor = availableDescriptors?.find(output => output.name.toLowerCase() === defaultViewName?.toLowerCase());
+
+            if (descriptor && preferredViewName)
+                {this.messageService.error(getSwitchToDefaultViewErrorMessage(preferredViewName, descriptor.name));}
```
marco-miller, line 599: Curly braces would need file-consistent reformatting.

marco-miller (changes requested):
- Adding to these in-file comments, I'd update the commit message so it reflects the currently proposed implementation (with that default env-var removed, IIUC).
- Same for PR's heading comments, accordingly.
- Also, `Fixes: Issue #840` can be replaced with `Fixes #840.`, which should auto-close the said Issue upon PR merge. This footer should be in the commit message as well.

theia-extensions/viewer-prototype/src/browser/trace-viewer/trace-viewer.tsx
```diff
@@ -583,7 +585,27 @@ export class TraceViewerWidget extends ReactWidget implements StatefulWidget {
...
+        // First, check if there is a preferred view
+        const preferredViewName: string | undefined = this.overviewPreferences[TRACE_OVERVIEW_DEFAULT_VIEW_KEY];
+        let descriptor: OutputDescriptor | undefined = availableDescriptors?.find(output => output.name.toLowerCase() === preferredViewName?.toLowerCase());
+
+        // Failed to find an output that matches the user preference, fallback to default
+        if (!descriptor) {
+            const defaultViewName = DEFAULT_OVERVIEW_OUTPUT_NAME;
+            descriptor = availableDescriptors?.find(output => output.name.toLowerCase() === defaultViewName?.toLowerCase());
+
+            if (descriptor && preferredViewName){
```
marco-miller, line 598: `preferredViewName) {`, rather.

marco-miller (changes requested):
Almost there I think.

hoangphamEclipse:
Looks like the build is failing with the error `Errors downloading some plugins.`.  with `error 429` (which I assume is HTTP). The error means `Too Many Requests`. Maybe the server hosting the plugins are down/too busy?

marco-miller:
> Looks like the build is failing with the error `Errors downloading some plugins.` with `error 429` (which I assume is HTTP). The error means `Too Many Requests`. Maybe the server hosting the plugins are down/too busy? May be we need to wait a bit?

I'll also try looking into this.

theia-extensions/viewer-prototype/src/browser/trace-overview-preference.ts
marco-miller: Marking as Unresolved to track the missing package.json how-to part, as discussed yesterday.

marco-miller:
@paul-marechal -Paul, would you be able to share your thoughts about this recurring build failure, please? This would be greatly appreciated. I tested the same commit using a dummy PR (#863), with that `vscode-builtin-yaml` line removed from `./package.json`, and it passed the build.

Should we "permanently" remove that line in here, or do we absolutely need it still, you think? Thanks for any clue you may have. I wonder if that plugin or version on Open VSX might have an issue. Previous yet recent builds did pass and didn't explicitly download that plugin it seems.

marco-miller:
> Should we "permanently" remove that line in here, or do we absolutely need it still, you think? Thanks for any clue you may have. I wonder if that plugin or version on Open VSX might have an issue. Previous yet recent builds did pass and didn't explicitly download that plugin it seems.

So this is a rate-limiting issue, as @hoangphamEclipse suspected and again based on the latest build log. @hoangphamEclipse, I think we need to look into confirming what the latest Open VXS throttling policy is. Then likely reduce the amount of plugins we download here. We might not need to download so many or all of them at once. Some might be removed without impacting anything; to be checked. Can you?

hoangphamEclipse:
Sure I will look into that.

hoangphamEclipse:
Looks like OpenVSX recently [implemented rate limiting](https://github.com/eclipse/openvsx/issues/502). [Documentation](https://github.com/eclipse/openvsx/issues/559) is still in progress.

hoangphamEclipse:
Updated commit message to include how to set the default view using package.json.

marco-miller (approved):
LGTM now, including the build flakiness which just passed. Thanks! @Rodrigoplp-work, letting you finish this review and rebase-merge at will. This should auto-close the related Issue, thanks to the GH linkeage in the commit message.

Rodrigoplp-work (approved):

PR #858, opened 2022-10-19
Toggle the filter tree for the trace overview

### LINKED ISSUE #787: Make the filter tree togglable or hidden for the trace overview.
The trace overview should have its filter tree togglable or hidden since its main purpose is to provide context for the whole trace.

hoangphamEclipse:
This commit creates a hamburger menu for the trace overview that will allow users to toggle the filter tree and select a different output view. The hamburger menu replaces the existing cog icon that is used for output source selection. A new state property is added to the AbstractTreeOutputComponent that allows showing and hiding the filter tree. The overview then toggles this state property.

Fix #787

### DIFF
--- a/packages/react-components/src/components/abstract-tree-output-component.tsx
+++ b/packages/react-components/src/components/abstract-tree-output-component.tsx
@@ -2,7 +2,11 @@ import { AbstractOutputComponent, AbstractOutputProps, AbstractOutputState } fro
 import * as React from 'react';
 import { ResponseStatus } from 'tsp-typescript-client/lib/models/response/responses';
 
-export abstract class AbstractTreeOutputComponent<P extends AbstractOutputProps, S extends AbstractOutputState> extends AbstractOutputComponent<P, S> {
+export type AbstractTreeOutputState = AbstractOutputState & {
+    showTree: boolean
+};
+
+export abstract class AbstractTreeOutputComponent<P extends AbstractOutputProps, S extends AbstractTreeOutputState> extends AbstractOutputComponent<P, S> {
 
     private readonly DEFAULT_Y_AXIS_WIDTH = 40;
     private readonly DEFAULT_SASH_WIDTH = 4;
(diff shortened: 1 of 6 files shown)


packages/react-components/src/components/abstract-tree-output-component.tsx
```diff
@@ -12,18 +16,20 @@ export abstract class AbstractTreeOutputComponent<P extends AbstractOutputProps,
 
     constructor(props: P) {
         super(props);
+
         this.onSashDown = this.onSashDown.bind(this);
         this.onSashMove = this.onSashMove.bind(this);
         this.onSashUp = this.onSashUp.bind(this);
+
```
marco-miller, line 23: Extra blank line introduced, would remove it.

packages/react-components/src/components/abstract-tree-output-component.tsx
```diff
@@ -39,6 +45,7 @@ export abstract class AbstractTreeOutputComponent<P extends AbstractOutputProps,
             }}>
                 {this.renderChart()}
             </div>
+
```
marco-miller, line 48: Extra blank line?

marco-miller (changes requested):
LGTM, with just a few nits.

marco-miller (dismissed):
Letting you @hoangphamEclipse decide if you need a second review; thanks, let me/us know.

Rodrigoplp-work (changes requested):
Awesome!
Could you add the name of the selected source to the title of the overview (the text that shows on the vertical bar on the left of the chart)?
So, after the user clicks on the hamburger button, clicks on "Select source output" and chooses for example "CPU Usage", the title would change from "Overview" to "Overview - CPU Usage".
If that is too much work we can leave it to be done on another ticket, as this was not specified in the requirements.

hoangphamEclipse:
> Awesome! Could you add the name of the selected source to the title of the overview (the text that shows on the vertical bar on the left of the chart)? So, after the user clicks on the hamburger button, clicks on "Select source output" and chooses for example "CPU Usage", the title would change from "Overview" to "Overview - CPU Usage". If that is too much work we can leave it to be done on another ticket, as this was not specified in the requirements.

This can be done in another PR.

Rodrigoplp-work (approved):

marco-miller (approved):
LGTM still, letting you @Rodrigoplp-work rebase/merge at will, if you tested this latest update. I didn't, but I tested an earlier version (of this PR).

PR #860, opened 2022-10-20
TableOutputComponent: Fix these search selection related cases

### LINKED ISSUE #859: Hitting Enter right after entering a search filter doesn't select the first entry in table(s)
1. Run trace-server.
2. Provision it with a trace or an experiment.
3. Run extension's `yarn start:browser`.
4. Browse to `http://localhost:3000/`.
5. Open the provisioned trace from the Trace Viewer.
6. Select the Events (or any other) Table for that trace.
7. Enter a Search filter in any one of the table columns.
8. Hit Enter right after entering the few filter characters.
9. Notice the wrong (non-first filtered) table entry being selected.

marco-miller:
This PR is made of these commits with more details in their respective commit messages:

1. Fix findMatchedEvent row 0 case.
2. Fix backward search upon index 0.
3. Fix selectStartIndex assigned to itself.
4. Wait for search before selecting.
5. Fix exception upon early Enter.
6. Reset navigating selection if new search.
7. Separate search and matching.
8. **Decrease debounce ms by 50x.**

**The latter (PR's head) commit** formally fixes #859.

### DIFF
--- a/packages/react-components/src/components/table-output-component.tsx
+++ b/packages/react-components/src/components/table-output-component.tsx
@@ -632,7 +632,7 @@ export class TableOutputComponent extends AbstractOutputComponent<TableOutputPro
                 this.selectEndIndex = this.selectStartIndex;
             }
             if (this.selectEndIndex !== -1 && this.selectStartIndex === -1) {
-                this.selectStartIndex = this.selectStartIndex;
+                this.selectStartIndex = this.selectEndIndex;
             }
 
             let currRowIndex = 0;
--- a/packages/react-components/src/components/table-output-component.tsx
+++ b/packages/react-components/src/components/table-output-component.tsx
@@ -42,6 +42,7 @@ export class TableOutputComponent extends AbstractOutputComponent<TableOutputPro
     private showIndexColumn = false;
     private frameworkComponents: any;
     private gridApi: GridApi | undefined = undefined;
+    private gridMatched = false;
     private gridRedrawn = false;
     private gridSearched = false;
     private columnApi: ColumnApi | undefined = undefined;
@@ -500,7 +501,6 @@ export class TableOutputComponent extends AbstractOutputComponent<TableOutputPro
                 this.selectRows();
             }
         }
-
     }
 
     private async fetchTableIndex(timestamp: bigint) {
@@ -537,7 +537,6 @@ export class TableOutputComponent extends AbstractOutputComponent<TableOutputPro
             }
         }
         return additionalParams;
-
     }
 
     private async fetchTableLines(fetchIndex: number, linesToFetch: number) {
@@ -587,7 +586,12 @@ export class TableOutputComponent extends AbstractOutputComponent<TableOutputPro
         }
     };
 
-    private searchEvents(colName: string, filterValue: string) {
+    private async searchEvents(colName: string, filterValue: string) {
+        const msBetweenChecks = 100;
+        while (this.gridMatched) {
+            // wait for grid to be done being matched -elsewhere (concurrently)
+            await new Promise(cb => setTimeout(cb, msBetweenChecks));
+        }
         this.gridRedrawn = true;
         if (filterValue === '') {
             this.filterModel.delete(colName);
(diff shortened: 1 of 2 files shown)


packages/react-components/src/components/table-output-component.tsx
```diff
@@ -626,13 +639,28 @@ export class TableOutputComponent extends AbstractOutputComponent<TableOutputPro
...
         if (this.gridApi) {
+            const msBetweenChecks = 100;
+            while (this.gridMatched) {
+                // wait for grid to be done being matched -from elsewhere (concurrently)
+                await new Promise(cb => setTimeout(cb, msBetweenChecks));
+            }
+            while (this.gridRedrawn) {
+                // wait for grid to be done being redrawn -elsewhere (before assuming it)
+                await new Promise(cb => setTimeout(cb, msBetweenChecks));
+            }
```
Rodrigoplp-work, line 650: I couldn't trigger these awaits. I'm afraid they are never called.

packages/react-components/src/components/table-output-component.tsx
```diff
@@ -603,6 +614,8 @@ export class TableOutputComponent extends AbstractOutputComponent<TableOutputPro
             });
             this.gridApi.redrawRows();
         }
+        this.gridSearched = true;
+        this.gridRedrawn = false;
```
Rodrigoplp-work, line 618: Good thinking, but doesn't seem to be enough. The problem is that `searchEvents` is triggered by `gridApi` every time a character is typed, but it changes the nodes on each pass, which delays the following triggers and changes. This makes it unreliable to control the `gridRedrawn` flag.
Maybe if we stop changing the nodes with `rowNode.data['isMatched'] = isMatched;`, and control the matched rows with a new independent array, then searchEvents would become fast enough to allow using `gridRedrawn` here.

packages/react-components/src/components/table-output-component.tsx
```diff
@@ -674,7 +708,8 @@ export class TableOutputComponent extends AbstractOutputComponent<TableOutputPro
                     // non-contiguous row found, stop searching in cache
                     currRowIndexFound = false;
                 }
-                if (currRowIndexFound && !isFound && rowNode.rowIndex && rowNode.data && rowNode.data['isMatched']) {
+                // only checking 'rowNode.rowIndex' below makes its '=== 0' case false:
+                if (currRowIndexFound && !isFound && (rowNode.rowIndex || rowNode.rowIndex === 0) && rowNode.data && rowNode.data['isMatched']) {
```
Rodrigoplp-work, line 712: Still, `rowNode.data.isMatched` is giving false positives due to the issue with `searchEvents`.

Rodrigoplp-work (changes requested):
Clever attempt! It reduced the number of failures when pressing enter in a search field. But it still happens if I'm fast enough. I believe we have a problem in `searchEvents` which may require a better way to control matches.

packages/react-components/src/components/table-output-component.tsx
marco-miller: I propose to keep this in as (small) safety, until we can work on the more major refactoring you proposed.

packages/react-components/src/components/table-output-component.tsx
marco-miller: I think we should keep this promising yet major refactoring for another iteration or PR. It may unveil more issues while fixing the design at a broader scope.

packages/react-components/src/components/table-output-component.tsx
marco-miller: I understand, however this fix remains necessary within its smaller scope, I think.

marco-miller:
> Clever attempt! It reduced the number of failures when pressing enter in a search field. But it still happens if I'm fast enough. I believe we have a problem in `searchEvents` which may require a better way to control matches.

I reduced the debounce timeout to 5x smaller in the head commit. I wasn't able to be faster than that locally. Can you retest using this latest value, please? I hesitated in putting 0 in there, so I thought of 10 as the smallest sensible value.

Rodrigoplp-work:
> > Clever attempt! It reduced the number of failures when pressing enter in a search field. But it still happens if I'm fast enough. I believe we have a problem in `searchEvents` which may require a better way to control matches.
> 
> I reduced the debounce timeout to 5x smaller in the head commit. I wasn't able to be faster than that locally. Can you retest using this latest value, please? I hesitated in putting 0 in there, so I thought of 10 as the smallest sensible value.

`0` doesn't solve the problem, so I agree with `10`.
I also agree with merging this because of the improvements it brings, and trying to work more on this in the future.
To that goal I suggest removing the comment "fixes #859" from the commits, so that we keep it open to track the issue.

marco-miller:
> `0` doesn't solve the problem, so I agree with `10`. I also agree with merging this because of the improvements it brings, and trying to work more on this in the future. To that goal I suggest removing the comment "fixes #859" from the commits, so that we keep it open to track the issue.

Thanks. I amended the top commit message accordingly.

Rodrigoplp-work (approved):

PR #863, opened 2022-10-26
Add configurable default view for the trace overview

### LINKED ISSUE #840: Default data provider of Overview View is hard-coded

marco-miller:
Currently, the trace overview default view is hard-coded in the trace viewer. With this commit, the default view for the trace extension can be set as a Preference setting in Theia. By default, the default view is Histogram, but that can be changed for each Theia build using the package.json file.

Fixes #840.


marco-miller:
This is a test PR to check if able to pass the build compared to #855's latest failures.

PR #865, opened 2022-11-02
Display data as it is made available

### LINKED ISSUE #816: Progressively display data in charts when it arrives (support Status.RUNNING)
Long loading traces can take over a minute, the server may send "intermediate" information with the status "RUNNING". We should display this data, not just discard it.

Rodrigoplp-work:
Changes the overview chart and the timegraph chart to display data as it is being loaded.

Changes are more notable in the overview. For the timegraph chart, even though the "analysis running" message was replaced by the direct display of data being loaded, its duration is normally short.

To see its effect, one should set the overview to open the "CPU Usage" analysis (set on preferences), because the "CPU Usage" displays a chart before having the user to select any series.

The effect is evident when a new large trace that makes several calls to the `fetchTree` endpoint is opened. After it is loaded, the result of the calls is cached and all charts open quickly.

Other charts like the "Disk I/O View" do not show the data as it is being loaded because they need the user to make a selection first. Therefore, on those charts, the "analysis running" message was kept on display while loading.

Fixes #816

https://user-images.githubusercontent.com/88502808/199540228-7a116efe-9a03-4d4e-8e15-e8dbfbbaa4fa.mp4

### DIFF
--- a/packages/react-components/src/components/abstract-xy-output-component.tsx
+++ b/packages/react-components/src/components/abstract-xy-output-component.tsx
@@ -251,6 +251,7 @@ export abstract class AbstractXYOutputComponent<P extends AbstractOutputProps, S
     }
 
     async fetchTree(): Promise<ResponseStatus> {
+        this.viewSpinner(true);
         const parameters = QueryHelper.timeRangeQuery(this.props.range.getStart(), this.props.range.getEnd());
         const tspClientResponse = await this.props.tspClient.fetchXYTree(this.props.traceId, this.props.outputDescriptor.id, parameters);
         const treeResponse = tspClientResponse.getModel();
@@ -265,31 +266,27 @@ export abstract class AbstractXYOutputComponent<P extends AbstractOutputProps, S
                 } else {
                     columns.push({title: 'Name', sortable: true});
                 }
-                if (treeResponse.status === ResponseStatus.COMPLETED) {
-                    const checkedSeries = this.getAllCheckedIds(treeResponse.model.entries);
-                    this.setState({
-                        outputStatus: treeResponse.status,
-                        xyTree: treeResponse.model.entries,
-                        checkedSeries,
-                        columns
-                    });
-                } else {
-                    this.setState({
-                        outputStatus: treeResponse.status,
-                        xyTree: treeResponse.model.entries,
-                        columns
-                    });
-                }
+                const checkedSeries = this.getAllCheckedIds(treeResponse.model.entries);
+                this.setState({
+                    outputStatus: treeResponse.status,
+                    xyTree: treeResponse.model.entries,
+                    checkedSeries,
+                    columns
+                }, () => {
+                    this.updateXY();
+                });
             } else {
                 this.setState({
                     outputStatus: treeResponse.status
                 });
             }
+            this.viewSpinner(false);
             return treeResponse.status;
         }
         this.setState({
             outputStatus: ResponseStatus.FAILED
         });
+        this.viewSpinner(false);
         return ResponseStatus.FAILED;
     }
 
@@ -368,7 +365,6 @@ export abstract class AbstractXYOutputComponent<P extends AbstractOutputProps, S
     }
 
     protected chooseChart(): JSX.Element {
-
         const param: XYChartFactoryParams = {
             viewRange: this.getDisplayedRange(),
             allMax: this.state.allMax,
@@ -402,11 +398,6 @@ export abstract class AbstractXYOutputComponent<P extends AbstractOutputProps, S
     }
 
     private async updateXY(): Promise<void> {
-        if (document.getElementById(this.getOutputComponentDomId() + 'handleSpinner')) {
-            // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-            document.getElementById(this.getOutputComponentDomId() + 'handleSpinner')!.style.visibility = 'visible';
-        }
-
         let start = BigInt(0);
         let end = BigInt(0);
         const viewRange = this.getDisplayedRange();
@@ -419,7 +410,7 @@ export abstract class AbstractXYOutputComponent<P extends AbstractOutputProps, S
 
         const tspClientResponse = await this.props.tspClient.fetchXY(this.props.traceId, this.props.outputDescriptor.id, xyDataParameters);
         const xyDataResponse = tspClientResponse.getModel();
-        if (tspClientResponse.isOk() && xyDataResponse) {
+        if (tspClientResponse.isOk() && xyDataResponse?.model?.series) {
             const series = xyDataResponse.model.series;
             if (series.length !== 0 && series[0].style) {
                 // Rely on type set for the first series to conclude for all series, if many.
@@ -432,10 +423,12 @@ export abstract class AbstractXYOutputComponent<P extends AbstractOutputProps, S
                 this.buildXYData(series);
             }
         }
+    }
 
+    private viewSpinner(status: boolean): void {
         if (document.getElementById(this.getOutputComponentDomId() + 'handleSpinner')) {
             // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-            document.getElementById(this.getOutputComponentDomId() + 'handleSpinner')!.style.visibility = 'hidden';
+            document.getElementById(this.getOutputComponentDomId() + 'handleSpinner')!.style.visibility = status ? 'visible' : 'hidden';
         }
     }
 
--- a/packages/react-components/src/components/xy-output-overview-component.tsx
+++ b/packages/react-components/src/components/xy-output-overview-component.tsx
@@ -18,7 +18,7 @@ type XYOutputOverviewState = AbstractXYOutputState & {
     showModal: boolean
 };
 
-type XYoutputOverviewProps = AbstractOutputProps & {
+type XYOutputOverviewProps = AbstractOutputProps & {
     experiment: Experiment;
 };
 
@@ -27,13 +27,13 @@ type XYoutputOverviewProps = AbstractOutputProps & {
  * It will only be setting these properties.
  */
 
-export class XYOutputOverviewComponent extends AbstractXYOutputComponent<XYoutputOverviewProps, XYOutputOverviewState> {
+export class XYOutputOverviewComponent extends AbstractXYOutputComponent<XYOutputOverviewProps, XYOutputOverviewState> {
     private initialViewRangeStartPosition = 0;
     private initialViewRangeEndPosition = 0;
     private viewRangeMoveStartOffsetX = 0;
     private keyMapping: Map<string, XY_OUTPUT_KEY_ACTIONS>;
 
-    constructor(props: XYoutputOverviewProps) {
+    constructor(props: XYOutputOverviewProps) {
         super(props);
         this.state = {
             outputStatus: ResponseStatus.RUNNING,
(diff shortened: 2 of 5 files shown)


Rodrigoplp-work:
> 1. For me locally, the CPU Usage view doesn't show a gradually loading trace with this applied. The behaviour looks like the one before this change. (Same for the Overview.)
> 2. And for the latter, the usual rotating arrow pair beside the message about running no longer shows (with this patch). That arrow pair still shows though for the CPU view.
> 3. Minor: there is a typo in the commit message; "as it (is) being".

1. The CPU Usage doesn't load gradually, unless it is set to be loaded on the Overview chart. After the trace is loaded and displayed on the Overview, the data is cached, so opening the CPU Usage shows the data right away, without the delay it takes to load it.
With a very big trace like the "teams" one, one could open the CPU Usage chart right after opening the trace, before the data is completely loaded. Still the CPU Usage chart won't load gradually, this time because of what I believe to be a limitation with ChartJS, that will render one chart at a time.
So, the fact that the CPU Usage chart doesn't open gradually is not a bug.
But the Overview chart, if first configured to display CPU Usage, should load gradually like in the video posted above.
2. Good catch! Rotating arrows added back in.
3. Typo fixed. Thanks.

marco-miller:
> 1. The CPU Usage doesn't load gradually, unless it is set to be loaded on the Overview chart. After the trace is loaded and displayed on the Overview, the data is cached, so opening the CPU Usage shows the data right away, without the delay it takes to load it.

I was mislead by my reading of the commit message part that says "and the timegraph chart". Maybe mention Overview in its title, and clarify it too in that sentence (that says "and the timegraph chart")?

>    With a very big trace like the "teams" one, one could open the CPU Usage chart right after opening the trace, before the data is completely loaded. Still the CPU Usage chart won't load gradually, this time because of what I believe to be a limitation with ChartJS, that will render one chart at a time.

IIUC, Overview only loads gradually, no other view, so is this issue about Overview, or another view? If applicable, maybe clarify this in the commit message, too.

>    So, the fact that the CPU Usage chart doesn't open gradually is not a bug.

Issue #816 however doesn't limit its scope to Overview only, AFAIK.

>    But the Overview chart, if first configured to display CPU Usage, should load gradually like in the video posted above.
> 2. Good catch! Rotating arrows added back in.

Ack, this all works for me. Thanks.

> 3. Typo fixed. Thanks.

I still see that typo up there.

marco-miller (approved):
Thanks. Commit message has a line exceeding the 72-chars max length limit, the one that had that typo fixed in it. -Minor.

marco-miller:
> Thanks. Commit message has a line exceeding the 72-chars max length limit, the one that had that typo fixed in it. -Minor.

Optional comment to potentially apply, above. And at this stage, I think @MatthewKhouzam wanted to co-review, before merge.

MatthewKhouzam (approved):
Works for me, @bhufmann please review?

bhufmann (changes requested):
Thanks for this enhancement.

I wonder if the poll timeout should be increased. It feels that 500 ms to fetch tree and XY is takes a lot of network time, download size and causes a lot of redraws of the graphs. Maybe we start with 500ms, then increase it after 2 to 3 times up-to a maximum delay

The Overview view shows the grey shading over the graph while indexing, but it should be inverted.

MatthewKhouzam (changes requested):
I tried opening with 10 charts simultaneously, and the UI thread froze, sorry!

packages/react-components/src/components/timegraph-output-component.tsx
```diff
@@ -423,22 +423,9 @@ export class TimegraphOutputComponent extends AbstractTreeOutputComponent<Timegr
...
-                            className='fa fa-refresh fa-spin'
-                            style={{ marginRight: '5px' }}
-                        />
-                    )}
-                    {
-                        'Analysis running'
-                    }
-                </div>
-            }
+            <div id='timegraph-main' className='ps__child--consume' onWheel={ev => { ev.preventDefault(); ev.stopPropagation(); }} style={{ height: this.props.style.height }} >
```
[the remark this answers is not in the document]
Rodrigoplp-work: Yes, that is true. Currently, while the first chart is making calls to get data from the backend, other charts have to hold, and they appear empty on screen.

As you correctly pointed out in https://bugs.eclipse.org/bugs/show_bug.cgi?id=544021, further investigation has to be made on this.

packages/react-components/src/components/abstract-xy-output-component.tsx
```diff
@@ -269,6 +269,8 @@ export abstract class AbstractXYOutputComponent<P extends AbstractOutputProps, S
                     outputStatus: treeResponse.status,
                     xyTree: treeResponse.model.entries,
                     columns
+                }, () => {
+                    this.updateXY();
```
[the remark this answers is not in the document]
Rodrigoplp-work: `this._debouncedUpdateXY()` will prevent the chart to display the data progressively while it is being loaded. It will show the first part of it, freeze and wait until all the data is displayed.

Therefore, to have the effect of watching the data being rendered while it is being loaded, `updateXY()` has to be called directly here.

I added a check to avoid calling `updateXY()` from here in charts that do not render any series before the user clicks on any checkboxes.

packages/react-components/src/components/abstract-xy-output-component.tsx
```diff
@@ -269,6 +269,19 @@ export abstract class AbstractXYOutputComponent<P extends AbstractOutputProps, S
                     outputStatus: treeResponse.status,
                     xyTree: treeResponse.model.entries,
                     columns
+                }, () => {
+                    // Only call updateXY here when the chart has a default series to plot.
+                    // If it prompts the user to "select a checkbox to see the analysis results",
+                    // updateXY will be called later.
+                    switch (this.props.outputDescriptor.name) {
+                        case 'CPU Usage':
```
bhufmann, line 277: This is back-end dependent and can't be merged like that. I don't understand why the xy charts should only be updated for certain xy charts and not for others. Please note that some back-ends don't have CPU usage nor Memory Usage views. When I ran it, the histogram was not updated. There was one rendering of XY data, but not more with the large trace I used to test. 

The goal of the PR is to show XY data in the overview view while analysis is still running, independent what type of XY chart the user selected. Scope of the implementation beside functionality is also performance and UX. If the performance is not acceptable, then that needs to be addressed as well.

Please test with large traces where the analysis takes a long time (e.g. maybe > 1 minute).

bhufmann (changes requested):

packages/react-components/src/components/abstract-xy-output-component.tsx
Rodrigoplp-work: Check for 'CPU Usage' removed.

packages/react-components/src/components/timegraph-output-component.tsx
```diff
@@ -422,22 +422,9 @@ export class TimegraphOutputComponent extends AbstractTreeOutputComponent<Timegr
...
```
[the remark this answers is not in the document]
Rodrigoplp-work: Changes removed from this PR.

Rodrigoplp-work:
It was failing for very large traces.

I added a throttle to the updateXY() function that avoids too close apart repeated calls to the fetchXY end point. Now calls are done in 3 seconds intervals.

packages/react-components/src/components/xy-output-overview-component.tsx
```diff
@@ -9,16 +9,15 @@ import { Experiment } from 'tsp-typescript-client';
 import { TraceOverviewSelectionDialogComponent } from './trace-overview-selection-dialog-component';
 
 const COLOR = {
-    SELECTION_RANGE: '#259fd8',
-    VIEW_RANGE: '#cccccc'
+    SELECTION_RANGE: '#259fd8'
```
bhufmann, line 12: why was VIEW_RANGE removed? By the way, the coloring in the overview view is broken. Is it related with this change?

packages/react-components/src/components/abstract-xy-output-component.tsx
```diff
@@ -381,12 +385,27 @@ export abstract class AbstractXYOutputComponent<P extends AbstractOutputProps, S
...
+     * this.props.tspClient.fetchXY(), all very close to each other, which can
+     * lead to errors.
+     * updateXY() is called repeatedly so that the user can see that data is being
+     * loaded, which would no tbe possible if using debounce.
+     */
+    private async recursiveFetch(): Promise<void> {
+        if (!this.isFetching) {
+            this.isFetching = true;
+            const timeout = 3000;
+            while (this.props.experiment?.indexingStatus === ResponseStatus.RUNNING) {
```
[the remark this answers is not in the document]
Rodrigoplp-work: Changed to check for both indexing and analysis.

packages/react-components/src/components/xy-output-overview-component.tsx
Rodrigoplp-work: It was meant to make the "analysis running" message more easily distinguishable from the background, and to use a color that changes according to the Theia theme instead of a harcoded value.

But yes, it makes the shaded areas used when selecting with the right button of the mouse disappear. So I reverted this change to use the hardcoded value.

Rodrigoplp-work:
> I suggest to update the method `waitAnalysisCompletion` in abstract-tree-output-component.tsx as follows:
> 
> ```js
>     protected async waitAnalysisCompletion(): Promise<void> {
>         let outputStatus = this.state.outputStatus;
>         let factor = 1.0;
>         const maxFactor = 10;
>         while (this.state && outputStatus === ResponseStatus.RUNNING) {
>             outputStatus = await this.fetchTree();
>             const timeout = 500 * factor;
>             await new Promise(resolve => setTimeout(resolve, timeout));
>             factor = factor > maxFactor ? factor: factor + 1;
>             // alternative: factor = factor > maxFactor ? factor = 1.0 : factor + 1;
>         }
>     }
> ```
> 
> With this we increase the time between queries and avoid requesting data too often. We can play with the `maxFactor` value and/or use the alternative code in comment.
> 
> Then you can request XY data after each `fetchTree()`

This is indeed good to remove the error for over calling the backend, but the update of the chart nearly stops after a few seconds.

I find the `recursiveFetch()` a little better at showing consistent progress, and it removes the need to empirically find a value for maxFactor.

I suggest keeping both the new `waitAnalysisCompletion` and the `recursiveFetch` methods.

packages/react-components/src/components/abstract-xy-output-component.tsx
```diff
@@ -381,12 +385,28 @@ export abstract class AbstractXYOutputComponent<P extends AbstractOutputProps, S
...
+     * this.props.tspClient.fetchXY(), all very close to each other, which can
+     * lead to errors.
+     * updateXY() is called repeatedly so that the user can see that data is being
+     * loaded, which would no tbe possible if using debounce.
+     */
+    private async recursiveFetch(): Promise<void> {
+        if (!this.isFetching) {
+            this.isFetching = true;
+            const timeout = 3000;
+            while (this.props.experiment?.indexingStatus === ResponseStatus.RUNNING ||
```
bhufmann, line 400: indexingStatus is independent from the analysis execution. You only need to check the outputStatus

bhufmann:
> > I suggest to update the method `waitAnalysisCompletion` in abstract-tree-output-component.tsx as follows:
> > ```js
> >     protected async waitAnalysisCompletion(): Promise<void> {
> >         let outputStatus = this.state.outputStatus;
> >         let factor = 1.0;
> >         const maxFactor = 10;
> >         while (this.state && outputStatus === ResponseStatus.RUNNING) {
> >             outputStatus = await this.fetchTree();
> >             const timeout = 500 * factor;
> >             await new Promise(resolve => setTimeout(resolve, timeout));
> >             factor = factor > maxFactor ? factor: factor + 1;
> >             // alternative: factor = factor > maxFactor ? factor = 1.0 : factor + 1;
> >         }
> >     }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > With this we increase the time between queries and avoid requesting data too often. We can play with the `maxFactor` value and/or use the alternative code in comment.
> > Then you can request XY data after each `fetchTree()`
> 
> This is indeed good to remove the error for over calling the backend, but the update of the chart nearly stops after a few seconds.
> 
> I find the `recursiveFetch()` a little better at showing consistent progress, and it removes the need to empirically find a value for maxFactor.
> 
> I suggest keeping both the new `waitAnalysisCompletion` and the `recursiveFetch` methods.

The point of the change in `waitAnalysisCompletion` was to be able to fetch the xy data after it fetch data returns and have better control when and how often the backend is queried during analysis execution. I'm not really for having another another independent while-loop with `recursiveFetch`.

packages/react-components/src/components/abstract-xy-output-component.tsx
```diff
@@ -269,12 +271,15 @@ export abstract class AbstractXYOutputComponent<P extends AbstractOutputProps, S
                     outputStatus: treeResponse.status,
                     xyTree: treeResponse.model.entries,
                     columns
+                }, () => {
+                    this.recursiveFetch();
```
bhufmann, line 275: see related comment https://github.com/eclipse-cdt-cloud/theia-trace-extension/pull/865#issuecomment-1380572897

bhufmann (changes requested):

packages/react-components/src/components/abstract-xy-output-component.tsx
bhufmann: I don't how it happened, but when testing it the while loop in `recursiveFetch()` never finishes. Maybe `props.experiment` is never updated?

packages/react-components/src/components/abstract-xy-output-component.tsx
Rodrigoplp-work: Checks for indexing status removed.

packages/react-components/src/components/abstract-xy-output-component.tsx
Rodrigoplp-work: I couldn't find a case where recursiveFetch fails to finish, but removed it as requested in https://github.com/eclipse-cdt-cloud/theia-trace-extension/pull/865#issuecomment-1380572897

bhufmann (changes requested):
When I run the latest version the spinner never goes away, even if analysis is finished. Please fix it.
image

Secondly, please rebase to latest in master before re-submitting it.

I'll continue reviewing it after these updates.

Thanks

Rodrigoplp-work:
> When I run the latest version the spinner never goes away, even if analysis is finished. Please fix it. image
> 
> Secondly, please rebase to latest in master before re-submitting it.
> 
> I'll continue reviewing it after these updates.
> 
> Thanks

Spinner bug fixed, and PR rebased.

bhufmann (changes requested):

packages/react-components/src/components/abstract-xy-output-component.tsx
```diff
@@ -272,24 +273,30 @@ export abstract class AbstractXYOutputComponent<P extends AbstractOutputProps, S
...
                         columns
+                    }, () => {
+                        this.updateXY();
                     });
                 } else {
                     this.setState({
                         outputStatus: treeResponse.status,
                         xyTree: treeResponse.model.entries,
                         columns
+                    }, () => {
```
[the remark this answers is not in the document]
Rodrigoplp-work: "If" statement removed.

bhufmann (approved):
Looks good to me. Thanks!

MatthewKhouzam (approved):
No longer seems to explode with 10 views open! :)

PR #867, opened 2022-11-07
Fixes large tables on high-dpi monitors

### LINKED ISSUE #862: Large tables (> 1 M rows) still broken on monitors with high res and high DPI
Traces with large tables (more than 1 million rows) have problem with AGGrid, rendering the last rows unresponsive (no hovering nor selection works).
This problem was solved by using AGGrid pagination feature. The size of the page was empirically set to 500,000 rows.
But on monitors with very high resolution (5120 x 2880 px and DPI scale = 2), the problems still happens.
Moving the browser to a second monitor with lower resolution (1920 x 1080 px and DPI scale = 1) solves the problem, not even needing to restart theia-trace-extension.
Suggestion: change the page size to 250,000, which makes it work on all types of monitors.

Rodrigoplp-work:
Ag-grid says tables with too many rows have problems because of the limitation of a div's size, which is defined by the browser (see https://www.ag-grid.com/javascript-data-grid/massive-row-count/).

In order to check what the limitation is, one has to set `debugFlag=true` in `table-output-component.tsx` and watch the console when a large table is loaded. Doing so in a computer with a small monitor (1920 x 1200 px, dpi scale = 1) heralds:

`AG Grid.RowContainerHeightService: maxDivHeight = 32,000,000`

If each row has a height of 32 px, that means the table will start presenting selection issues below the millionth row. In fact, we observed problems precisely with tables that large.

A browser on a larger monitor (5120 x 2880 px, dpi scale = 2) heralds instead:

`AG Grid.RowContainerHeightService: maxDivHeight = 16,000,000`

The same row height of 32 px will have the table misbehaving after 500,000 rows, which is what was observed.

All this is a consequence of the particular solution Ag-grid adopted to deal with large tables, a technique they call "stretching" and is explained in the link above.

To circumvent this issue, this PR changes the number of rows a page is allowed to have from 500,000 to 250,000.

Fixes https://github.com/eclipse-cdt-cloud/theia-trace-extension/issues/862.

Signed-off-by: Rodrigo Pinto <rodrigo.pinto@calian.ca

### DIFF
--- a/packages/react-components/src/components/table-output-component.tsx
+++ b/packages/react-components/src/components/table-output-component.tsx
@@ -41,7 +41,7 @@ export class TableOutputComponent extends AbstractOutputComponent<TableOutputPro
     private fetchColumns = true;
     private columnArray = new Array<any>();
     private pagination = true;
-    private paginationPageSize = 500000;
+    private paginationPageSize = 250000;
     private paginationTotalPages = 0;
     private showIndexColumn = false;
     private frameworkComponents: any;


marco-miller (approved):
LGTM but I use my base laptop screen, so my testing is limited for that specific Issue.

MatthewKhouzam:
I have a high dpi monitor, I confirm it works, I would like to have an explanation of why though.

MatthewKhouzam (approved):
We need to know why though.

bhufmann:
Can we merge this one?

marco-miller:
> Can we merge this one?

I think @MatthewKhouzam was expecting a documented (yet potentially brief) explanation first. Ideally persisted in the git commit message, or so I propose.

Rodrigoplp-work:
Found the reason behind the problem. Updated the commit message and the PR post.

marco-miller:
Good findings. There is a typo in the commit message and PR header; 'streching' should read 'stretching'.

marco-miller (changes requested):
Per previously posted comment; minor.

marco-miller (approved):
Thanks!

MatthewKhouzam (approved):

PR #876, opened 2022-11-10
Workaround rate limitation issues with open-vsx

bhufmann:
Download the plug-ins sequentially instead of parallel. This should avoid the failure "Download:plugins fails with 429".

see discussion:
https://community.theia-ide.org/t/download-plugins-fails-with-429/2659


marco-miller (approved):
LGTM -thanks!

PR #879, opened 2022-11-11
Markers sync

williamsyang-work:
This updates theia-trace-extension to use `TimeGraphMarkersChart` for the markers layer.  Also, sorts the markers so navigation emulates the behavior in the trace-compass application.


packages/react-components/src/components/timegraph-output-component.tsx
```diff
@@ -707,9 +714,87 @@ export class TimegraphOutputComponent extends AbstractTreeOutputComponent<Timegr
...
+        this.markersChartLayer.selectedMarker = undefined;
         this.setState({ markerCategoryEntries: annotationEntries, markerLayerData: markerLayerData });
     }
 
+    /**
+     * Sets next and prev for marker navigation with < & >.
+     * Directly modifies the input.
+     * @param rowsMap TimeGraphRowModel[]
+     */
+    private setNextAndPrevious = (rowsMap: TimelineChart.TimeGraphRowModel[]) => {
```
bhufmann, line 728: Not sure if we should do that to avoid unnecessary processing. I put a comment regarding this in https://github.com/eclipse-cdt-cloud/timeline-chart/pull/216.

bhufmann (changes requested):

PR #880, opened 2022-11-14
Fix pagination bar bug

### LINKED ISSUE #849: Show pagination bar of table-output-component during indexing
With the upgrade of AgGrid in #813 pagination is used if the total number of events is greater than the page size (=500000). When opening a trace the first time the total number of events not known and the back-end is queried for the number of events multiple times until indexing is finished. At the beginning the total number of events will be smaller and by design the pagination is disabled. However, when the number of events exceeds during indexing while table is open, then pagination bar should appear.

As work-around, the table can be closed and re-opened

Rodrigoplp-work:
With traces that have very large tables, if the event table is opened before the trace is fully loaded, the pagination bar does not appear.

The problem is that Ag-grid does not make the pagination bar reactive, so it only responds to the initial values of the table. Updates of the number of rows are ignored.

The solution is to create a custom pagination bar that can respond reactively to data being loaded.

Fixes #849


Rodrigoplp-work:
https://user-images.githubusercontent.com/88502808/201933893-1aa17b9e-1295-451f-ad0d-91c039639493.mp4

packages/react-components/src/components/table-output-component.tsx
```diff
@@ -38,7 +41,8 @@ export class TableOutputComponent extends AbstractOutputComponent<TableOutputPro
     private fetchColumns = true;
     private columnArray = new Array<any>();
     private pagination = true;
-    private paginationPageSize = 500000;
+    private paginationPageSize = 250000;
```
marco-miller, line 44: Should this size change be part of this PR? I mean, versus #867 which already applies this fix?

marco-miller (commented):
LGTM with the basic testing I did locally, with a question (hereby); thanks!

packages/react-components/src/components/table-output-component.tsx
Rodrigoplp-work: Indeed. Thanks. Fixed.

bhufmann (commented):
It looks good and works well. 

When testing this PR, I started wondering if we would like to just show the pagination bar of AgGrid all the time no matter if the number of events is smaller than the page size. The advantage is that user always will be able to see the total number of events in the trace. The disadvantage is that the bar takes real estate when not needed. 

I just wanted to bring up this alternative. But I'm ok with both solutions.

marco-miller (approved):
LGTM now; thanks.

PR #881, opened 2022-11-15
Add Arrow Up/Down and Page Up/Down support for table-output-component

### LINKED ISSUE #848: Support next/previous with keyboard across pages (pagination) in table-output-component
PR #813 upgrades AgGrid to a later version and it required to use pagination. However, selecting an event across pages using arrow down/up or page down/up is not switching to the next page.

hoangphamEclipse:
This commit adds support for using the Arrow Up/Down and Page Up/Down to navigate across pages for views that implements the table-output-component. Previously, user can only use these keys to navigate across a page, but can not move forward or backward a page.

Fixes #848.


hoangphamEclipse:
Marked as draft due to the following issue: 

UI/UX wise, when user switch pages, we should focus on the cell that is 1) in the same column as previously selected and 2) on the first and last row when we are going forward and backward a page, respectively. I still cannot get this working. I tried using gridAPI.setFocusedCell() but it failed.

PR #886, opened 2022-11-23
Search Time Events in Time Graph Views

hriday-panchasara:
- Added a dialog-box to timegraph-output-component to search for events
- The view updates its contents as the user enters the string
- The filter is applied and unsuccessful time events are dimmed

This patch depends on:
- https://git.eclipse.org/r/c/tracecompass.incubator/org.eclipse.tracecompass.incubator/+/200991
- https://github.com/eclipse-cdt-cloud/timeline-chart/pull/243

[Screen Recording 2022-11-23 at 11.45.00 AM.webm](https://user-images.githubusercontent.com/92893187/203631862-460e6275-9397-4959-a152-7c8120d40979.webm)

The time event filtering uses its own domain-specific language (DSL) . Various logical operators are suppported:
- && performs a logical AND operation between two filter expressions. Example: ioctl && 5755
- || performs a logical OR operation between two filter expressions. Example: ioctl || splice
- ! performs a logical NOT operation between two filter expressions. Example: !ioctl

Filter expressions have the following pattern field operator argument in general. Several relational operators are available:
- test whether the value of the specified field is equal to the given argument. Example: TID == 5755
- test whether the value of the specified field is not equal to the given argument. Example: TID != 5755
- matches test whether the value of the specified field matches to the given argument. Example: TID matches 5.*5$
- contains test whether the value of the specified field contains to the given argument. Example: TID contains 7
- test whether the value of the specified field is greater than the given argument. Example: "SOFT IRQ" > 4
- test whether the value of the specified field is less than the given argument. Example: "SOFT IRQ" < 4

Making use of Material UI to help create the search box. I found it much easier to use as compared to Semantic UI. Here's a mockup of a future patch that would allow users to save multiple filters - the chips are built using Material UI:
Screen Shot 2022-11-23 at 11 42 35 AM

### DIFF
--- a/packages/react-components/src/components/data-providers/tsp-data-provider.ts
+++ b/packages/react-components/src/components/data-providers/tsp-data-provider.ts
@@ -50,11 +50,13 @@ export class TspDataProvider {
     async getData(
         ids: number[],
         entries: TimeGraphEntry[],
+        fetchArrows: boolean,
         totalTimeRange: TimeRange,
         worldRange?: TimelineChart.TimeGraphRange,
         nbTimes?: number,
         annotationMarkers?: string[],
-        markerSetId?: string
+        markerSetId?: string,
+        additionalProperties?: { [key: string]: any }
     ): Promise<TimelineChart.TimeGraphModel> {
         this.timeGraphEntries = [...entries];
         if (!this.timeGraphEntries.length || !worldRange || !nbTimes) {
@@ -72,7 +74,13 @@ export class TspDataProvider {
         this.totalRange = totalTimeRange.getEnd() - totalTimeRange.getStart();
         const start = totalTimeRange.getStart() + worldRange.start;
         const end = totalTimeRange.getStart() + worldRange.end;
-        const timeGraphStateParams = QueryHelper.selectionTimeRangeQuery(start, end, nbTimes, ids);
+        const timeGraphStateParams = QueryHelper.selectionTimeRangeQuery(
+            start,
+            end,
+            nbTimes,
+            ids,
+            additionalProperties ? additionalProperties : {}
+        );
         const statesPromise = this.client.fetchTimeGraphStates(this.traceUUID, this.outputId, timeGraphStateParams);
 
         const additionalProps: { [key: string]: any } = {};
@@ -89,13 +97,11 @@ export class TspDataProvider {
         const arrowStart = worldRange.start + this.timeGraphEntries[0].start;
         const arrowEnd = worldRange.end + this.timeGraphEntries[0].start;
         const fetchParameters = QueryHelper.timeRangeQuery(arrowStart, arrowEnd, nbTimes);
-        const arrowsPromise = this.client.fetchTimeGraphArrows(this.traceUUID, this.outputId, fetchParameters);
 
         // Wait for responses
-        const [tspClientAnnotationsResponse, tspClientStatesResponse, tspClientArrowsResponse] = await Promise.all([
+        const [tspClientAnnotationsResponse, tspClientStatesResponse] = await Promise.all([
             annotationsPromise,
-            statesPromise,
-            arrowsPromise
+            statesPromise
         ]);
 
         // the start time which is normalized to logical 0 in timeline chart.
@@ -147,7 +153,15 @@ export class TspDataProvider {
             }
         }
 
-        const arrows = this.getArrows(tspClientArrowsResponse, worldRange, nbTimes);
+        let arrows: TimelineChart.TimeGraphArrow[] = [];
+        if (fetchArrows) {
+            const tspClientArrowsResponse = await this.client.fetchTimeGraphArrows(
+                this.traceUUID,
+                this.outputId,
+                fetchParameters
+            );
+            arrows = this.getArrows(tspClientArrowsResponse, worldRange, nbTimes);
+        }
 
         return {
             id: 'model',
@@ -218,16 +232,14 @@ export class TspDataProvider {
     }
 
     private getRowModel(row: TimeGraphRow, chartStart: bigint, rowId: number, entry: TimeGraphEntry) {
-        let gapStyle: OutputElementStyle;
-        if (!entry.style) {
-            gapStyle = this.getDefaultForGapStyle();
-        } else {
-            gapStyle = entry.style;
-        }
+        let dimGapStates = false;
         const states: TimelineChart.TimeGraphState[] = [];
         let prevPossibleState = entry.start;
         let nextPossibleState = entry.end;
         row.states.forEach((state: TimeGraphState, idx: number) => {
+            if (!dimGapStates && state.tags && state.tags === 1) {
+                dimGapStates = true;
+            }
             const end = state.end - chartStart;
             states.push({
                 id: row.entryId + '-' + idx,
@@ -237,7 +249,8 @@ export class TspDataProvider {
                     end
                 },
                 data: {
-                    style: state.style
+                    style: state.style,
+                    tags: state.tags
                 }
             });
             this.totalRange = this.totalRange < end ? end : this.totalRange;
@@ -248,6 +261,17 @@ export class TspDataProvider {
                 nextPossibleState = state.end - chartStart;
             }
         });
+        let gapStyle: OutputElementStyle;
+        if (!entry.style) {
+            gapStyle = this.getDefaultForGapStyle();
+        } else {
+            gapStyle = entry.style;
+        }
+        if (dimGapStates) {
+            gapStyle.values.opacity = 0.3;
+        } else if (gapStyle.values.opacity) {
+            delete gapStyle.values.opacity;
+        }
 
         return {
             id: rowId,
--- a/packages/react-components/src/components/timegraph-output-component.tsx
+++ b/packages/react-components/src/components/timegraph-output-component.tsx
@@ -32,6 +32,9 @@ import { isEqual } from 'lodash';
 import { convertColorStringToHexNumber } from 'traceviewer-base/lib/utils/convert-color-string-to-hex';
 import { faSpinner } from '@fortawesome/free-solid-svg-icons';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import TextField from '@mui/material/TextField';
+import InputAdornment from '@mui/material/InputAdornment';
+import { debounce } from 'lodash';
 
 type TimegraphOutputProps = AbstractOutputProps & {
     addWidgetResizeHandler: (handler: () => void) => void;
@@ -50,6 +53,7 @@ type TimegraphOutputState = AbstractTreeOutputState & {
     collapsedMarkerNodes: number[];
     columns: ColumnHeader[];
     dataRows: TimelineChart.TimeGraphRowModel[];
+    searchString: string;
 };
 
 const COARSE_RESOLUTION_FACTOR = 8; // resolution factor to use for first (coarse) update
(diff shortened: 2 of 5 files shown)


packages/react-components/src/components/timegraph-output-component.tsx
```diff
@@ -384,7 +398,7 @@ export class TimegraphOutputComponent extends AbstractTreeOutputComponent<Timegr
         // TODO Show header, when we can have entries in-line with timeline-chart
         return <>
             <div ref={this.timeGraphTreeRef} className='scrollable' onScroll={() => this.synchronizeTreeScroll()}
-                style={{ height: parseInt(this.props.style.height.toString()) - this.getMarkersLayerHeight() }}
+                style={{ height: parseInt(this.props.style.height.toString()) - this.getMarkersLayerHeight() - (document.getElementById(this.props.traceId + this.props.outputDescriptor.id + 'searchBar')?.offsetHeight ?? 0)}}
```
bhufmann, line 401: I get a linter warning here:
This line has a length of 224. Maximum allowed is 180.eslint[max-len](https://eslint.org/docs/rules/max-len)

packages/react-components/src/components/timegraph-output-component.tsx
```diff
@@ -504,9 +518,49 @@ export class TimegraphOutputComponent extends AbstractTreeOutputComponent<Timegr
         return <div id='main-timegraph-content' ref={this.horizontalContainer} style={{ height: this.props.style.height }} >
             {this.getChartContainer()}
             {this.getMarkersContainer()}
+            <div id={this.props.traceId + this.props.outputDescriptor.id + 'searchBar'} className='timgraph-search-bar'>
+                <TextField
+                    InputProps={{
+                        placeholder: "Search",
```
bhufmann, line 524: Single quotes: 'Search'

packages/react-components/src/components/timegraph-output-component.tsx
```diff
@@ -504,9 +518,49 @@ export class TimegraphOutputComponent extends AbstractTreeOutputComponent<Timegr
...
+                        startAdornment: (
+                            <InputAdornment
+                                sx={{
+                                    color: "var(--theia-ui-font-color0)"
+                                }}
+                                position="start"
```
bhufmann, line 530: single quotes: 'start'

packages/react-components/src/components/timegraph-output-component.tsx
```diff
@@ -504,9 +518,49 @@ export class TimegraphOutputComponent extends AbstractTreeOutputComponent<Timegr
...
             {this.getChartContainer()}
             {this.getMarkersContainer()}
+            <div id={this.props.traceId + this.props.outputDescriptor.id + 'searchBar'} className='timgraph-search-bar'>
+                <TextField
+                    InputProps={{
+                        placeholder: "Search",
+                        startAdornment: (
+                            <InputAdornment
+                                sx={{
+                                    color: "var(--theia-ui-font-color0)"
```
bhufmann, line 528: single quotes

packages/react-components/src/components/timegraph-output-component.tsx
```diff
@@ -504,9 +518,49 @@ export class TimegraphOutputComponent extends AbstractTreeOutputComponent<Timegr
...
+                                position="start"
+                            >
+                                <i className='codicon codicon-search'></i>
+                            </InputAdornment>
+                        ),
+                        className: "timegraph-search-box",
+                        endAdornment: (
+                            <InputAdornment 
+                                sx={{
+                                    color: "var(--theia-ui-font-color0)"
```
bhufmann, line 539: single quotes

packages/react-components/src/components/timegraph-output-component.tsx
```diff
@@ -504,9 +518,49 @@ export class TimegraphOutputComponent extends AbstractTreeOutputComponent<Timegr
...
+                            <InputAdornment
+                                sx={{
+                                    color: "var(--theia-ui-font-color0)"
+                                }}
+                                position="start"
+                            >
+                                <i className='codicon codicon-search'></i>
+                            </InputAdornment>
+                        ),
+                        className: "timegraph-search-box",
```
bhufmann, line 535: single quotes

packages/react-components/src/components/timegraph-output-component.tsx
```diff
@@ -504,9 +518,49 @@ export class TimegraphOutputComponent extends AbstractTreeOutputComponent<Timegr
...
+                    }}
+                    value={this.state.searchString}
+                    onChange={this.handleSearchChange}
+                />
+            </div>
         </div>;
     }
 
+    private handleSearchChange(event: React.ChangeEvent<HTMLInputElement>) {
+        this.setState({ searchString: event.target.value ?? '' });
```
bhufmann, line 557: I noticed that when typing an equal sign "=", the timeline-chart zooms out, because it's a keyboard shortcut. So, when the focus is on the textbox the key events must not be handled in the timeline-chart.

packages/react-components/src/components/timegraph-output-component.tsx
```diff
@@ -535,7 +589,7 @@ export class TimegraphOutputComponent extends AbstractTreeOutputComponent<Timegr
             options={
                 {
                     id: this.props.traceId + this.props.outputDescriptor.id + 'focusContainer',
-                    height: parseInt(this.props.style.height.toString()) - this.getMarkersLayerHeight(),
+                    height: parseInt(this.props.style.height.toString()) - this.getMarkersLayerHeight() - (document.getElementById(this.props.traceId + this.props.outputDescriptor.id + 'searchBar')?.offsetHeight ?? 0),
```
bhufmann, line 592: eslint warning:
This line has a length of 218. Maximum allowed is 180.eslint[max-len](https://eslint.org/docs/rules/max-len)

packages/react-components/src/components/data-providers/tsp-data-provider.ts
```diff
@@ -60,7 +60,13 @@ export class TspDataProvider {
         this.totalRange = totalTimeRange.getEnd() - totalTimeRange.getStart();
         const start = totalTimeRange.getStart() + worldRange.start;
         const end = totalTimeRange.getStart() + worldRange.end;
-        const timeGraphStateParams = QueryHelper.selectionTimeRangeQuery(start, end, nbTimes, ids);
+        
+        const additionalParams: { [key: string]: any } = {};
+        if (searchString && searchString.length > 0) {
+            additionalParams['regex_map_filters'] = {"1": [searchString]}
```
bhufmann, line 66: single quotes: '1'

packages/react-components/src/components/data-providers/tsp-data-provider.ts
```diff
@@ -60,7 +60,13 @@ export class TspDataProvider {
         this.totalRange = totalTimeRange.getEnd() - totalTimeRange.getStart();
         const start = totalTimeRange.getStart() + worldRange.start;
         const end = totalTimeRange.getStart() + worldRange.end;
-        const timeGraphStateParams = QueryHelper.selectionTimeRangeQuery(start, end, nbTimes, ids);
+
```
bhufmann, line 63: remove spaces

packages/react-components/src/components/data-providers/tsp-data-provider.ts
```diff
@@ -60,7 +60,13 @@ export class TspDataProvider {
...
         const start = totalTimeRange.getStart() + worldRange.start;
         const end = totalTimeRange.getStart() + worldRange.end;
-        const timeGraphStateParams = QueryHelper.selectionTimeRangeQuery(start, end, nbTimes, ids);
+        
+        const additionalParams: { [key: string]: any } = {};
+        if (searchString && searchString.length > 0) {
+            additionalParams['regex_map_filters'] = {"1": [searchString]}
+        }
+        const timeGraphStateParams = QueryHelper.selectionTimeRangeQuery(start, end, nbTimes, ids, additionalParams);
+
```
bhufmann, line 69: remove spaces

bhufmann (changes requested):
Hi @hriday-panchasara, thanks for the search feature in the time graph. I was able to run it successfully (with some server update). I've added inline comments as well as some global comments below:

- Please install eslint if you don't have already installed it
- I noticed that when using the Trace Compass server of Trace Compass incubator master, that I get a Java Class cast exception, when executing DataProviderProviderUtils.extractRegexFilter() because the key is an instance of String and not an instance of Integer (Entry<Integer, Collection<String>> entry : regexesMap.entrySet()). This is because of the Jersey default deserializer creates a String instead of an integers. So, going forward we need to have a custom deserializer and make sure that the map is actually an Integer -> Collection<String> map. When I tepmorarily updated the code in my development environment, I was able to successfully apply the filter. 
Please note that, depending on the Jersey version, this might or might not happen. If you didn't notice it then my guess is that you run the Trace Compass server of 2020-06 branch.
- Right not the text box is added at every timeline-chart. Before merging there needs to be a UX design for this, where the text box will be located, should it be per timeline-chart or globally or support both, open and close of text box needed etc.

bhufmann (changes requested):
My main focus was the review of the changes in the timeline-chart library. I noticed that the gap states are not dimmed which is required. I put a comment below for that.

packages/react-components/src/components/timegraph-output-component.tsx
```diff
@@ -821,7 +899,7 @@ export class TimegraphOutputComponent extends AbstractTreeOutputComponent<Timegr
                 }
                 return {
                     color: backgroundColor ? backgroundColor.color : 0x000000,
-                    opacity: backgroundColor ? backgroundColor.alpha : 1.0,
+                    opacity: (metadata.tags && metadata.tags === 1) ? 0.3 : (backgroundColor ? backgroundColor.alpha : 1.0),
```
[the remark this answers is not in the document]
hriday-panchasara: I've added opacity to gapStyle in my recent commit, but I'm not sure if its the right way to do so.

bhufmann (changes requested):
The gap seems to be applied correctly. However, when removing the filter the states seemed to still dimmed. Could you please look into this?

bhufmann:
> @hriday-panchasara The gap style seems to be applied correctly. However, when removing the filter the states seemed to still dimmed. Could you please look into this?

This is still an issue with the recent timeline-chart update of PR https://github.com/eclipse-cdt-cloud/timeline-chart/pull/243.

hriday-panchasara:
> > @hriday-panchasara The gap style seems to be applied correctly. However, when removing the filter the states seemed to still dimmed. Could you please look into this?
> 
> This is still an issue with the recent timeline-chart update of PR [eclipse-cdt-cloud/timeline-chart#243](https://github.com/eclipse-cdt-cloud/timeline-chart/pull/243).

Yup, I'm assuming its related to the timeline-chart repo. I'll keep all patches open until this is resolved

hriday-panchasara:
> > @hriday-panchasara The gap style seems to be applied correctly. However, when removing the filter the states seemed to still dimmed. Could you please look into this?
> 
> This is still an issue with the recent timeline-chart update of PR [eclipse-cdt-cloud/timeline-chart#243](https://github.com/eclipse-cdt-cloud/timeline-chart/pull/243).

I've added a small fix for this issue, could you let me know if it looks good?

bhufmann (changes requested):
This one needs a rebase as well.

bhufmann (changes requested):
changing the tick menu doesn't update the view right  away, a zoom action is needed. This doesn't happen in master and seems to be related with this and corresponding timeline-chart change.

bhufmann (changes requested):
Please squash both commits. Otherwise we introduce a commit that doesn't compile.

bhufmann (approved):
Looks good to me. Thanks!

PR #893, opened 2022-12-12
Add new command open-trace-with-path to open a trace with the given path

jonahgraham:
This new command takes a path to a trace directory and opens the trace in the trace viewer.

The use case is we have an extension that collects traces from target hardware, when that trace collection is complete we want to open it in the UI. Adding this new command allows us to open the trace in the viewer by passing the path to the new trace.

The alternative we considered was, in our extension, connecting to the trace server and issuing a tspClient.openTrace and then tspClient.createExperiment. While this does load the trace on the server, we still need a way for the UI to refresh. Rather the providing a new command to refresh the UI we propose this new command as it also saves us from duplicating all the work to interact with the tspClient.

### DIFF
--- a/theia-extensions/viewer-prototype/src/browser/trace-viewer/trace-viewer-contribution.ts
+++ b/theia-extensions/viewer-prototype/src/browser/trace-viewer/trace-viewer-contribution.ts
@@ -188,7 +188,7 @@ export class TraceViewerContribution extends WidgetOpenHandler<TraceViewerWidget
         });
         registry.registerCommand(OpenTraceWithPathCommand, {
             isVisible: () => false,
-            execute: (path: string, options: TraceViewerWidgetOpenerOptions) => path && this.open(new URI(path), options),
+            execute: (path: string) => path && this.open(new URI(path)),
         });
         registry.registerCommand(OpenTraceWithRootPathCommand, {
             isVisible: () => false,
--- a/theia-extensions/viewer-prototype/src/browser/trace-viewer/trace-viewer-contribution.ts
+++ b/theia-extensions/viewer-prototype/src/browser/trace-viewer/trace-viewer-contribution.ts
@@ -5,7 +5,15 @@ import URI from '@theia/core/lib/common/uri';
 import { TraceViewerWidget, TraceViewerWidgetOptions } from './trace-viewer';
 import { FileDialogService, OpenFileDialogProps } from '@theia/filesystem/lib/browser';
 import { WorkspaceService } from '@theia/workspace/lib/browser/workspace-service';
-import { OpenTraceCommand, StartServerCommand, StopServerCommand, TraceViewerCommand, KeyboardShortcutsCommand, OpenTraceWithRootPathCommand } from './trace-viewer-commands';
+import {
+  OpenTraceCommand,
+  StartServerCommand,
+  StopServerCommand,
+  TraceViewerCommand,
+  KeyboardShortcutsCommand,
+  OpenTraceWithRootPathCommand,
+  OpenTraceWithPathCommand,
+} from './trace-viewer-commands';
 import { PortBusy, TraceServerConfigService } from '../../common/trace-server-config';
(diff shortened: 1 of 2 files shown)


theia-extensions/viewer-prototype/src/browser/trace-viewer/trace-viewer-contribution.ts
```diff
@@ -178,6 +186,10 @@ export class TraceViewerContribution extends WidgetOpenHandler<TraceViewerWidget
         registry.registerCommand(OpenTraceCommand, {
             execute: () => this.launchTraceServer()
         });
+        registry.registerCommand(OpenTraceWithPathCommand, {
```
bhufmann, line 189: So, this command is executed by an external trigger and the path is known, right? There is no need to open the directory dialog. That's the main difference to OpenWithRootPathCommand  below, which opens the directory dialog with a given location.

theia-extensions/viewer-prototype/src/browser/trace-viewer/trace-viewer-contribution.ts
```diff
@@ -178,6 +186,10 @@ export class TraceViewerContribution extends WidgetOpenHandler<TraceViewerWidget
...
+            isVisible: () => false,
+            execute: (path: string, options: TraceViewerWidgetOpenerOptions) => path && this.open(new URI(path), options),
```
bhufmann, line 191: Do you need the TraceViewerWidgetOpenerOptions? The options has the traceUUID which is used when opening a trace that is already on the server and the traceUUID is known. That is triggered from the `Opened Traces` view. If the option is passed and the traceUUID is set then the path is ignored.

theia-extensions/viewer-prototype/src/browser/trace-viewer/trace-viewer-contribution.ts
```diff
@@ -178,6 +186,10 @@ export class TraceViewerContribution extends WidgetOpenHandler<TraceViewerWidget
...
+        });
```
bhufmann, line 192: Please note that calling this.open(...) might also trigger launching the trace server if the server is not running (health check failure). This might be something that you don't want in your application and we might need to improve that implementation that fits the your application and the current solution with the default trace server.

bhufmann (commented):
@jonahgraham I have 2 questions. For the second one you might need to update the PR depending on the answer.

theia-extensions/viewer-prototype/src/browser/trace-viewer/trace-viewer-contribution.ts
jonahgraham: Yes, that is correct.

theia-extensions/viewer-prototype/src/browser/trace-viewer/trace-viewer-contribution.ts
jonahgraham: I don't need it - I included it as I tried to follow the existing guidance. Basically I though that the plan was to keep extending the options as additional flags were needed to control the open.

But as that can be added later once there are additional options, I will update the patch without this part.

theia-extensions/viewer-prototype/src/browser/trace-viewer/trace-viewer-contribution.ts
jonahgraham: For now I do want that same flow, I think there are some discussions here to be had, but for now what is there is fine.

jonahgraham:
> @jonahgraham I have 2 questions. For the second one you might need to update the PR depending on the answer.

Thanks for the review. I have pushed the change removing the unneeded options.

bhufmann (approved):
Looks good to me. Thanks.

jonahgraham:
Thanks @bhufmann for the quick turnaround. Does this mean it can be merged? I have merge permission on GitHub's permission model now that you have approved, but I don't want to overstep.

MatthewKhouzam (approved):
Looks great! thanks!

PR #898, opened 2022-12-22
Add tooltip when hovering over trace in opened traces view

### LINKED ISSUE #897: Traces with same name can't be distinguished in open traces view
The list of traces in a trace set (a.k.a experiment) can't be distinguished if they have the same name. Also, it's not possible to know where the traces are located in the file system. 

image

The trace path should be shown as hover tooltip (for example).

bhufmann:
Fixes #897

### DIFF
--- a/packages/react-components/src/trace-explorer/trace-explorer-opened-traces-widget.tsx
+++ b/packages/react-components/src/trace-explorer/trace-explorer-opened-traces-widget.tsx
@@ -210,7 +210,7 @@ export class ReactOpenTracesWidget extends React.Component<ReactOpenTracesWidget
         return (
             <div className='trace-element-path-container'>
                 {tracePaths.map(trace => (
-                    <div className='trace-element-path child-element' id={trace.UUID} key={trace.UUID}>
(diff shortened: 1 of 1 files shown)


hoangphamEclipse:
Thank you for the PR. I tried the code and the tooltip only worked when I hover over the line that starts with `>` (the second line). I think we should make it work for the whole div (the whole blue rectangle in the screenshot below) - except for the `Remove trace` icon. This is more user-friendly because intuitively not all users will know where to hover exactly. What do you think?

  image

bhufmann:
> Thank you for the PR. I tried the code and the tooltip only worked when I hover over the line that starts with `>` (the second line). I think we should make it work for the whole div (the whole blue rectangle in the screenshot below) - except for the `Remove trace` icon. This is more user-friendly because intuitively not all users will know where to hover exactly. What do you think?
> 
> image

There might be multiple trace under the trace set and the hover text should be per trace.

hoangphamEclipse:
True, I forgot about that scenario.

hoangphamEclipse (approved):
LGTM

PR #909, opened 2023-01-16
Add ADR about time graph filtering TSP updates

hriday-panchasara:

### DIFF
--- a/doc/adr/0009-time-graph-search-functionality.md
+++ b/doc/adr/0009-time-graph-search-functionality.md
@@ -1,0 +1,166 @@
+# 9. Tsp updates for search functionality in time-graph views
+
+Date: 2023-02-03
+
+## Status
+
+Accepted
+
+## Context
+
+We want to support search and filter expressions for time graph views in Trace Extension. In Trace Compass, time graph views allow users to pull up a search bar that is used for searching/filtering of events. A detailed description of searching/filtering of time graph views in Trace Compass can be found [here][trace-compass-user-guide].
+
+This ADR is intended to discuss the changes needed in the [Trace Server Protocol (TSP)][tsp-api] specification and review the UI mockups.
+
+Currently, the trace-server-protocol documentation for [getStates][tsp-states-documentation] does not involve any query parameters for searching or filtering states. The `/states` endpoint takes in the parameters `requested_timerange` and `requested_items`. The following is an example:
+```
+{
+  "parameters": {
+    "requested_timerange": {
+      "start": 111111111,
+      "end": 222222222,
+      "nbTimes": 1920
+    },
+    "requested_items": [
+      1,
+      2
+    ]
+  }
+}
+```
+- The array `requested_timerange` is the explicit array of requested sample times. The array `requested_items` is the list of entryId's being requested.
+
+But it is possible to augment the request body by adding `regex_map_filters` and `full_search`. These parameters are currently accepted by the time graph data provider - and they can be used to support search/filter expressions. Here's an example where we can use these parameters to search for states with the label "openat":
+```
+{
+  "parameters": {
+    "requested_timerange": {
+      "start": 111111111,
+      "end": 222222222,
+      "nbTimes": 1920
+    },
+    "requested_items": [
+      1,
+      2
+    ],
+    "regex_map_filters": {1: ["openat"]},
+    "full_search": true
+  }
+}
+```
+-   `regex_map_filters` is an array with requests for searching. The  key  of  this  map  can  be \"1\" (means  DIMMED) or \"4\" (means  EXCLUDED) and  the  value  is  an  array  of  the  desired  search  query (e.g. {\"1\": [\"openat\", \"duration>10ms\"]})". The time event filtering uses its own **domain-specific language (DSL)**, which supports various operators and filter expressions. The definition of regexes is outside the scope of this ADR.
+-  The `full_search` key is an optional parameter that enables the full_seach (deep search) or not. 
+   - If true then the data provider is expected to perform a full_search of elements that occur between the usual elements found at the requested times (e.g. in gaps).  If any element matches the specified regex filter, at least one such  matching element per gap should be returned. It is not expected to return  all matching elements found in a single gap, for performance reasons.  
+   - On the client side when we try to perform a graph search we could first do a normal non-full_search query for all the visible rows, in order to quickly show something to the user. This would be followed by a full_search in the background. Because this process can be very slow, we could do it one row at a time and update the graph incrementally as each row's fine-grained result comes back. There might be better ways to combine non-full_search with full_search, therefore the sequence of operations to get the best user experience should be in another ADR. 
+
+Here's an example Sequence Diagram for the search request above:
+
+```mermaid
+sequenceDiagram
+    actor User
+    participant timegraph as timegraph-output-component
+    participant server as Trace Server
+    User->>timegraph: enter search filter regex ("openat")
+    timegraph->>server: POST /experiments/{expUUID}/outputs/timegraph/{outputId}/states
+    Note over timegraph,server: requested_timerange=[current_timerange]
+    Note over timegraph,server: requested_items=[selected_entry_Ids]
+    Note over timegraph,server: regex_map_filters={"1": ["openat"]}
+    Note over timegraph,server: full_search=true
+    server->>timegraph: 200: TimelineChart.TimeGraphModel
+    timegraph->>timegraph: TimeGraphChartProviders.stateStyleProvider updates state style
+    timegraph->>timegraph: Update timegraph
+```
+
+I've attached a link to a prototype in trace-extension that makes use of the aforementioned parameters to  search time events in time graph views:
+- [https://github.com/eclipse-cdt-cloud/theia-trace-extension/pull/886][pull-886]
+- This prototype adds a search-box to the timegraph-output-component to search for events. The filter is applied and unsuccessful time events are dimmed.
+- At the moment its only looking at sampled search, we will add full-search functionality in the near future.
+- The search-box only takes in a single search string - we will add the ability to apply multiple filters through future patches.
+- This might not be the final UI, but it gives the idea on how to achieve it.
+ 
+Even through the current TSP can support search functionality, the use of `regex_map_filters` does not seem to be very intuitive, as a client developer would not be sure of what to pass and what the expected response should be. A suggestion is to rename `regex_map_filters` to `filter_query_parameters`. `filter_query_parameters` takes a `filter_expressions_map`, which contains the search expressions as shown in `regex_map_filters`. 
+
+There is also some confusion with the `full_search` flag, and a suggestion is to change it to `strategy`. `strategy` would accept 2 possible values `sampled` and `deep`. The `strategy` flag is an optional parameter within `filter_query_parameters`, and if `strategy` is omitted then `sampled` search would be the default value.
+The following example does the same query given above but uses `filter_query_parameters` and `strategy` instead:
+```
+{
+  "parameters": {
+    "requested_timerange": {
+      "start": 111111111,
+      "end": 222222222,
+      "nbTimes": 1920
+    },
+    "requested_items": [
+      1,
+      2
+    ],
+    "filter_query_parameters": {
+	  "strategy": "deep",
+	  "filter_expressions_map": {    
+        "1": ["openat"],	   
+	  }
+    },
+  }
+}
+```
+Here are the following patches for implementing time graph filter capability through `filter_query_parameters`:
+-   Update query parameters for `/states` in Trace Compass Incubator: [https://git.eclipse.org/r/c/tracecompass.incubator/org.eclipse.tracecompass.incubator/+/196610][filter-query-parameters-pull-states]
+-   Updates to swagger documentation: [https://git.eclipse.org/r/c/tracecompass.incubator/org.eclipse.tracecompass.incubator/+/196592][filter-query-parameters-pull-swagger]
+-   Updates to the TSP documentation: [https://github.com/eclipse-cdt-cloud/trace-server-protocol/pull/84][filter-query-parameters-pull-tsp]
+
+In the previous examples we noticed that the array of search expressions takes a key value of "1” (means DIMMED) or “4” (means EXCLUDED). We should also discuss the possibility of explicitly mentioning "DIMMED" or "EXCLUDED".
+
+We also need to take a look at the response for `/states` when we add search/filter expressions in our request. The response body would still be a list of time graph entry states for each row. Every state has a bit-mask called `tags`, which helps indicate whether it matched the search/filter expression or not. For example, if we ask for a 'search' expression the states that don't match it would be dimmed, and the `tags` will have the corresponding bit set:
+```
+{
+   "states": [
+     {
+       "tags": 1,
+       "start": 0,
+       "style": {
+         "parentKey": "string",
+         "values": {
+           "property1": {},
+           "property2": {}
+         }
+       },
+       "label": "string",
+       "end": 0
+     }
+   ]
+}
+```
+- When `tags` is set to 1 it is interpreted by the client as setting a dimmed opacity to the state. Similarly, when it is set to 4 it would be considered as being excluded. The reason for this bit mask is to minimize the amount of data transported per state. As previously mentioned, we should consider having `tags` correspond explicitly to either "DIMMED" or "EXCLUDED", although this might cause a slight effect on the performance.
+
+### Future Enhancements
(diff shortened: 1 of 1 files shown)


doc/adr/0009-time-graph-search-functionality.md
```diff
@@ -0,0 +1,133 @@
...
+    "regex_map_filters": {1: ["openat"]},
+    "full_search": true
+  }
+}
+```
+-   “regex_map_filters” is an array with requests for searching. The  key  of  this  map  can  be \"1\" (means  DIMMED) or \"4\" (means  EXCLUDED) and  the  value  is  an  array  of  the  desired  search  query (e.g. {\"1\": [\"openat\", \"duration>10ms\"]})"
+-  The “full_search” key is an optional parameter that enables the full_seach (deep search) or not. 
+   - If true then the data provider is expected to perform a full search of elements that occur between the usual elements found at the requested times (e.g. in gaps).  If any element matches the specified regex filter, at least one such  matching element per gap should be returned. It is not expected to return  all matching elements found in a single gap, for performance reasons.  
+   - In Trace Compass when we perform a graph search, we first do a normal non-full-search query for all the visible rows, in order to quickly show something to the user. Then we perform a full-search in the background. Because this process can be very slow, we do it one row at a time and update the graph incrementally as each row's fine-grained result comes back.
```
bhufmann, line 54: You're mentioning Trace Compass. Please explain it from the theia-trace-extension / TSP point of view.

doc/adr/0009-time-graph-search-functionality.md
```diff
@@ -0,0 +1,133 @@
...
+      2
+    ],
+    "regex_map_filters": {1: ["openat"]},
+    "full_search": true
+  }
+}
+```
+-   “regex_map_filters” is an array with requests for searching. The  key  of  this  map  can  be \"1\" (means  DIMMED) or \"4\" (means  EXCLUDED) and  the  value  is  an  array  of  the  desired  search  query (e.g. {\"1\": [\"openat\", \"duration>10ms\"]})"
+-  The “full_search” key is an optional parameter that enables the full_seach (deep search) or not. 
+   - If true then the data provider is expected to perform a full search of elements that occur between the usual elements found at the requested times (e.g. in gaps).  If any element matches the specified regex filter, at least one such  matching element per gap should be returned. It is not expected to return  all matching elements found in a single gap, for performance reasons.
```
bhufmann, line 53: Adjust the description based on the new name.

doc/adr/0009-time-graph-search-functionality.md
```diff
@@ -0,0 +1,133 @@
...
+   - In Trace Compass when we perform a graph search, we first do a normal non-full-search query for all the visible rows, in order to quickly show something to the user. Then we perform a full-search in the background. Because this process can be very slow, we do it one row at a time and update the graph incrementally as each row's fine-grained result comes back.
+
+
+I've attached a link to a prototype in trace-extension that makes use of the aforementioned parameters to  search time events in time graph views:
+- [https://github.com/eclipse-cdt-cloud/theia-trace-extension/pull/886](https://github.com/eclipse-cdt-cloud/theia-trace-extension/pull/886)
+- This prototype adds a search-box to the timegraph-output-component to search for events. The filter is applied and unsuccessful time events are dimmed.
```
bhufmann, line 59: This might not be the final UI, but it gives the idea how to achieve it.

doc/adr/0009-time-graph-search-functionality.md
```diff
@@ -0,0 +1,133 @@
...
+- At the moment its only looking at sampled search, we will add full-search functionality in the near future.
```
bhufmann, line 60: change full-search

doc/adr/0009-time-graph-search-functionality.md
```diff
@@ -0,0 +1,133 @@
...
+- The search-box only takes in a single search string - we will add the ability to apply multiple filters through future patches.
+ 
+Even through the current TSP can support search functionality, the use of "regex_map_filters" does not seem to be very intuitive, as a client developer would not be sure of what to pass and what the expected response should be.
+
+A suggestion is to rename "regex_map_filters" to "filter_query_parameters". "filter_query_parameters" takes a "filter_expressions_map", which contains the search expressions as shown in "regex_map_filters". It would also include an optional "is_full_search" boolean. The following example does the same query given above but uses "filter_query_parameters" instead:
```
bhufmann, line 65: `strategy` instead of `is_full_search`.

doc/adr/0009-time-graph-search-functionality.md
```diff
@@ -0,0 +1,133 @@
...
+           "property2": {}
+         }
+       },
+       "label": "string",
+       "end": 0
+     }
+   ]
+}
+```
+- When "tags" is set to 1 it is interpreted by the client as setting a dimmed opacity to the state. Similarly, when it is set to 4 it would be considered as being excluded. As previously mentioned, it could be a better alernative to have "tags" correspond explicitly to either "DIMMED" or "EXCLUDED".
```
bhufmann, line 114: The reason for a bit mask is to minimize the amount of data transported per state. In the front-end you can apply a style for each state.

doc/adr/0009-time-graph-search-functionality.md
```diff
@@ -0,0 +1,133 @@
...
+
+
```
bhufmann, line 116: Can add a comment about the filter DSL? Mabye under future enhancements. Basically, the first implementation will pass the string to the back-end and it will apply it. There is no feedback if the syntax was correct. 

However, there is a possibility to create LSP server (Language Server Protocol) in the back-end and front-end will use it to do syntax validation. Trace Compass incubator already has such a server that we can integrate and have an example implementation.

bhufmann (changes requested):

doc/adr/0009-time-graph-search-functionality.md
```diff
@@ -0,0 +1,133 @@
...
+## Status
+
+Proposed
+
+## Context
+
+We want to support search and filter expressions for time graph views in Trace Extension. In Trace Compass, time graph views allow users to pull up a search bar that is used for searching/filtering of events. A detailed description of searching/filtering of time graph views in Trace Compass can be found [here](https://archive.eclipse.org/tracecompass/doc/stable/org.eclipse.tracecompass.doc.user/Trace-Compass-Main-Features.html#Filtering_Time_Events_in_Time_Graph_Views).
+
+This ADR is intended to discuss the changes needed in the [Trace Server Protocol (TSP)](https://github.com/eclipse-cdt-cloud/trace-server-protocol/blob/10cc9ba5419656315333cdac4fb8e392b1c752e7/API-proposed.yaml#L471) specification and review the UI mockups.
```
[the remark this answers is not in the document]
hriday-panchasara: Done

doc/adr/0009-time-graph-search-functionality.md
```diff
@@ -0,0 +1,133 @@
...
+
+Currently, the trace-server-protocol documentation for [getStates](https://eclipse-cdt-cloud.github.io/trace-server-protocol/#tag/TimeGraph/operation/getStates) does not involve any query parameters for searching or filtering states. The '/states' endpoint takes in the parameters 'requested_timerange' and 'requested_items'. The following is an example:
```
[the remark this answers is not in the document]
hriday-panchasara: Done

doc/adr/0009-time-graph-search-functionality.md
hriday-panchasara: Done

doc/adr/0009-time-graph-search-functionality.md
hriday-panchasara: Done

doc/adr/0009-time-graph-search-functionality.md
hriday-panchasara: Done

doc/adr/0009-time-graph-search-functionality.md
hriday-panchasara: Done

doc/adr/0009-time-graph-search-functionality.md
hriday-panchasara: Done

hriday-panchasara:
Added a sequence diagram to show the order of http operations to make the filter work. This diagram is pretty basic, I was not sure if I needed to go in depth.

doc/adr/0009-time-graph-search-functionality.md
```diff
@@ -0,0 +1,154 @@
...
+}
+```
+- When `tags` is set to 1 it is interpreted by the client as setting a dimmed opacity to the state. Similarly, when it is set to 4 it would be considered as being excluded. The reason for this bit mask is to minimize the amount of data transported per state. As previously mentioned, we should consider having `tags` correspond explicitly to either "DIMMED" or "EXCLUDED", although this might cause a slight effect on the performance.
+
+### Future Enhancements
+The initial implementation on the client will pass a single filter query(string) to the back-end and apply the changes. There is no feedback if the syntax is correct. However, there is a possibility to create **LSP (Language Server Protocol)**  in the back-end so that the front-end can use it to do syntax validation.
+
+## Decision
+
+TBD
```
bhufmann, line 140: Here you should propose which solution to implement and the way to get there. Maybe describe increments if applicable.

doc/adr/0009-time-graph-search-functionality.md
```diff
@@ -0,0 +1,154 @@
...
+
+## Consequences
+
+### Easier to do
+
+Enhance feature capabilities of time graph views.
```
bhufmann, line 146: Maybe describe in short the main benefit for the end-user.

bhufmann (changes requested):
Small additional comments from my side. It would be good that other committers would talke part of the review before I approve it.

doc/adr/0009-time-graph-search-functionality.md
```diff
@@ -0,0 +1,154 @@
...
+      "nbTimes": 1920
+    },
+    "requested_items": [
+      1,
+      2
+    ]
+  }
+}
+```
+- The array `requested_times` is the explicit array of requested sample times. The array `requested_items` is the list of entryId's being requested.
```
Rodrigoplp-work, line 31: Would it be "- The array `requested_timerange`" instead of `requested_times`?

doc/adr/0009-time-graph-search-functionality.md
```diff
@@ -0,0 +1,154 @@
...
+    Note over timegraph,server: requested_items=[selected_entry_Ids]
+    Note over timegraph,server: regex_map_filters={"1": ["openat"]}
+    Note over timegraph,server: full_search=true
+    server->>timegraph: 200: TimelineChart.TimeGraphModel
+    timegraph->>timegraph: TimeGraphChartProviders.stateStyleProvider updates state style
+    timegraph->>timegraph: Update timegraph
+```
+
+I've attached a link to a prototype in trace-extension that makes use of the aforementioned parameters to  search time events in time graph views:
+- [https://github.com/eclipse-cdt-cloud/theia-trace-extension/pull/886](https://github.com/eclipse-cdt-cloud/theia-trace-extension/pull/886)
```
Rodrigoplp-work, line 75: This format works, but we tend to prefer listing links at the end of the document, below the last line of text (see ADR 0002-GraphQL for an example).

So here would be:
`- [https://github.com/eclipse-cdt-cloud/theia-trace-extension/pull/886][pull-886]`

And at the end of the ADR:
`[pull-886]: https://github.com/eclipse-cdt-cloud/theia-trace-extension/pull/886`

doc/adr/0009-time-graph-search-functionality.md
```diff
@@ -0,0 +1,154 @@
...
+        "1": ["openat"],	   
+	  }
+    },
+  }
+}
+```
+Here are the following patches for implementing time graph filter capability through `filter_query_parameters`:
+-   Update query parameters for `/states` in Trace Compass Incubator: [https://git.eclipse.org/r/c/tracecompass.incubator/org.eclipse.tracecompass.incubator/+/196610](https://git.eclipse.org/r/c/tracecompass.incubator/org.eclipse.tracecompass.incubator/+/196610)
+-   Updates to swagger documentation: [https://git.eclipse.org/r/c/tracecompass.incubator/org.eclipse.tracecompass.incubator/+/196592](https://git.eclipse.org/r/c/tracecompass.incubator/org.eclipse.tracecompass.incubator/+/196592)
+-   Updates to the TSP documentation: [https://github.com/eclipse-cdt-cloud/trace-server-protocol/pull/84](https://github.com/eclipse-cdt-cloud/trace-server-protocol/pull/84)
```
Rodrigoplp-work, line 109: Move links to the bottom of the ADR.

Rodrigoplp-work (changes requested):
Excellent contribution! I have just a few minor requests intended to keep the format of the ADR a bit more in line with the previous ones.

bhufmann (approved):
LGTM. Thanks for the proposal. Looking forward seeing that in action.

doc/adr/0009-time-graph-search-functionality.md
```diff
@@ -0,0 +1,166 @@
+# 9. Tsp updates for search functionality in time-graph views
+
+Date: 2023-02-03
+
+## Status
+
+Proposed
```
bhufmann, line 7: @hriday-panchasara  We should change it to Accepted.

Rodrigoplp-work (dismissed):

bhufmann (approved):
Renewed approval! Thanks.

doc/adr/0009-time-graph-search-functionality.md
```diff
@@ -0,0 +1,166 @@
...
+    "requested_timerange": {
+      "start": 111111111,
+      "end": 222222222,
+      "nbTimes": 1920
+    },
+    "requested_items": [
+      1,
+      2
+    ],
+    "filter_query_parameters": {
```
frallax, line 97: Does anyone think that this proposal should also consider / compare to the possibility of using graphQL to make such "filtering queries"? I am not sure if [adr 002] was meant to look into this or not, but I think that at least a few words regarding if graphQL would be adapt to do similar "filter queries" or not could be added

[adr 002]: https://github.com/eclipse-cdt-cloud/theia-trace-extension/blob/master/doc/adr/0002-GraphQL.md

doc/adr/0009-time-graph-search-functionality.md
```diff
@@ -0,0 +1,166 @@
...
+}
+```
+- When `tags` is set to 1 it is interpreted by the client as setting a dimmed opacity to the state. Similarly, when it is set to 4 it would be considered as being excluded. The reason for this bit mask is to minimize the amount of data transported per state. As previously mentioned, we should consider having `tags` correspond explicitly to either "DIMMED" or "EXCLUDED", although this might cause a slight effect on the performance.
+
+### Future Enhancements
+The initial implementation on the client will pass a single filter query(string) to the back-end and apply the changes. There is no feedback if the syntax is correct. However, there is a possibility to create **LSP (Language Server Protocol)**  in the back-end so that the front-end can use it to do syntax validation.
+
+## Decision
+
+We should first decide on the TSP updates - focusing on the suggestion to use `filter_query_parameters` and `strategy` to add search/filter expressions to our `/states` query.
```
frallax, line 140: is there an action point / issue (to link here) to follow up on this activity? Just to make sure that this "time-graph filtering feature" will see the light after the ADR is merged, and does not get forgotten / paused

MatthewKhouzam (approved):
I like this. But I want this to be a simple +1, not a merge now +1

doc/adr/0009-time-graph-search-functionality.md
bhufmann: No, I don't think GraphQL should be considered for "filtering queries". ADR 002 investigated GraphQL as replacement of the current TSP implementation.

doc/adr/0009-time-graph-search-functionality.md
bhufmann: After re-reading the statement, I realized that it needs to be re-phrased. The decision is to go with this suggestion and NOT to start a new discussion on how the TSP should look like.

bhufmann: @hriday-panchasara could you please re-phrase that sentence? The decision is to use `filter_query_parameters`, `strategy` to add search/filter expressions to the `/states` query that will return `tags` bit mask.

bhufmann (changes requested):
Sorry for the late request, please update the decision section. This ADR is meant to decide on the API  and the decision section should reflect that.

Rodrigoplp-work (approved):

PR #911, opened 2023-01-18
Time range data widget adr

williamsyang-work:
ADR for Time Range Data Widget:

See for the document: https://github.com/williamsyang-work/theia-trace-extension/blob/time-range-data-widget-adr/doc/adr/0008-TimeRangeData-widget.md


doc/adr/0008-TimeRangeData-widget.md
```diff
@@ -0,0 +1,53 @@
...
+
+If the selected time is invalid, the user will be notified onSubmit with a red asterisk and warning message at the top.  Submit and Cancel buttons will remain visible.
+
+![current](0008/time-range-data-widget-3.png)
+
+If the selected time is valid, the selection range will be set and the widget will return to it's initial state.
+
+#### UX Miscellaneous Behaviors
+* The view range will not change when the selection range is changed.  This includes panning and zooming.
+* The time will be displayed in absolute time as nanoseconds, since that is the format users timestamp data will be.
```
MatthewKhouzam, line 39: can we have a push to view range, and 'follow selection'?

doc/adr/0008-TimeRangeData-widget.md
```diff
@@ -0,0 +1,53 @@
...
+
+#### UX Flow
+
+A "Time Range Data" widget will be added to the left panel.
+
+![current](0008/widget-sidebar.png)
+
+The widget will display: View Start, View End, Selection Start, Selection End, and an Edit button.
+
+![current](0008/time-range-data-widget-1.png)
```
MatthewKhouzam, line 25: can show when the selection is modified? Maybe that will enable the submit button? Also submit on enter?

doc/adr/0008-TimeRangeData-widget.md
```diff
@@ -0,0 +1,53 @@
...
+
+### Proposed UX Design
+
+#### UX Flow
+
+A "Time Range Data" widget will be added to the left panel.
+
+![current](0008/widget-sidebar.png)
+
+The widget will display: View Start, View End, Selection Start, Selection End, and an Edit button.
```
MatthewKhouzam, line 23: Do we need the edit button?

hoangphamEclipse: Same opinion here, from the perspective of a user, I would prefer to do/navigate less to perform a task.

doc/adr/0008-TimeRangeData-widget.md
hoangphamEclipse: Maybe we can have a dialog popup for the user to confirm, something like `Follow selection`, or even better, a icon button to quickly move to the selection range - this might be useful to help users to navigate quickly back to the selection range in case they moved elsewhere, or did not follow the selection initially.

doc/adr/0008-TimeRangeData-widget.md
williamsyang-work: I think having the editable numbers live in a permanent input box would fix this.

doc/adr/0008-TimeRangeData-widget.md
williamsyang-work: IMO pop-ups should be avoided, as they can get annoying after seeing them numerous times.

Moving the view range based on selection range has been an ongoing discussion.  When it comes to view-range changing based on selection-range in the theia-trace-extension, the behavior is mixed. An application wide solution can be implemented in the future.  For example: a toggleable setting for "follow selection", or a button that centers the selection in the view range.

I think the topic needs a separate discussion and should not bottleneck the implementation of the widget.  Not moving the view range is the general behavior of the application, we can change it in the widget after a consensus is reached.

doc/adr/0008-TimeRangeData-widget.md
williamsyang-work: > Do we need the edit button?



> Same opinion here, from the perspective of a user, I would prefer to do/navigate less to perform a task.


I can fix up an updated wireframe for this.  Input box will be visible for all situations, and simply clicking on it begins the editing process.

doc/adr/0008-TimeRangeData-widget.md
hoangphamEclipse: This sounds good, I think an application wide feature for this will benefit the user more. Let's keep this topic for further discussion. We should backlog this somewhere if we have not done this already.

doc/adr/0008-TimeRangeData-widget.md
hoangphamEclipse: > I can fix up an updated wireframe for this. Input box will be visible for all situations, and simply clicking on it begins the editing process.

Sounds good.

hoangphamEclipse:
Thank you @williamsyang-work for the update. It looks good for me, but can you squash the 3 commits together before we merge?

hoangphamEclipse (approved):
LGTM.

MatthewKhouzam (approved):
lgtm

doc/adr/0008-TimeRangeData-widget.md
```diff
@@ -0,0 +1,54 @@
+# 8. Time Range Data Widget
+
+Date: 2023-1-10
+
+## Status
+
+New
```
bhufmann, line 7: Should the status be `Accepted` before merging it? However, the decision section needs to state the decision for that.

marco-miller: Right. Above, `New` should have read `Proposed`, based [on this](https://cognitect.com/blog/2011/11/15/documenting-architecture-decisions). Beside the possible `Accepted` (with documented decision indeed), one may leave the ADR as `Proposed`, for a decision to be finished later on. Could be mergeable saying so then, with a TBD decision.

doc/adr/0008-TimeRangeData-widget.md
Rodrigoplp-work: Agreed. The last thing to do would be to remove the "and an Edit button" from the sentence above.

Rodrigoplp-work (changes requested):

doc/adr/0008-TimeRangeData-widget.md
williamsyang-work: Status has been updated to "Accepted".

hoangphamEclipse (approved):
I checked the doc and the comments are addressed.

bhufmann (approved):
LGTM. Thanks!

PR #913, opened 2023-01-19
Reduce timeline-chart cache by reducing worldRangeFactor to 0.25

### LINKED ISSUE #912: Reduce time graph cache size (world range factor)
Currently, the `timegraph-output-comonent` fetches states for 3x the visible view range, so that left and right to the view range data is cached. This enables for efficient window drag to the left and right. However, for large window ranges and for traces with a lot of states within the view range it leads to longer remote fetch operations due to the data size, time to serialization and transport the data over the network. The drawing time is also longer. 

To reduce the size of data I propose to reduce the cached data to 1/4 of the view range on both ends. With this only 1.5 of the current view range is fetched.

bhufmann:
With this change only 0.25 times (instead of 1.0) of the view range is cached on the before and after of the window range. This reduces the amount of data to be fetched and managed down. This will improve the performance of the timeline-chart when, for example, zooming.

Horizontal panning inside the world range is still available due to the cached data.

Fixes #912

### DIFF
--- a/packages/react-components/src/components/trace-context-component.tsx
+++ b/packages/react-components/src/components/trace-context-component.tsx
@@ -188,6 +188,7 @@ export class TraceContextComponent extends React.Component<TraceContextProps, Tr
             const nanos = zeroPad(theNumber % BigInt(1000));
             return seconds + '.' + millis + ' ' + micros + ' ' + nanos;
         };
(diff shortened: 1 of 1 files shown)


MatthewKhouzam (approved):
Works!

PR #915, opened 2023-01-24
 part 3 of speedyTSP.

MatthewKhouzam:
Wish I didn't have to do this part


MatthewKhouzam:
This showed no benefit after testing.

TL;DR the time is spend serializing/deserializing numbers but it takes about as much time to serialize/deserialize a long number like 100000000 as a shorter number like 100.

PR #919, opened 2023-01-26
Markers sync

williamsyang-work:
Feature Additions:

1.  Ability to select states / rows in the markers chart.
2.  Ability to select rows from any tree by clicking on the row name from tree.
3.  Ability to navigate markers using "<" and ">".

These changes require an uplift of the timeline-chart before being merged into master.  See: https://github.com/eclipse-cdt-cloud/timeline-chart/pull/216.


bhufmann:
@williamsyang-work I think this one is not needed anymore and you can close it.

PR #921, opened 2023-01-31
Events Table search filter persists when scrolling horizontally

### LINKED ISSUE #841: Events Table search filter text gets lost when scrolling horizontally
Events Table search filter text gets lost when scrolling horizontally in the table

table-search-text-scroll-bug

ankusht-work:
When scrolling horizontally in the ag-grid table the SearchFilterRenderer components are unmounted when out of view and then created again when scrolled back into view. This patch adds the filterModel as a prop to the SearchFilterRenderer in order to set a default value when it is created again.

Fixes #841

Signed-off-by: Ankush-Tyagi <ankush.tyagi@ericsson.com>


bhufmann (approved):
Looks good to me. Fixes the bug. Thanks for the contribution.

hriday-panchasara (approved):
LGTM, thanks!

PR #937, opened 2023-02-23
Add null check to item properties

ngondalia:
Changes made:
- Remove the properties from the tooltip object that have null values

This will provide the ability to remove keys from item properties by passing in null values.

Signed-off-by: Neel Gondalia ngondalia@blackberry.com


bhufmann (approved):
Looks good to me. @Rodrigoplp-work could you please give a 2nd review. Thanks.

packages/react-components/src/components/timegraph-output-component.tsx
```diff
@@ -469,14 +469,21 @@ export class TimegraphOutputComponent extends AbstractTreeOutputComponent<Timegr
...
                 'Row': element.row.model.name,
                 ...tooltip
             };
+            Object.keys(fullTooltip).forEach(key => {
+                // eslint-disable-next-line no-null/no-null
+                if (fullTooltip[key as keyof typeof fullTooltip] === null) {
+                    delete fullTooltip[key as keyof typeof fullTooltip];
+                }
+            });
+            return fullTooltip;
```
bhufmann, line 486: Should we do it also for the annotations and arrows below?

bhufmann (changes requested):
Sorry, forgot to add on comment before.

PR #938, opened 2023-02-28
Export DataTree to csv through FileDialogService

hriday-panchasara:
- Using Theia Backend to write files instead
- Users can choose file name/path in both browser & electron
- Show success/error message after

### DIFF
--- a/packages/base/src/signals/signal-manager.ts
+++ b/packages/base/src/signals/signal-manager.ts
@@ -32,6 +32,7 @@ export declare interface SignalManager {
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     fireUnPinView(output: OutputDescriptor, payload?: any): void;
     fireOverviewOutputSelectedSignal(payload: { traceId: string, outputDescriptor: OutputDescriptor}): void;
+    fireSaveDatatreeAsCsv(payload: {traceId: string, data: string}): void;
 }
 
 export const Signals = {
@@ -60,7 +61,8 @@ export const Signals = {
     PIN_VIEW: 'view pinned',
     UNPIN_VIEW: 'view unpinned',
     OPEN_OVERVIEW_OUTPUT: 'open overview output',
-    OVERVIEW_OUTPUT_SELECTED: 'overview output selected'
+    OVERVIEW_OUTPUT_SELECTED: 'overview output selected',
+    SAVE_DATATREE_CSV: 'save datatree as csv'
 };
 
 export class SignalManager extends EventEmitter implements SignalManager {
@@ -140,6 +142,9 @@ export class SignalManager extends EventEmitter implements SignalManager {
     }
     fireOverviewOutputSelectedSignal(payload: { traceId: string, outputDescriptor: OutputDescriptor}): void {
         this.emit(Signals.OVERVIEW_OUTPUT_SELECTED, payload);
+    }
+    fireSaveDatatreeAsCsv(payload: {traceId: string, data: string}): void {
+        this.emit(Signals.SAVE_DATATREE_CSV, payload);
     }
 }
 
--- a/packages/react-components/src/components/datatree-output-component.tsx
+++ b/packages/react-components/src/components/datatree-output-component.tsx
@@ -9,6 +9,7 @@ import { getAllExpandedNodeIds } from './utils/filter-tree/utils';
 import { TreeNode } from './utils/filter-tree/tree-node';
 import ColumnHeader from './utils/filter-tree/column-header';
 import debounce from 'lodash.debounce';
+import { signalManager } from 'traceviewer-base/lib/signals/signal-manager';
 
 type DataTreeOutputProps = AbstractOutputProps & {
 };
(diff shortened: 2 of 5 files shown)


hoangphamEclipse:
Thank you @hriday-panchasara for your review. I ran the code and found a few small issues:

1) When I have multiple traces opened and I tried to save the data tree view to a CSV file, there are multiple dialogs popped up (1 for each opened trace). I think you need to use the trace ID to determine which trace is in the current active tab. Please see the video below for more details. You see that right after I saved the CSV for the `apt` trace, another dialog popped up to ask me to save the CSV for the `64-bit` trace.

https://user-images.githubusercontent.com/92533264/224063658-2df801ac-85cd-4653-8ea0-73baa3f7eec5.mp4

2) I tried to save both files, and both files have the same content from `apt`. I think this issue will be resolved once you solved [1].

3) The content of the file contains some weird unit ` Âµs`. Please see the attached saved CSV file below. FYI, I am on Windows, maybe it is an OS encoding issue?
[apt.csv](https://github.com/eclipse-cdt-cloud/theia-trace-extension/files/10932914/apt.csv)

hriday-panchasara:
> Thank you @hriday-panchasara for your review. I ran the code and found a few small issues:
> 
> 1. When I have multiple traces opened and I tried to save the data tree view to a CSV file, there are multiple dialogs popped up (1 for each opened trace). I think you need to use the trace ID to determine which trace is in the current active tab. Please see the video below for more details. You see that right after I saved the CSV for the `apt` trace, another dialog popped up to ask me to save the CSV for the `64-bit` trace.
> 
> 2. I tried to save both files, and both files have the same content from `apt`. I think this issue will be resolved once you solved [1].
> 3. The content of the file contains some weird unit ` Âµs`. Please see the attached saved CSV file below. FYI, I am on Windows, maybe it is an OS encoding issue?
>    [apt.csv](https://github.com/eclipse-cdt-cloud/theia-trace-extension/files/10932914/apt.csv)

Could you show me what your csv looks like for #3? I am using Numbers on Mac to open the attached csv and it looks identical to the Latency Statistics Table for this trace:
Screen Shot 2023-03-10 at 7 38 01 PM

hoangphamEclipse:
> Could you show me what your csv looks like for 3? I am using Numbers on Mac to open the attached csv and it looks identical to the Latency Statistics Table for this trace:

Here's a screenshot of the CSV file opened in Excel. As you can see, some of the units are `Âµs`:

image

However, today I tried opening it with NotePad++ and it seems fine: 

image

hriday-panchasara:
> 3\. µs

This issue is most likely due to a character encoding problem between Numbers and Excel. The character 'µ' is a non-ASCII character and its encoding can be handled differently by different software applications.

Numbers and Excel may be using different character encodings to interpret the text in the CSV file, causing the 'µ' character to be displayed differently in Excel than in Numbers. The 'Â' character that appears before the 'µ' character in Excel is likely a result of a character encoding error, where the 'µ' character is interpreted as two separate characters in the ISO-8859-1 encoding used by Excel. Numbers and LibreOffice use UTF-8 charset which is why this issue was noticed in those applications.

We can add the option to export statistics to .xls in the future, but for now here's how text/csv should be imported in Excel to ensure correct encoding and have header rows imported:

[Screen Recording 2023-03-14 at 4.19.54 PM.webm](https://user-images.githubusercontent.com/92893187/225139619-1f309143-cda8-44c0-9f8c-2267d5faf21f.webm)

hriday-panchasara:
> > Could you show me what your csv looks like for 3? I am using Numbers on Mac to open the attached csv and it looks identical to the Latency Statistics Table for this trace:
> 
> Here's a screenshot of the CSV file opened in Excel. As you can see, some of the units are `Âµs`:
> 
> image
> 
> However, today I tried opening it with NotePad++ and it seems fine:
> 
> image

This issue is most likely due to a character encoding problem between Numbers and Excel. The character 'µ' is a non-ASCII character and its encoding can be handled differently by different software applications.

Numbers and Excel may be using different character encodings to interpret the text in the CSV file, causing the 'µ' character to be displayed differently in Excel than in Numbers. The 'Â' character that appears before the 'µ' character in Excel is likely a result of a character encoding error, where the 'µ' character is interpreted as two separate characters in the ISO-8859-1 encoding used by Excel. Numbers and LibreOffice use UTF-8 charset which is why this issue was noticed in those applications.

We can add the option to export statistics to .xls in the future, but for now here's how text/csv should be imported in Excel to ensure correct encoding and have header rows imported:

[Screen Recording 2023-03-14 at 4.19.54 PM.webm](https://user-images.githubusercontent.com/92893187/225139619-1f309143-cda8-44c0-9f8c-2267d5faf21f.webm)

hoangphamEclipse (approved):
Thank you for the detailed answers. If the issue regarding the unit is due to the fact that I did not import the file properly into Excel, then that is not a problem with the code. I re-tested the feature and it looks good to me.

hoangphamEclipse:
> This issue is most likely due to a character encoding problem between Numbers and Excel. The character 'µ' is a non-ASCII character and its encoding can be handled differently by different software applications.
> 
> Numbers and Excel may be using different character encodings to interpret the text in the CSV file, causing the 'µ' character to be displayed differently in Excel than in Numbers. The 'Â' character that appears before the 'µ' character in Excel is likely a result of a character encoding error, where the 'µ' character is interpreted as two separate characters in the ISO-8859-1 encoding used by Excel. Numbers and LibreOffice use UTF-8 charset which is why this issue was noticed in those applications.
> 
> We can add the option to export statistics to .xls in the future,

@MatthewKhouzam should we backlog or document this somewhere? I don't know how useful it is. I think you can make the decision.

PR #943, opened 2023-03-16
Time Range Data widget

williamsyang-work:
Adds the Time Range Data Widget to the sidebar as specified in [ADR #8: Time Range Data Widget](https://github.com/eclipse-cdt-cloud/theia-trace-extension/blob/master/doc/adr/0008-TimeRangeData-widget.md)

The unit controller is linked to the widget by signals.  When a new active tab is loaded or changed, the trace-viewer Theia component dispatches the new active unit controller via signal-manager.  This signal is picked up by the time-range-data-widget react component.  Because of this, the unit controller is now a public value in the trace-context-component.

https://user-images.githubusercontent.com/98342456/225728390-da931895-9ce1-48f3-a88e-73d9011d1893.mp4

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,10 +87,11 @@ You can also run two scripts to watch for changes and rebuild automatically:
 
 ## Try the trace extension
 
-This repository contains an example trace-viewer application that includes the trace extension. It has two versions:
+This repository contains an example trace-viewer application that includes the trace extension. It has three versions:
 
 * *browser*: a "browser" application, accessed with a web browser
 * *electron*: a native desktop application
+* *docker*: a Docker-izable "browser" application, accessed with a web browser
 
 You can find those example applications under `examples/`.
 
--- a/examples/docker/Dockerfile
+++ b/examples/docker/Dockerfile
@@ -1,0 +1,18 @@
+FROM node:16-alpine
+
+RUN mkdir /home/app
+WORKDIR /home/app
+
+RUN apk add --no-cache bash git python3 make pkgconfig libsecret-dev g++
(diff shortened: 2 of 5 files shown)


packages/base/src/signals/signal-manager.ts
```diff
@@ -141,6 +143,9 @@ export class SignalManager extends EventEmitter implements SignalManager {
     fireOverviewOutputSelectedSignal(payload: { traceId: string, outputDescriptor: OutputDescriptor}): void {
         this.emit(Signals.OVERVIEW_OUTPUT_SELECTED, payload);
     }
+    fireNewActiveUnitController(unitController: TimeGraphUnitController): void {
+        this.emit(Signals.NEW_ACTIVE_UNIT_CONTROLLER, unitController);
```
bhufmann, line 147: passing the unitcontroller instance on to the time range widget will work in this application. However, it will not work when integrating it in the vscode-trace-extension, because the time range widget and the trace viewer are running in webviews where each webview can't share objects. So, you will have to implement this feature a bit differently and send the selection start/end and view range start/end  in new signal.

Besides, you will have to manage when experiments are switched. Not sure if you can use the existing experiment selected signal for that.

bhufmann (changes requested):
Thanks for this useful feature! I have an initial comment that needs to be addressed before continuing with the review.

williamsyang-work:
Something happened with merge conflicts regarding the exporting CSV signals.  Going to make a new branch and re-open a pull request.  I'll add the reason for refactoring in the description.

PR #946, opened 2023-03-20
Make this traceviewer-base/src import consistent

marco-miller:
By using traceviewer-base/lib instead, consistently with such imports.


bhufmann (approved):
Good catch! Thanks

PR #948, opened 2023-03-21
Rename csv export signal

hriday-panchasara:
Make csv export signal more generic so that it can be used by other outputs in the future


bhufmann (approved):
Looks good to me. Thanks!

PR #952, opened 2023-03-29
Added Export csv support for XY components

ngondalia:
Changes made:

* Component-based drop-down menu implementation that takes showing condition, click function with argument
* Added support for exporting table data as csv for xy-output-component

This will provide export CSV option for XY graphs as well

Signed-off-by: Neel Gondalia ngondalia@blackberry.com

### DIFF
--- a/packages/react-components/src/components/abstract-output-component.tsx
+++ b/packages/react-components/src/components/abstract-output-component.tsx
@@ -11,6 +11,7 @@ import { TooltipComponent } from './tooltip-component';
 import { TooltipXYComponent } from './tooltip-xy-component';
 import { ResponseStatus } from 'tsp-typescript-client/lib/models/response/responses';
 import { signalManager } from 'traceviewer-base/lib/signals/signal-manager';
+import { DropDownComponent, OptionState } from './drop-down-component';
 
 export interface AbstractOutputProps {
     tspClient: TspClient;
@@ -49,8 +50,7 @@ export interface AbstractOutputProps {
 export interface AbstractOutputState {
     outputStatus: string;
     styleModel?: OutputStyleModel;
-    optionsDropdownOpen?: boolean;
-    additionalOptions?: boolean;
+    dropDownOpen?: boolean;
 }
 
 export abstract class AbstractOutputComponent<P extends AbstractOutputProps, S extends AbstractOutputState> extends React.Component<P, S> {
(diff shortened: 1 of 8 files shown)


bhufmann (changes requested):
Thanks for the contribution. I started with testing the feature. I found an issue to looking into. The pin menu item is now always available even after pinning a view. I think it's ok, pinning an unpinned view is replacing the previous pinned view. However, when the previous pinned view becomes unpinned he height is not set properly (see attached gif).

pin-unpin-issue1

Exporting the table in xy-charts works fine. I thought it would export the XY data as well. I should have read the commit message more in detail. That makes me think, if the menu item should be clearer because uses probably think the same way. WDYT?

After your updates I'll continue with the review.

AlexDoanBB:
@bhufmann I didn't know there was a change in pin view behavior, I was only changing the menu, I tried to keep everything else working the same way. I'll look into that.

Regarding exporting the xy data, I planned to make a separate button for that in a separate PR. I'll change the menu item label to "Export table to csv...", would that be okay?

bhufmann:
> @bhufmann I didn't know there was a change in pin view behavior, I was only changing the menu, I tried to keep everything else working the same way. I'll look into that.
> 
No Problem. That's why there is a review :-)

> Regarding exporting the xy data, I planned to make a separate button for that in a separate PR. I'll change the menu item label to "Export table to csv...", would that be okay?
Sounds good to me.

bhufmann (approved):
LGTM. Thanks for this contribution.

hoangphamEclipse (commented):
Thank you for the contribution. I ran some tests and I have a few questions:

1) The dropdown menu is opened even after I had save the CSV. This is not the case when I pin a view, which cause the drop down menu to close immediately. The latter case should be the expected behavior.

2) In some case we don't have data for XY charts. For example, if you open the `apt` trace in the [tracevizlab 103 example](https://github.com/tuxology/tracevizlab), the memory usage view does not show any data. Should we still allow users to export CSV in such cases? I tried exporting the CSV and it worked, but I got a CSV file that only contains a single word `Name`. Maybe it is a future consideration, for an improvement patch that is not within the scope of this PR. What do you think @bhufmann?

bhufmann:
> Thank you for the contribution. I ran some tests and I have a few questions:
> 
>     1. The dropdown menu is opened even after I had save the CSV. This is not the case when I pin a view, which cause the drop down menu to close immediately. The latter case should be the expected behavior. What do you guys think?
> 
I tried it and I can reproduce it. The menu should be hidden after selecting menu item.

>     2. In some case we don't have data for XY charts. For example, if you open the `apt` trace in the [tracevizlab 103 example](https://github.com/tuxology/tracevizlab), the memory usage view does not show any data. Should we still allow users to export CSV in such cases? I tried exporting the CSV and it worked, but I got a CSV file that only contains a single word `Name`. Maybe it is a future consideration, for an improvement patch that is not within the scope of this PR. What do you think @bhufmann?
I think it's ok to show the menu item. If a user chooses to export an empty table, then the user will get a basically empty CSV file.

bhufmann (changes requested):
Sorry, for the late request. We noticed that the dropdown menu doesn't close after selecting the export item. Could you please fix this?

hoangphamEclipse (approved):
LGTM

bhufmann (approved):
Looks good to me. Thanks!

PR #956, opened 2023-04-13
[lerna] option verifyAccess to false

bhufmann:
This will permit publishing to npm using an automation token.

See: https://github.com/lerna/lerna/issues/2788#issuecomment-776726711


marco-miller (approved):
Based on https://github.com/eclipse-cdt-cloud/timeline-chart/pull/250.

PR #958, opened 2023-04-17
Created a widget for placeholder view in react-components

hriday-panchasara:
- Move the placeholder widget to the react-components npm package
- This would allow it to be used by both theia and vscode extension


bhufmann (approved):
Looks good to me. Thanks!

PR #968, opened 2023-04-24
Add Prettier to script to help enforce code style

### LINKED ISSUE #861: Add Prettier as a script to help enforcing code style
Prettier allows the definition of code style for the project.

Install Prettier, define the rules and setup a script in `./package.json` so that it can be used by everyone, regardless of the IDE / code editor of preference.

Also, once the script is ready, it can be added to the build workflow to automatically check the code on every commit.

hoangphamEclipse:
This commit adds Prettier to the trace extension. Prettier allows the extension to define rules for code style and formatting. This helps keep the format of the code consistent and allow quick formatting of new code.

Fix #861.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -63,3 +63,5 @@ jobs:
     - run: yarn --frozen-lockfile --ignore-scripts
 
     - run: yarn lint
+
+      - run: yarn format
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,6 @@
+{
+  "endOfLine": "lf",
+  "semi": false,
+  "singleQuote": true,
+  "trailingComma": "es5"
+}
--- a/package.json
+++ b/package.json
@@ -21,7 +21,8 @@
     "lint": "lerna run lint",
     "test": "lerna run test --",
     "publish:latest": "lerna publish --registry=https://registry.npmjs.org/ --exact --no-git-tag-version --no-push",
-    "publish:next": "lerna publish --registry=https://registry.npmjs.org/ --exact --canary minor --preid=next.$(date -u '+%Y%m%d%H%M%S').$(git rev-parse --short HEAD) --dist-tag=next --no-git-tag-version --no-push --yes"
+    "publish:next": "lerna publish --registry=https://registry.npmjs.org/ --exact --canary minor --preid=next.$(date -u '+%Y%m%d%H%M%S').$(git rev-parse --short HEAD) --dist-tag=next --no-git-tag-version --no-push --yes",
+    "format": "prettier --write */**/*.{ts,tsx}"
   },
   "keywords": [
     "theia-extension",
@@ -48,6 +49,7 @@
     "@theia/cli": "1.34.2",
     "jsonc-parser": "^3.0.0",
     "lerna": "^4.0.0",
+    "prettier": "2.8.8",
     "typescript": "4.5.5"
   },
   "workspaces": [
--- a/configs/base.eslintrc.json
+++ b/configs/base.eslintrc.json
@@ -9,8 +9,12 @@
   },
   "plugins": [
     "@typescript-eslint",
+    "prettier",
     "import",
     "no-null"
+  ],
+  "extends": [
+    "prettier"
   ],
   "env": {
     "browser": true,
--- a/yarn.lock
+++ b/yarn.lock
@@ -11941,6 +11941,11 @@ prepend-http@^2.0.0:
   resolved "https://registry.yarnpkg.com/prepend-http/-/prepend-http-2.0.0.tgz#e92434bfa5ea8c19f41cdfd401d741a3c819d897"
   integrity sha1-6SQ0v6XqjBn0HN/UAddBo8gZ2Jc=
 
+prettier@2.8.8:
+  version "2.8.8"
+  resolved "https://registry.yarnpkg.com/prettier/-/prettier-2.8.8.tgz#e8c5d7e98a4305ffe3de2e1fc4aca1a71c28b1da"
+  integrity sha512-tdN8qQGvNjw4CHbY+XXk0JgCXn9QiF21a55rBe5LJAU+kDyC4WQn4+awm2Xfk2lQMk5fKup9XgzTZtGkjBdP9Q==
+
 pretty-format@^27.0.2:
   version "27.5.1"
   resolved "https://registry.yarnpkg.com/pretty-format/-/pretty-format-27.5.1.tgz#2181879fdea51a7a5851fb39d920faa63f01d88e"


williamsyang-work:
@hoangphamEclipse I've been wondering how we can re-format this codebase without losing the commit history and came across this article: [How to introduce a code formatter without messing up Git history](https://medium.com/codex/how-to-introduce-a-code-formatter-without-messing-up-git-history-4a16bd074c10).

Basically, we can create a `.git-blame-ignore-revs` file and it will ignore the huge linting commit.  There is a little more to it than just that.  We'd have to add information to the READ_ME.md file on how users can set up their git settings to utilize this `.git-blame-ignore-revs` file.  But other than that it's pretty straightforward.

Not sure if that's out of the scope of this ticket.  If so, it's something I can work on in post.

.github/workflows/build.yml
```diff
@@ -63,3 +63,5 @@ jobs:
     - run: yarn --frozen-lockfile --ignore-scripts
 
     - run: yarn lint
+
+      - run: yarn format
```
bhufmann, line 67: running prettier here to format the code will update the source code on the build machine. I don't think it will fail the build. Nor will it check-in the changes (Don't create an auto-commit). So, I don't think we should run `yarn format` here.

I'm not sure how you intended to use prettier. Should/Will `yarn lint` fail if it is not according to prettier settings? Please let me know.

.prettierrc
```diff
@@ -0,0 +1,6 @@
+{
+  "endOfLine": "lf",
+  "semi": false,
```
bhufmann, line 3: most of the current code has semicolons at the end. I would set it to true.

.prettierrc
```diff
@@ -0,0 +1,6 @@
...
+  "singleQuote": true,
+  "trailingComma": "es5"
+}
```
bhufmann, line 6: most of our code uses 4 space for tab width. I think we should keep it.

package.json
```diff
@@ -21,7 +21,8 @@
     "lint": "lerna run lint",
```
bhufmann, line 21: For me `yarn lint` fails with the following error:
```
lerna notice cli v4.0.0
lerna info Executing command in 3 packages: "yarn run lint"
traceviewer-base: $ eslint .
traceviewer-base: Oops! Something went wrong! :(
traceviewer-base: ESLint: 7.32.0
traceviewer-base: ESLint couldn't find the config "prettier" to extend from. Please check that the name of the config is correct.
traceviewer-base: The config "prettier" was referenced from the config file in "/home/eedbhu/git/theia-training/theia-trace-extension/configs/base.eslintrc.json".
traceviewer-base: If you still have problems, please stop by https://eslint.org/chat/help to chat with the team.
traceviewer-base: error Command failed with exit code 2.
traceviewer-base: info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
lerna ERR! yarn run lint exited 2 in 'traceviewer-base'
error Command failed with exit code 2.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

package.json
```diff
@@ -21,7 +21,8 @@
...
     "test": "lerna run test --",
     "publish:latest": "lerna publish --registry=https://registry.npmjs.org/ --exact --no-git-tag-version --no-push",
-    "publish:next": "lerna publish --registry=https://registry.npmjs.org/ --exact --canary minor --preid=next.$(date -u '+%Y%m%d%H%M%S').$(git rev-parse --short HEAD) --dist-tag=next --no-git-tag-version --no-push --yes"
+    "publish:next": "lerna publish --registry=https://registry.npmjs.org/ --exact --canary minor --preid=next.$(date -u '+%Y%m%d%H%M%S').$(git rev-parse --short HEAD) --dist-tag=next --no-git-tag-version --no-push --yes",
+    "format": "prettier --write */**/*.{ts,tsx}"
```
bhufmann, line 25: This will also format files under node_modules directory and lib directory. Only the source code of this repo  should be formatted, not dependencies nor build outputs.

bhufmann (changes requested):
The github actions for `build_and_test` and `lint` are not executed with this change. Please make sure that they are executed (My guess is something is wrong with the change in `build.yml`).

.prettierrc
```diff
@@ -0,0 +1,6 @@
+{
+  "endOfLine": "lf",
+  "semi": false,
+  "singleQuote": true,
+  "trailingComma": "es5"
```
bhufmann, line 5: I don't like the trailing commas being added, e.g. after `end` and `}` in this example:
absoluteRange: {
               start,
               end,
  },
```

```

.github/workflows/build.yml
bhufmann: > > I'm not sure how you intended to use prettier.
> 
> I am taking over this task from Estelle so I had to do a little bit digging. I checked #861 and in the issue Rodrigo says that Prettier should be run to check the format of the code in CI. From what I understand, we were supposed to:
> 
> [1] Enable Prettier in the project so everyone can use it. We can set up a [pre-commit hook](https://prettier.io/docs/en/precommit.html) so that Prettier is ran every time the developer makes a commit. With this we ensure that the code is well-formatted before it get pushed to GitHub.
> 
Having prettier installed for users to run sounds good to me. I'm not sure to run the prettier script automatically with a pre-commit hook. Files get staged and it might get confusing. User could choose to add a pre-commit hook script in their repo if they wish, but that's on them. Note, if we realize later that the pre-commit hook should be there, we can add it later on.

> [2] Enable Prettier in CI to ensure that the code is well-formatted before merging. So we need to run `prettier --check` in CI as documented [here](https://prettier.io/docs/en/install.html).
> 
Updating the CI to add the check and fail the build makes sense. User's can then update the patch to run prettier.

> If this workflow makes sense, let me know and I can proceed with the changes.
> 
> > Should/Will `yarn lint` fail if it is not according to prettier settings?
> 
> According to Prettier documentation, there might be cases where Prettier might have a conflict with the styling config of linters. Most of the time, Prettier should not be causing any conflicts.
> 
> I think currently using ESLint as the linter for this repo (please correct me if I am wrong). Worst case, there is a version of Prettier that plays well with ESLint called [`eslint-config-prettier`](https://prettier.io/docs/en/integrating-with-linters.html). This should prevent any unexpected conflict between the two plugins. However, this means that we will be running Prettier as ESLint.
Thanks for the info.

.prettierrc
bhufmann: Ok, thanks for the link to the related discussion. I don't want to have a long discussion on comma vs semi-colon, since it might never end :-). I'm ok to have semi-colon instead of commas.

bhufmann (changes requested):
I added some replies to open questions.

.prettierrc
hoangphamEclipse: I will update this settings, by default it is 2.

.prettierrc
hoangphamEclipse: Setting `trailingComma` to `none` should fix it.

hoangphamEclipse:
> Basically, we can create a `.git-blame-ignore-revs` file and it will ignore the huge linting commit.

@williamsyang-work Sorry for the late reply, but I think this is a good idea. It is within scope of this commit. I think I will do that. Btw, this PR is closed in favor of a better one #973.

PR #972, opened 2023-05-05
Update parameter for onViewRangeChanged handler

hoangphamEclipse:
The [performance improvement patch for the timeline chart](https://github.com/eclipse-cdt-cloud/timeline-chart/pull/246) changes the parameters provided to handlers of the onViewRangeChanged event on the unit controller. This commit updates the Theia Trace Extension to ensure that it works with the performance improvement patch.

This PR is to be merged after the performance improvement patch.


bhufmann (changes requested):
Code looks good. Please update the yarn.lock to get the latest timeline-chart and this PR needs also a rebase. There are some merge conflicts.

bhufmann (approved):
Looks good to me. Performance of resize looks much better with the updated timeline-chart. Thanks!

PR #977, opened 2023-05-18
Abstract CSS style for React components

hoangphamEclipse:
Since the Trace Extension was migrated to VSCode, the CSS classes must be abstracted and mapped to the correct Theia and VSCode CSS variables. This commit replaces the prefix --theia of CSS variables with --trace-viewer in the react-components package. For example, the variable --theia-ui-padding becomes --trace-viewer-ui-padding. The packages that use these components then have to import a CSS file that maps these CSS variables to the CSS variables of their local environment, or they need to define the CSS for those components.

This is a prerequisite to resolve [this issue](https://github.com/eclipse-cdt-cloud/vscode-trace-extension/issues/67).

PR #980, opened 2023-05-25
Rename --trace-viewer-foreground where applicable

hoangphamEclipse:
Previously, the --theia-menu-foreground CSS variable was replaced by  the variable --trace-viewer-foreground in an attempt to abstract the CSS definition. This is incorrect and the variable should have been mapped to --trace-viewer-menu-foreground instead. This commit corrects this error (where --theia-menu-foreground was previously --theia-menu-foreground).

Relates to #977.


bhufmann (approved):
LGTM, thanks!

PR #985, opened 2023-06-12
Fix toggle columns missing in title bar menu

### LINKED ISSUE #981: Events Table title bar menu doesn't have the toggle column menu anymore
Events Table title bar menu used to show the toggle column menu to show and hide columns. This menu entry is not shown anymore. Only `Pin view` menu entry is there. This is a regression of commit 35df458 that introduced a separate `DropDownComponent` to handle the title bar menu.

image

The methods `renderToggleColumnsTable`, `closeOptionsDropDown` and `showAdditionalOptions` of the `table-output-component.tsx` are not hooked up anymore.

hoangphamEclipse:
The toggle columns menu item in the title bar menu was removed due to the refactor done in commit 35df458. This commit uses the addOption method that was introduced in commit 35df458 to bring back the toggle columns options.

Fixes #981.


bhufmann (approved):
LGTM. Thanks for the fix!

PR #988, opened 2023-06-13
Added indicator for empty column search

dleclairbb:
Added a check to see if any row elements match search. The color of search text updates to red if no results, and is unassigned (reverts to theme) if there are none.

This makes it more clear that a search does not have any results.

Signed-off-by: Dylan leclair dleclair@blackberry.com

### DIFF
--- a/packages/react-components/src/components/table-renderer-components.tsx
+++ b/packages/react-components/src/components/table-renderer-components.tsx
@@ -11,8 +11,8 @@ type CellRendererProps = ICellRendererParams & {
 
 type SearchFilterRendererProps = IFloatingFilterParams & {
     onFilterChange: (colName: string, filterValue: string) => void;
-    onclickNext: () => void;
-    onclickPrevious: () => void;
+    onclickNext: () => Promise<boolean>;
+    onclickPrevious: () => Promise<boolean>;
     colName: string;
     filterModel: Map<string, string>;
 };
@@ -20,6 +20,8 @@ type SearchFilterRendererProps = IFloatingFilterParams & {
 interface SearchFilterRendererState {
     hasHovered: boolean;
     hasClicked: boolean;
+    isSearchSuccessful: boolean;
+    isSearching: boolean;
 }
 
 export class LoadingRenderer extends React.Component<ICellRendererParams> {
@@ -86,11 +88,13 @@ export class SearchFilterRenderer extends React.Component<SearchFilterRendererPr
         super(props);
         this.state = {
             hasHovered: false,
-            hasClicked: this.props.filterModel.has(this.props.colName)
+            hasClicked: this.props.filterModel.has(this.props.colName),
+            isSearchSuccessful: true,
+            isSearching: false
         };
 
         this.debouncedChangeHandler = debounce((colName, inputVal) => {
-            this.props.onFilterChange(colName, inputVal);
+            this.onFilterChange(colName, inputVal);
         }, 10);
 
         this.onInputBoxChanged = this.onInputBoxChanged.bind(this);
@@ -101,6 +105,8 @@ export class SearchFilterRenderer extends React.Component<SearchFilterRendererPr
         this.onUpClickHandler = this.onUpClickHandler.bind(this);
         this.onCloseClickHandler = this.onCloseClickHandler.bind(this);
         this.onKeyDownEvent = this.onKeyDownEvent.bind(this);
+        this.onFilterChange = this.onFilterChange.bind(this);
+        this.runSearch = this.runSearch.bind(this);
     }
 
     render(): JSX.Element {
@@ -110,7 +116,6 @@ export class SearchFilterRenderer extends React.Component<SearchFilterRendererPr
                 onMouseEnter={this.onMouseEnterHandler}
                 onMouseLeave={this.onMouseLeaveHandler}
                 onClick={this.onClickHandler}
-                title="Enter a regular expression"
             >
                 {!this.state.hasClicked && !this.state.hasHovered && (
                     <FontAwesomeIcon style={{ marginLeft: '10px' }} icon={faSearch} />
@@ -139,26 +144,42 @@ export class SearchFilterRenderer extends React.Component<SearchFilterRendererPr
                             autoFocus={true}
                             onKeyDown={this.onKeyDownEvent}
                             onInput={this.onInputBoxChanged}
-                            style={{ width: '50%', margin: '10px' }}
+                            style={{
+                                width: '50%',
+                                margin: '10px',
+                                color: this.state.isSearchSuccessful ? '' : 'red'
+                            }}
                             defaultValue={this.props.filterModel.get(this.props.colName) ?? ''}
+                            title="Enter a regular expression, then press Enter"
                         />
+                        {this.state.isSearching && (
+                            <FontAwesomeIcon
+                                spin
+                                icon={faSpinner}
+                                style={{ marginRight: '10px', marginTop: '20px' }}
+                                title="Searching..."
+                            />
+                        )}
                         <FontAwesomeIcon
                             className="hoverClass"
                             icon={faTimes}
                             style={{ marginTop: '20px' }}
                             onClick={this.onCloseClickHandler}
+                            title="Clear search"
                         />
                         <FontAwesomeIcon
                             className="hoverClass"
                             icon={faAngleDown}
                             style={{ marginLeft: '10px', marginTop: '20px' }}
                             onClick={this.onDownClickHandler}
+                            title="Find next"
                         />
                         <FontAwesomeIcon
                             className="hoverClass"
                             icon={faAngleUp}
                             style={{ marginLeft: '10px', marginTop: '20px' }}
                             onClick={this.onUpClickHandler}
+                            title="Find previous"
                         />
                     </div>
                 )}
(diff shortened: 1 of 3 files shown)


bhufmann (commented):
Thanks the contribution. 

I have a concern with the  change. The table-output-component is a virtual table where a `cacheBlockSize: 200` is loaded into memory. When entering a search field these 200 events are checked and matching events are highlighted. With your change if no matching events are found within the 200 events, then the search string will be red. However, the trace could have more events and matching events are outside the cache block size. I find it misleading to change it to red because there might be matches in the whole trace. To know if there are any matches potentially up-to the whole trace would have to be checked to know if there are any matches, which I don't we should do. 

Not sure what we can do to improve your proposal to indicate that the user has to press enter to find next which will trigger a back-end call to find the next matching event in the trace from current selection.

WDYT?

dleclairbb:
I'm not sure. There could be another color to indicate that the search has not been successful for the current cache -- I was going to suggest yellow but it would likely become unreadable. Another way to indicate it might be to add a count of the results out of the results searched (ex. 2/200), which the user can compare against the size of the whole trace but this risks cluttering up the table with extra information.

bhufmann:
> I'm not sure. There could be another color to indicate that the search has not been successful for the current cache -- I was going to suggest yellow but it would likely become unreadable. Another way to indicate it might be to add a count of the results out of the results searched (ex. 2/200), which the user can compare against the size of the whole trace but this risks cluttering up the table with extra information.

Not sure for showing the count relative to the cache search, because the end-user is not (and doesn't need to be) aware that it's a virtual table. The end-user might confuse the count with the total number of events or the page counts shown on the bottom right of the table. 

One thing is missing, when a user presses enter and the back-end is queried to find the next matching event, then there is no visual indication that the search is ongoing. That operation might be quite long depending on the search string and size of trace. I think that should be added.

dleclairbb:
> One thing is missing, when a user presses enter and the back-end is queried to find the next matching event, then there is no visual indication that the search is ongoing. That operation might be quite long depending on the search string and size of trace. I think that should be added.

What kind of indication would you suggest for this? Perhaps some kind of animated wheel next to the search bar?

bhufmann:
> > One thing is missing, when a user presses enter and the back-end is queried to find the next matching event, then there is no visual indication that the search is ongoing. That operation might be quite long depending on the search string and size of trace. I think that should be added.
> 
> What kind of indication would you suggest for this? Perhaps some kind of animated wheel next to the search bar?

Yes, animated wheel. That's one way beside the search bar. There might be multiple columns with a search string and the wheel should be on all of them. Each table cells show a wheel when the user moves out of the cache and the back-end is queried to get more data.

Another way could be to show the wheel top of the the title bar string `Events Table` of the `table-output-component` like what the `time-graph-component` does when fetching from the back-end. 

The first one proposal would be consistent with the table cells during query and it's clear that it is related to the search. The second one would be consistent with time-graphs but might be less discoverable than the first one. 

I would be ok with both. Maybe the first one, because seems to be clearer. WDYT?

dleclairbb:
I agree that the first one is clearer, but upon further reflection, I think this issue is actually a few smaller issues put together:

- there is no indication of whether the search for the next result is taking place (which is solved by your suggestions)
- the search appears to be be live, but the user must hit enter (or the next result button) to actually query the entire table which can be confusing (particularly for new users)
- when a search is unsuccessful, there is no clear visual indication of it (especially if a user is unaware of having to hit enter)

To address these issues, I propose that:
- An animated wheel is added to the search columns in the event table when the backend is being queried for the next result (your first suggestion)
- Upon changing a column's search contents, the table will **automatically** jump to focus on the first result found (while this is happening, the animated wheel is displayed)
- Upon changing a column's search contents, if the query to the backend yields no results, the next/previous buttons will be replaced with "No results" (and potentially the search text goes red as shown here?)

Thank you for your feedback -- please let me know what you think.

bhufmann:
> I agree that the first one is clearer, but upon further reflection, I think this issue is actually a few smaller issues put together:
> 
> * there is no indication of whether the search for the next result is taking place (which is solved by your suggestions)
> * the search appears to be be live, but the user must hit enter (or the next result button) to actually query the entire table which can be confusing (particularly for new users)
> * when a search is unsuccessful, there is no clear visual indication of it (especially if a user is unaware of having to hit enter)
> 
> To address these issues, I propose that:
> 
> * An animated wheel is added to the search columns in the event table when the backend is being queried for the next result (your first suggestion)

Great!
> * Upon changing a column's search contents, the table will **automatically** jump to focus on the first result found (while this is happening, the animated wheel is displayed)

The current behaviour is, when the user types, the current cache is searched and matches are highlighted. To select the first (and next) the user has to press ENTER. For previous the user has to press SHIFT+ENTER. 

Your suggestion will potentially trigger a back-end call on every keystroke stroke might not yield the expected behaviour for the user either. There will be multiple back-end searches triggered and it needs the be handled (only keep last). 

So, I like keep using ENTER which it's more deterministic for the user (find NEXT). Maybe instead, there could be hint that the user needs to press ENTER.

> * Upon changing a column's search contents, if the query to the backend yields no results, the next/previous buttons will be replaced with "No results" (and potentially the search text goes red as shown here?)
> 
Yes, it's good to show that no event was found. Showing "No result" and maybe text goes red sounds good to me. 

> Thank you for your feedback -- please let me know what you think.

dleclairbb:
Awesome, I'll start chipping away at that. Do you any preference over what kind of indicator is used to hint to the user about hitting enter? Maybe a hover text or search icon? 

As an aside: What's the best workflow for developing with this extension? So far I've been rebuilding with `yarn` and launching the application with `yarn start:server`, which has been rather slow (it would also be nice to debug somehow).

bhufmann:
> Awesome, I'll start chipping away at that. Do you any preference over what kind of indicator is used to hint to the user about hitting enter? Maybe a hover text or search icon?

Not sure. Maybe hover text or return icon on right of the input box. 

Btw, are you aware that there are buttons on the right of the search input box for next/previous search?. If you don't see them you might need make the column wider. They show up when user clicks into the search box.

image

> 
> As an aside: What's the best workflow for developing with this extension? So far I've been rebuilding with `yarn` and launching the application with `yarn start:server`, which has been rather slow (it would also be nice to debug somehow).

I use Vscodium for development and I have chrome debugger (legacy) and firefox debugger installed. I do the following steps:
```bash
yarn
yarn start:browser
````
Then I use the launch configuration to start the chrome in debug mode (you could use the firefox one as well). I debug using the VsCodium debugger. Alternatively, you could open the chrome manually goto localhost:3000 and use the chrome dev tools for debugging and trouble-shooting.

After changing code I redo these steps. Some people use `yarn watch`, but I'm not familiar with it at all because I don't use it.

**Comment by dleclairbb** [comment]
This should be ready to go now! The build failed, but it was because of an issue fetching a dependency :/ 

The only thing that is *slightly* different from what we discussed is that, after trying it out, I decided not to display "No results" when a backend query yields no results -- It did not fit well into column header, and when I moved it next to the search box it would shrink the input bar itself and felt really bad.

Please let me know what you think!

**Review by bhufmann — CHANGES_REQUESTED** [review]
I like the spinning wheel and I think it improves the UX.

I noticed when the search is ongoing and the user presses the x button to the right to stop the search, then the input box is cleared (which is expected). However, when the user clicks into the input box again, then the spinning wheel comes back and the old text is shown in the input box. So, it looks like there is some reset missing after pressing the x. 

Could you please have a look?

Thanks

**Comment by dleclairbb** [comment]
I was able to reproduce and find a fix for the spinning wheel persisting when clearing the search, but I'm unable to reproduce the input text reappearing. I've tried both successful and unsuccessful searches. Are there any more details you can provide?

**Comment by bhufmann** [comment]
> I was able to reproduce and find a fix for the spinning wheel persisting when clearing the search, but I'm unable to reproduce the input text reappearing. I've tried both successful and unsuccessful searches. Are there any more details you can provide?

I think this only happens if a back-end query is ongoing and hasn't returned. Once the backend call is returned then the search field is cleared when you click again.

Follow the code in `onCloseClickHandler()` of `table-renderer-components.tsx` which calls `this.props.onFilterChange(this.props.colName, ')` which is provided by `table-renderer-compoents.tsx`. The corresponding callback `searchEvents()` has a while loop to wait till the long lasting search is finished (`while(this.gridMatched)`).

Without the ongoing, long-lasting back-end call, the text is cleared correctly.

At the moment, I think we can't really do much until there is way to cancel back-end calls (which doesn't exist). 

So, for this PR I think you just need to remove the spinning wheel.

File: packages/react-components/src/components/table-renderer-components.tsx [hunk 1]
```diff
@@ -139,45 +144,84 @@ export class SearchFilterRenderer extends React.Component<SearchFilterRendererPr
...
+    }
+
+    private runSearch(callback: () => Promise<boolean> | undefined): void {
+        this.setState({ isSearching: true, isSearchSuccessful: true }, () => {
+            this.forceUpdate(() => {
+                const result = callback();
+                if (result === undefined) {
+                    return;
+                }
+                result.then(isFound => {
```
bhufmann, line 202: Should we handle also the reject case, i.e. reset the flags? Rejecting is not implemented, but to be future proof. WDYT?

dleclairbb: I could tack on: 

```ts
                result.catch(() => {
                    this.setState({
                        isSearching: false,
                        isSearchSuccessful: false,
                    });
                });
```

dleclairbb: Should I worry about error messages / the like? I think resetting the flags should be sufficient.

bhufmann: That's what I had in mind. Reset the flags to avoid being in an incorrect state. No error message is ok.

dleclairbb: Just added it in! One of my seniors over here wanted me to leave an error message in since it might be handy down the road. 

Thanks again for your help :)

bhufmann: I agree with your colleague that logging errors is important and we should do it. However, there is no proper logging infrastructure in place at them moment and we avoid using console for that. 

If I approve the PR, there is a good chance that the 2nd reviewer (another committer in project) will ask to remove it. Maybe remove it right away.

dleclairbb: Should be good to go now.

bhufmann (approved):
Looks good to me. Thanks for the contribution. I'll ask for a 2nd review of another committer before merging. Should be done soon.

hoangphamEclipse (approved):
Thank you for the contribution. LGTM!

PR #989, opened 2023-06-15
releng: Define publishing latest semantic versions

marco-miller:
Add a `publish-latest` GitHub workflow similar to the existing `publish-next`.
- Make the former trigger on merging a commit that bumps the package versions depending on the chosen yarn script command.
- Hence, make the latter trigger only on the other kinds of commits.
- Align `publish-next` (with `publish-latest`) also so its previous trivial nits don't show as differences, upon comparing the two files for review purposes or so.

Beside the usual goal of publishing onto the npm Registry, use the `github-tag-action` in `publish-latest`, to push the corresponding tag after that publishing.
- Make the hereby generated commit (message) automatically convey how to name that bumped tag.
- For that purpose, `github-tag-action` uses a message token such as `#major`, `#minor` or `#patch`.
- Here, `publish-latest` expects `GITHUB_TOKEN` to be set, similarly to the already used `NPM_AUTH_TOKEN`.

Add the related yarn scripts and document how to use them (commands) in the `CONTRIBUTING` markdown.
- Document also how the aforementioned workflow is expected to automate the process that includes these commands.
- And, mention how to work-around any potential workflow issue manually; if any until that can be fixed.
- This adds the benefit of documenting how that automation proceeds, step-wise.

Fix `.gitignore`'s `examples`-related lines so they work also upon using the hereby added yarn commands.

Once confirmed, this approach can be ported to [1] and [2] in turn.
- With this approach, introducing a Jenkins job [3] wouldn't be necessary.

[1] https://github.com/eclipse-cdt-cloud/timeline-chart
[2] https://github.com/eclipse-cdt-cloud/tsp-typescript-client
[3] https://ci.eclipse.org/cdt-cloud/

### DIFF
--- a/.github/workflows/publish-latest.yml
+++ b/.github/workflows/publish-latest.yml
@@ -1,0 +1,30 @@
+name: NPM package (latest)
+on:
+  push:
+    branches: [ master ]
+jobs:
+  build:
+    if: github.ref == 'refs/heads/master' && github.event_name == 'push' && github.repository == 'eclipse-cdt-cloud/theia-trace-extension' && "contains(github.event.head_commit.message, '[latest]:')"
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      # For github-tag-action below
+    steps:
+    - uses: actions/checkout@v2
+    # Setup .npmrc file to publish to npm
+    - uses: actions/setup-node@v3
+      with:
+        node-version: '16.x'
+        registry-url: 'https://registry.npmjs.org'
+        # Defaults to the user or organization that owns the workflow file
+        scope: '@theia-ide'
+    - run: yarn
+    - run: yarn publish:latest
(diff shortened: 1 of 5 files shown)


hoangphamEclipse (approved):
Hi Marco, your PR looks good to me. However, it is my first time reviewing a PR that altering the GitHub workflow and I am not sure what should be tested and checked. I can't run the publish command, since that requires a LINUX machine.

@bhufmann maybe you can help suggest what should be verify?

bhufmann (approved):
It looks good to me. I was able to test with a local npm-like repository and it worked. The github action looks good, but the final confirmation we'll get when it's merged and we do an actual release. Thanks for the contribution.

marco-miller:
So we had to revert this commit soon after merge, per #996.

PR #990, opened 2023-06-22
Refactor server communication to JSON-RPC

### LINKED ISSUE #976: Refactor server communication to JSON-RPC
By default the Theia extension and the React components do not communicate with the Trace Compass server via the Theia backend, but rather connect to a separate exposed port (e.g. 8080) via REST. This approach has some downsides. For example in cloud deployments the additional port needs to be exposed as well.

We would like to suggest to refactor the communication in the Theia extension to delegate the communication via Theia's JSON-RPC from the Theia frontend over the Theia backend to the Trace Compass server and back.

Of course the pure React components should stay independent of Theia and allow to connect to arbitrary end points.

Are you interested in this change? If yes, we will gladly contribute an initial implementation.

sgraband:
**Update to newest tsp-typescript-client next version**

To consume the new types introduced with https://github.com/eclipse-cdt-cloud/tsp-typescript-client/pull/83.
Invoke the yarn clean script before running the prepare script.
Make sure all imports are from `lib` instead of `src`.

**Update types of base package and react-components**

Use the newly introduced types of tsp-typescript-client.
The type of TspClient becomes ITspClient.
The constructor for the base implementation becomes HttpTspClient.

**Refactor server communication to JSON-RPC**

TspClient Proxy:
Provide an implementation of the ITspClient called `TheiaRpcTspProxy`.
The implementation forwards all calls to a JSON RPC proxy called `TheiaClientProxy`.
Responses are then mapped to a `TspClientResponse`.
On the backend the `TheiaClientProxy` returns a lazyTspClient.
The lazyTspClient returns a `HttpTspClient` (default implementation).

Trace Server url provider:
The implementation of the provider was moved to the backend.
This way the env variable can be read out directly.
For the port preference a proxy mechanism was added.
Changes to the preference are sent via JSON RPC and handled by the url provider.

Trace Server connection status:
The implementation was moved to the backend.
Frontends can register themselves on the backend.
Each registered frontend will be pinged when the status has changed.

Misc:
Use `ITspClient` instead of `TspClient` for types, as `TspClient` is deprecated.

Fixes https://github.com/eclipse-cdt-cloud/theia-trace-extension/issues/976

Contributed on behalf of STMicroelectronics

### DIFF
--- a/packages/base/src/experiment-manager.ts
+++ b/packages/base/src/experiment-manager.ts
@@ -1,5 +1,5 @@
 import { Trace } from 'tsp-typescript-client/lib/models/trace';
-import { TspClient } from 'tsp-typescript-client/lib/protocol/tsp-client';
+import { ITspClient } from 'tsp-typescript-client/lib/protocol/tsp-client';
 import { Query } from 'tsp-typescript-client/lib/models/query/query';
 import { OutputDescriptor } from 'tsp-typescript-client/lib/models/output-descriptor';
 import { Experiment } from 'tsp-typescript-client/lib/models/experiment';
@@ -9,10 +9,10 @@ import { signalManager, Signals } from './signals/signal-manager';
 
 export class ExperimentManager {
     private fOpenExperiments: Map<string, Experiment> = new Map();
-    private fTspClient: TspClient;
+    private fTspClient: ITspClient;
     private fTraceManager: TraceManager;
 
-    constructor(tspClient: TspClient, traceManager: TraceManager) {
+    constructor(tspClient: ITspClient, traceManager: TraceManager) {
         this.fTspClient = tspClient;
         this.fTraceManager = traceManager;
         signalManager().on(Signals.EXPERIMENT_DELETED, (experiment: Experiment) =>
@@ -79,7 +79,10 @@ export class ExperimentManager {
             traceURIs.push(traces[i].UUID);
         }
 
-        const tryCreate = async function (tspClient: TspClient, retry: number): Promise<TspClientResponse<Experiment>> {
+        const tryCreate = async function (
+            tspClient: ITspClient,
+            retry: number
+        ): Promise<TspClientResponse<Experiment>> {
             return tspClient.createExperiment(
                 new Query({
                     name: retry === 0 ? name : name + '(' + retry + ')',
--- a/theia-extensions/viewer-prototype/src/browser/trace-viewer/trace-viewer-contribution.ts
+++ b/theia-extensions/viewer-prototype/src/browser/trace-viewer/trace-viewer-contribution.ts
@@ -21,12 +21,12 @@ import {
 } from './trace-viewer-commands';
 import { PortBusy, TraceServerConfigService } from '../../common/trace-server-config';
 import { TracePreferences, TRACE_PATH, TRACE_ARGS } from '../trace-server-preference';
-import { TspClient } from 'tsp-typescript-client/lib/protocol/tsp-client';
 import { TspClientProvider } from '../tsp-client-provider-impl';
 import { ChartShortcutsDialog } from '../trace-explorer/trace-explorer-sub-widgets/charts-cheatsheet-component';
 import { signalManager } from 'traceviewer-base/lib/signals/signal-manager';
-import { TraceServerConnectionStatusService } from '../trace-server-status';
+import { TraceServerConnectionStatusClientImpl } from '../trace-server-connection-status-client-impl';
 import { FileStat } from '@theia/filesystem/lib/common/files';
+import { ITspClient } from 'tsp-typescript-client';
 
 interface TraceViewerWidgetOpenerOptions extends WidgetOpenerOptions {
     traceUUID: string;
@@ -37,7 +37,7 @@ export class TraceViewerContribution
     extends WidgetOpenHandler<TraceViewerWidget>
     implements CommandContribution, KeybindingContribution
 {
-    private tspClient: TspClient;
+    private tspClient: ITspClient;
 
     constructor(@inject(TspClientProvider) private tspClientProvider: TspClientProvider) {
         super();
@@ -94,7 +94,7 @@ export class TraceViewerContribution
                         progress.report({ message: 'Trace server started.', work: { done: 100, total: 100 } });
                     }
                     progress.cancel();
-                    TraceServerConnectionStatusService.renderStatus(true);
+                    TraceServerConnectionStatusClientImpl.renderStatus(true);
                     signalManager().fireTraceServerStartedSignal();
                     this.openDialog(rootPath);
                 }
@@ -163,7 +163,7 @@ export class TraceViewerContribution
                         } else {
                             progress.report({ message: 'Trace server started.', work: { done: 100, total: 100 } });
                         }
-                        TraceServerConnectionStatusService.renderStatus(true);
+                        TraceServerConnectionStatusClientImpl.renderStatus(true);
                         signalManager().fireTraceServerStartedSignal();
                         return super.open(traceURI, options);
                     }
@@ -230,7 +230,7 @@ export class TraceViewerContribution
                         } else {
                             progress.report({ message: 'Trace server started.', work: { done: 100, total: 100 } });
                         }
-                        TraceServerConnectionStatusService.renderStatus(true);
+                        TraceServerConnectionStatusClientImpl.renderStatus(true);
                         signalManager().fireTraceServerStartedSignal();
                         return;
                     }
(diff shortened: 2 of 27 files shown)


sgraband:
@bhufmann regarding these two points in the description:

> - note that this interface could be moved to tsp-typescript-client
> - then the TspClient could also follow the same interface

The plan would be to make the `TspClient` in the [tsp-typescript-client](https://github.com/theia-ide/tsp-typescript-client) package and interface and also offer a `DefaultTspClient` that implements the new interface, and is the same as the current `TspClient`. This way we can remove the `TspFrontendClient` interface from this PR and just implement the new `TspClient` interface. Then both the `DefaultTspClient` and the client used in the Theia frontend would follow the same interface.

Is there anything that would speak against this rename and redesign from your perspective? Otherwise, i would also open the contribution on  [tsp-typescript-client](https://github.com/theia-ide/tsp-typescript-client) and adjust this PR. In the meantime you could already check out this PR as it would basically only be interface changes and the implementations will not change.

bhufmann:
Thanks for the contribution. I'm a bit stretch with time at the moment for the review and test and I had backlog it. Having said that, one team member will look into the impact on the latency so that we have an idea what the additional communication step adds for the visualization of large(r) data.

packages/base/src/experiment-manager.ts
```diff
@@ -6,13 +6,14 @@ import { Experiment } from 'tsp-typescript-client/lib/models/experiment';
 import { TraceManager } from './trace-manager';
 import { TspClientResponse } from 'tsp-typescript-client/lib/protocol/tsp-client-response';
 import { signalManager, Signals } from './signals/signal-manager';
+import { TspFrontendClient } from './tsp-frontend-client';
 
 export class ExperimentManager {
     private fOpenExperiments: Map<string, Experiment> = new Map();
-    private fTspClient: TspClient;
+    private fTspClient: TspClient | TspFrontendClient;
```
colin-grant-work, line 13: There are a lot of cases where this type option has been added. Would it make sense to define a single interface that both services satisfy and just refer to that?

colin-grant-work: A converse to [a comment](https://github.com/eclipse-cdt-cloud/tsp-typescript-client/pull/83/files#r1270706958) I've just made over there: perhaps the `TspFrontendClient` could be named more precisely to something like `TheiaRpcClient` (still implementing the new `TspClient` interface)?

theia-extensions/viewer-prototype/src/browser/trace-viewer/trace-viewer-contribution.ts
```diff
@@ -257,7 +257,7 @@ export class TraceViewerContribution
                 try {
                     await this.traceServerConfigService.stopTraceServer();
                     this.messageService.info('Trace server terminated successfully.');
-                    TraceServerConnectionStatusService.renderStatus(false);
+                    TraceServerConnectionStatusClientImpl.renderStatus(false);
                 } catch (err) {
                     this.messageService.error('Failed to stop the trace server.');
                 }
```
marco-miller, line 263: I get two compilation errors locally, in this file, if I rebase this PR on current latest master tip. This is the error I get at both lines 97 and 166 (current master version):

```
Cannot find name 'TraceServerConnectionStatusService'.
TraceServerConnectionStatusService.renderStatus(true);
```

marco-miller (changes requested):
Thanks for this promising effort, well appreciated! Some comments for me as well, herein.

packages/base/src/tsp-frontend-client.ts
```diff
@@ -0,0 +1,270 @@
...
+    /**
+     * Fetch XY tooltip
+     * @param expUUID Experiment UUID
+     * @param outputID Output ID
+     * @param xValue X value
+     * @param yValue Optional Y value
+     * @param seriesID Optional series ID
+     * @returns Map of key=name of the property and value=string value associated
+     */
+    fetchXYToolTip(
```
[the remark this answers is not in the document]
sgraband: This is no longer relevant, as we implement the upstream interface.

theia-extensions/viewer-prototype/src/browser/trace-viewer/trace-viewer-contribution.ts
sgraband: Fixed it with the latest changes.

marco-miller:
> @marco-miller I consumed the changes of [eclipse-cdt-cloud/tsp-typescript-client#83](https://github.com/eclipse-cdt-cloud/tsp-typescript-client/pull/83) and rearranged the commits to hopefully make it easier to review. Your comments of your earlier review should also be incorporated as well as the naming discussions. Could you take another look? If you have any suggestions on how to split up the last commit i am happy to incorporate them, but i feel like the changes all are connected somehow so i wasn't sure on how (or if) to split them.

Thanks for this. I recently changed focus at work, so leaving this follow-up and review to our other reviewers listed here. Cheers.

theia-extensions/viewer-prototype/src/node/trace-server-url-provider-impl.ts
```diff
@@ -45,41 +49,36 @@ export class TraceServerUrlProviderImpl implements TraceServerUrlProvider, Front
...
+        this._traceServerUrlTemplate = variable ? this.normalizeUrl(variable) : TRACE_SERVER_DEFAULT_URL;
+        this.updateTraceServerUrl();
+    }
+
+    async onPortPreferenceChanged(
+        newPort: number | undefined,
+        oldValue?: number,
+        preferenceChanged = false
+    ): Promise<void> {
+        this._traceServerPort = newPort;
```
bhufmann, line 70: Changing the port doesn't seem to be propagate properly to the TspClient and messages are not successful.

To reproduce:
- Command: `Start Trace Server`
- Open Settings -> Trace Viewer
- Change Port to 8081 (which will stop the trace server)
- Change "Trace Server: Arguments" to -vmargs -Dtraceserver.port=8081
- Command `Start Trace Server`
- There are error messages: "ERROR Error: Unsuccessful health check: Service Unavailable status: undefined"

Note: the server process is actually started (do "ps -aef  | grep trace | grep server")
Note: heath check is executed after the trace sever is spawned

I did debug a bit and noticed that the http-tsp-client in the backend doesn't use the correct port.

bhufmann (changes requested):
I'll continue with reviewing from Marco. Thanks for the updates. 

I did some testing and the main feature is working well. However, I found some issues during testing. Could you please investigate this? I'll continue with this PR once updated.

theia-extensions/viewer-prototype/src/node/trace-server-connection-status-backend-impl.ts
```diff
@@ -0,0 +1,29 @@
...
+
+    constructor() {
+        const listener = (status: boolean) => {
+            this.clients.forEach(client => client.updateStatus(status));
+        };
+        RestClient.addConnectionStatusListener(listener);
+    }
+
+    setClient(client: TraceServerConnectionStatusClient): void {
+        this.clients.push(client);
```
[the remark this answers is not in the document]
sgraband: I fixed this behavior with the latest commit.

theia-extensions/viewer-prototype/src/node/trace-server-url-provider-impl.ts
sgraband: I will need to dig a little bit more into this to find out why this happens. Just for my understanding:
The Port preference is simply there for the requests to the server, while the params will be used for the startup of the Trace Server (when startet from Theia), right?

And changing the Port should kill all servers running on the old Port?

sgraband: This behavior should now be fixed with the latest commit. I also rebased on master.

bhufmann: Yes, it's fixed.

packages/base/src/experiment-manager.ts
```diff
@@ -6,13 +6,14 @@ import { Experiment } from 'tsp-typescript-client/lib/models/experiment';
...
     private fTraceManager: TraceManager;
 
-    constructor(tspClient: TspClient, traceManager: TraceManager) {
+    constructor(tspClient: TspClient | TspFrontendClient, traceManager: TraceManager) {
```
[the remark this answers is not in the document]
bhufmann: Ack

bhufmann (approved):
It looks good to me. Thank you very much for the contribution!

MatthewKhouzam (approved):
This chain improves the code by making communications simpler. I very much approve it.

Thanks Simon!

PR #991, opened 2023-06-29
Restore State in ReactTimeRangeDataWidget

williamsyang-work:
Adds a public method that sets the activeData and experimentDataMap.  This is used in the vscode-trace-extension repo because we persist state data with the VSCode API and need to restore it externally.


packages/react-components/src/trace-explorer/trace-explorer-time-range-data-widget.tsx
```diff
@@ -32,7 +32,7 @@ export class ReactTimeRangeDataWidget extends React.Component<
     ReactTimeRangeDataWidgetProps,
     ReactTimeRangeDataWidgetState
 > {
-    private experimentDataMap: Map<string, ExperimentTimeRangeData>;
+    public experimentDataMap: Map<string, ExperimentTimeRangeData>;
```
bhufmann, line 35: Should not be public.

packages/react-components/src/trace-explorer/trace-explorer-time-range-data-widget.tsx
```diff
@@ -267,6 +267,19 @@ export class ReactTimeRangeDataWidget extends React.Component<
         }
     };
 
+    /**
+     * Allows public to set active data and experiment data map.
+     * This is used in vscode-trace-extension where we need to manually
```
bhufmann, line 272: we should not mention vscode in the comment since vscode is just on implementation.

packages/react-components/src/trace-explorer/trace-explorer-time-range-data-widget.tsx
```diff
@@ -267,6 +267,19 @@ export class ReactTimeRangeDataWidget extends React.Component<
...
+     * persist and restore data using the VSCode API when component is
+     * not visible.  User inputs are not restored.
+     * @param activeData
+     * @param map
+     */
+    public restoreState = (activeData: ExperimentTimeRangeData, map: Map<string, ExperimentTimeRangeData>) => {
```
bhufmann, line 278: can't you just send a signal to itself when it's time to restore as if the signal came from the viewer panel?

bhufmann (changes requested):
I have some comments. Please note that the build fails due to linting errors. Please fix as well.

packages/react-components/src/trace-explorer/trace-explorer-time-range-data-widget.tsx
williamsyang-work: Can we access this with a getter method?  Or should we remove all direct ref's to this react component from outside?

packages/react-components/src/trace-explorer/trace-explorer-time-range-data-widget.tsx
williamsyang-work: Ok.

packages/react-components/src/trace-explorer/trace-explorer-time-range-data-widget.tsx
williamsyang-work: Yes, this can be accessed by a signal.

packages/react-components/src/trace-explorer/trace-explorer-time-range-data-widget.tsx
williamsyang-work: This has been changed to a getter method.

packages/react-components/src/trace-explorer/trace-explorer-time-range-data-widget.tsx
```diff
@@ -267,6 +267,23 @@ export class ReactTimeRangeDataWidget extends React.Component<
         }
     };
 
+    /**
+     * Allows public setting of active data and experiment data map.
+     * @param activeData
+     * @param map
+     */
+    public restoreState = (activeData: ExperimentTimeRangeData, map: Map<string, ExperimentTimeRangeData>): void => {
```
bhufmann, line 275: maybe give call parameter `map` something more meaningful

williamsyang-work: This has been changed to experimentDataMap

bhufmann (commented):
The linting is failing. Could you please fix it? Thanks.

bhufmann (dismissed):
Code looks good. It works with well with https://github.com/eclipse-cdt-cloud/vscode-trace-extension/pull/152.

bhufmann (changes requested):
Please remove the console.dir

packages/react-components/src/trace-explorer/trace-explorer-time-range-data-widget.tsx
```diff
@@ -267,6 +268,48 @@ export class ReactTimeRangeDataWidget extends React.Component<
...
+        });
+    };
+
+    get persistedState(): {
+        experimentDataMap: Map<string, ExperimentTimeRangeData>;
+        activeData?: ExperimentTimeRangeData;
+    } {
+        return {
+            experimentDataMap: this.experimentDataMap,
+            activeData: this.state.activeData
```
bhufmann, line 305: Can't  you store the activeData as member and don't use the state here. Wouldn't it solve the timing issue?

packages/react-components/src/trace-explorer/trace-explorer-time-range-data-widget.tsx
```diff
@@ -267,6 +268,48 @@ export class ReactTimeRangeDataWidget extends React.Component<
...
+        };
+    }
+
+    get isBusy(): boolean {
```
bhufmann, line 309: I'm not for adding such a public API.

bhufmann (dismissed):
PR needs a rebase. Also, the commits need to be squashed before merging.

MatthewKhouzam:
*time :) first word in the review

packages/react-components/src/trace-explorer/trace-explorer-time-range-data-widget.tsx
```diff
@@ -200,7 +176,8 @@ export class ReactTimeRangeDataWidget extends React.Component<
         v1 = BigInt(v1);
         v2 = BigInt(v2);
 
-        const offset = this.state.activeData?.absoluteRange?.start || BigInt(0);
+        let offset = this.state.activeData?.absoluteRange?.start || BigInt(0);
```
MatthewKhouzam, line 178: can this be const offset = BigInt(this.state.activeData?.absoluteRange?.start || 0);?

packages/react-components/src/trace-explorer/trace-explorer-time-range-data-widget.tsx
```diff
@@ -59,91 +61,65 @@ export class ReactTimeRangeDataWidget extends React.Component<
...
-            this.onAbsoluteRangeUpdate(experiment);
-        }
+    
+    private onExperimentUpdated = (experiment: Experiment): void => {
+        this.onAbsoluteRangeUpdate(experiment);
+        this.renderIfActive();
     };
-
+    
     private onExperimentSelected = (experiment: Experiment | undefined): void => {
```
MatthewKhouzam, line 87: This changes the functionality.

if we don't have an experiement, we will set active to null. Before, it would not. I think this is desirable, but is it deliberate?

packages/react-components/src/trace-explorer/trace-explorer-time-range-data-widget.tsx
```diff
@@ -59,91 +61,65 @@ export class ReactTimeRangeDataWidget extends React.Component<
...
+        this.setActiveExperiment(newActiveData);
     };
+    
+    private onExperimentClosed = (experiment: Experiment | string): void => {
+        this.experimentDataMap.delete(experiment);
+    };
+    
+    private setActiveExperiment = (timeData?: ExperimentTimeRangeData): void => {
+        this.experimentDataMap.setActiveExperiment(timeData);
+        const activeData = timeData ? this.experimentDataMap.get(timeData.UUID) : undefined;
```
MatthewKhouzam, line 103: can we inline activeData? no is fine

packages/react-components/src/trace-explorer/trace-explorer-time-range-data-widget.tsx
williamsyang-work: Yes.  When the experiment is undefined it indicates an experiment has been closed.  This was deliberate.

MatthewKhouzam (approved):

MatthewKhouzam:
I would like to confirm it works with vscode # 152 before merging. But this commit in and of itself is good IMO.

MatthewKhouzam:
Re-requesting a review to unblock.

MatthewKhouzam (dismissed):

MatthewKhouzam (approved):
Please update commit messages

PatrickTasse (approved):

MatthewKhouzam:
Great job!

PR #992, opened 2023-06-29
Refresh open traces view on start during open

bhufmann:
When triggering an open command, a trace server will be started if necessary. With this PR, upon successful start of trace server, the status indication and the opened trace view will be refreshed.


hoangphamEclipse (approved):
LGTM.

PR #999, opened 2023-07-19
Measure performance of TSPClient before and after JSON-RPC

hoangphamEclipse:
This commit adds a study on the impact of the JSON-RPC patch on the Theia Front End - Back End - Trace Server communication speed.

Relates to #990.

### DIFF
--- a/doc/performance/0002-http-vs-rcp-client-performance-analysis.md
+++ b/doc/performance/0002-http-vs-rcp-client-performance-analysis.md
@@ -1,0 +1,404 @@
+# Comparison of the performance of TSPClient before and after JSON-RPC implementation
+
+## Context
+
+By default the Theia extension and the React components do not communicate with the Trace Compass server via the Theia backend, but rather connect to a separate exposed port (e.g. 8080) via REST. This approach has some downsides. For example in cloud deployments the additional port needs to be exposed as well. The full description of the issue can be found in [issue #976](https://github.com/eclipse-cdt-cloud/theia-trace-extension/issues/976).
+
+Visually, the current implementation is:
+
+```text
+Theia FE <--TSP over HTTP--> Trace Server
+```
+
+[Pull request #990](https://github.com/eclipse-cdt-cloud/theia-trace-extension/pull/990) proposes the following solution for issue #976:
+
+```text
+Theia FE <--msgpackr(previously was JSON-RPC)--> Theia BE <--TSP over HTTP--> Trace Server
+```
+
+Instead of communicating with the Trace Server directly from the front-end, the proposed solution will have the front-end (Theia FE) send its HTTP requests to the Theia Back-end (Theia BE). Then, the BE will perform the HTTP request and return the response to the FE.
+
+It is important to note that, previously, Theia used `JSON-RPC` to serialize data for its FE-BE communication. However, recently, it has opted for a more performant standard called `message pack`. Theia now uses `msgpackr` (an implementation of message pack). More information regarding this decision, as well as some benchmarking statistics, can be found [here](https://github.com/eclipse-theia/theia/issues/11159). This change is done since Theia version `1.29.0`. The tests below are done with Theia version `1.37.2`. In the code, the connection handler is still named `JsonRpcConnectionHandler`, although Theia will default to using `msgpackr`. Hence, in this documentation, I use `JSON-RPC` and `msgpackr` interchangeably, to refer to the changes made in PR #990.
+
+The goal of this document is to compare the performance of the Trace Extension before and after PR #990. Since we have an additional link in the communication chain, it is expected that the communication time will either be the same or increased. We would like to know the magnitude of this increased, and how it affects the performance of the Theia Trace Extension in general.
+
+## Test specs
+
+### Setup
+
+```text
+                                ..,
+                    ....,,:;+ccllll   -------------------------------
+      ...,,+:;  cllllllllllllllllll   OS: Windows 10 Pro x86_64
+,cclllllllllll  lllllllllllllllllll   Host: Dell Inc. Latitude 5420
+llllllllllllll  lllllllllllllllllll   Kernel: 10.0.19044
+llllllllllllll  lllllllllllllllllll   Uptime: 4 days, 22 hours, 21 mins
+llllllllllllll  lllllllllllllllllll   Packages: 4 (scoop)
+llllllllllllll  lllllllllllllllllll   Shell: bash 4.4.23
+llllllllllllll  lllllllllllllllllll   Resolution: 1920x1080, 1920x1080
+                                      DE: Aero
+llllllllllllll  lllllllllllllllllll   WM: Explorer
+llllllllllllll  lllllllllllllllllll   WM Theme: Custom
+llllllllllllll  lllllllllllllllllll   CPU: 11th Gen Intel i5-1145G7 (8) @ 1.500GHz
+llllllllllllll  lllllllllllllllllll   GPU: Caption
+llllllllllllll  lllllllllllllllllll   GPU: Intel(R) Iris(R) Xe Graphics
+`'ccllllllllll  lllllllllllllllllll   GPU: Citrix Indirect Display Adapter
+       `' \*::  :ccllllllllllllllll   GPU
+                       ````''*::cll   Memory: 24368MiB / 32145MiB
+                        ``
+```
+
+Theia version: 1.37.2 (have to be at least 1.29.0).
+
+For all logs to appear in the debug console in visual studio code, make sure that the Browser FE and Theia BE is ran in debug mode.
+
+Also, to add logs to the TSPClient, you need to clone the `tsp-typescript-client` repo and link it to the Theia Trace extension.
+
+### Benchmarking criteria
+
+To benchmark the performance of the Trace Extension before and after PR #990, timestamps are logged to the debug console of VSCode. There are 4 timestamps to be logged (in order of execution):
+
+1. The timestamp when we initiate the communication to the TSPClient (e.g when the FE calls the TSPClient).
+2. The timestamp when we reach the TSPClient object. In the case of JSON-RPC, 
+3. The timestamp when HTTP requests are finished
+4. The timestamp when we receive the data back from the TSPClient
+
+Then, the following statistics is calculated:
+
+1. TSPClient processing time (ms): The time took to generate, send and receive response for an HTTP request. This is the time from when the TSPClient is reached until the result is ready to be returned to the FE.
+2. Front-end to TSPClient (ms): The time to reach the TSPClient from the Theia FE. In the current implementation, this value is expected to be negligible since the TSPClient is accessed directly by the FE. However, in the JSON-RPC implementation, some latency is expected since now the TSPClient is on the Theia BE.
+3. TSPClient to Front-end (ms): The time took to return the HTTP response data from the BE to the FE. Same as [2], some latency is expected for the JSON-RPC implementation.
+4. Total duration: The total time took to send an HTTP request and receive response from the TSPClient. In the JSON-RPC implementation, this includes communication time with the Theia BE.
+5. Overhead: The time added by adding the Theia BE to the communication chain.
+
+Then, the median, average, max and min value for each category is calculated to compare the performance pre and post JSON-RPC.
+
+### Test cases
+
+This benchmarking attempt measures the performance of 3 operations, before and after the JSON-RPC patch. Each test case below are run 15 times on the current implementation and on the JSON-RPC patch:
+
+1. tspClient.experimentOutputs(): Get the available outputs for a specific trace.
+
+```text
+Test 1
+Params: experiment ID
+Returns: an array of OutputDescriptors
+
+Steps:
+1. Open the Theia Trace extension in a browser.
+2. Open a trace using the trace explorer on the left. The available views section on the left side should be populated.
+3. Check the debug console and record the numbers.
+```
+
+2. tspClient.fetchXYTree(): Get the filter tree for the left side view of an XY chart.
+
+```text
+Test 2
+Params: experiment ID, output ID, and time range parameters
+Returns: the model of XY trees
+
+Steps:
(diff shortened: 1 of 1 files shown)


doc/performance/0002-http-vs-rcp-client-performance-analysis.md
```diff
@@ -0,0 +1,94 @@
...
+
+### Pre-JSON-RPC-patch
+
+|experimentOutputs() (ms)(*)| fetchXYTree() (ms)| fetchTimeGraphStates() (ms)
+|---|---|---|
+|472.5|90.7|100.8|
+|301|98.4|108.2|
+|465.3|95.7|139.2|
+|865.9|91.6|126.7|
+|905.1|103.6|201.8|
```
marco-miller, line 56: Are these samples all taken after the trace server is pretty much done warming up (caching at least some of the involved data)? I'm asking because, a server that is still busy loading most of the related data off disk is likely to take more latency time serving it back.

doc/performance/0002-http-vs-rcp-client-performance-analysis.md
```diff
@@ -0,0 +1,94 @@
...
+
+On average we have:
+
+|experimentOutputs() (ms)(*)| fetchXYTree() (ms)| fetchTimeGraphStates() (ms)
+|---|---|---|
+|601.6|96.0|135.34|
+
+(*): For the case of fetchTimeGraphStates(), the number varies quite a bit. If we run more test attempts, sometimes we have outliers that are very small (24-35ms). However, it looks like most of the time, the range is 400-600ms.
```
marco-miller, line 64: Could be because of some more cached data being available at such moments in time, within the trace server I mean.

doc/performance/0002-http-vs-rcp-client-performance-analysis.md
```diff
@@ -0,0 +1,94 @@
...
+|160.1|95.7|142.9|
+|116.5|97.9|128.5|
+|159.2|100.5|133.8|
+|175.9|91|123.8|
+
+On average we have:
+
+|experimentOutputs() (ms)(**)| fetchXYTree() (ms)| fetchTimeGraphStates() (ms)
+|---|---|---|
+|223.0|94.3|164.5|
```
marco-miller, line 80: Averages have significance limits, as we know. Now, here, it is surprising that adding a json-rpc leg diminishes the latency overall. Including the aforementioned logging points might help narrow this issue down somewhat, in case their respective locations would be debatable.

doc/performance/0002-http-vs-rcp-client-performance-analysis.md
```diff
@@ -0,0 +1,94 @@
...
+
+(**): Similar to (\*) sometimes we have outliers too (around 26-35ms).
+
+## Conclusion
+
+The results are not consistent for each different test cases. On average:
```
marco-miller, line 86: `each different test case.`

doc/performance/0002-http-vs-rcp-client-performance-analysis.md
```diff
@@ -0,0 +1,94 @@
...
+
+1. For experimentOutputs(), the JSON-RCP update improves the speed of communication.
```
marco-miller, line 88: Surprising and to be debunked, per my previous comment.

doc/performance/0002-http-vs-rcp-client-performance-analysis.md
```diff
@@ -0,0 +1,94 @@
...
+2. For fetchXYTree(), the JSON-RCP update does not make much difference in the speed of communication.
```
marco-miller, line 89: Too narrow to judge only based on this specific case, indeed; would "discard" its instance.

doc/performance/0002-http-vs-rcp-client-performance-analysis.md
```diff
@@ -0,0 +1,94 @@
...
+3. For fetchTimeGraphStates(), the JSON-RCP update takes more time compared to the HTTP direct implementation.
```
marco-miller, line 90: Which might be the expectation, as previously stated. Now these bare extra milliseconds may not suffice in concluding on a significant loss of performance.

doc/performance/0002-http-vs-rcp-client-performance-analysis.md
```diff
@@ -0,0 +1,94 @@
...
+
+One of the possible explanation is the difference in the amount of data that needs to be sent to the back end. This is true for the case of the `fetchXYTree()` and the `fetchTimeGraphStates()` tests. Since they both serialize a good amount of data, the performance either stays the same - in the case of `fetchXYTree()` where we have small amount of data - or becomes worse - in the case of `fetchTimeGraphStates()` where we have a decent amount of data.
```
marco-miller, line 92: `of the possible explanations`

marco-miller, line 92: `small amounts of`

marco-miller, line 92: true in the case of the `fetchXYTree()` and `fetchTimeGraphStates()` tests

marco-miller, line 92: In this paragraph, we use both `good amount `and `small amount` to describe the `fetchXYTree` case; which sounds in opposition. I mean, `good` (meaning rather significant) -versus `small`. It seems like we could drop the `fetchXYTree` case overall, as previously proposed.

doc/performance/0002-http-vs-rcp-client-performance-analysis.md
```diff
@@ -0,0 +1,94 @@
...
+
+However, this does not explain the communication speed improves in the case of `experimentOutputs()`. Thus, more investigation is required to fully understand the impact of JSON-RCP on the Theia Trace Extension.
```
marco-miller, line 94: `explain why the`

doc/performance/0002-http-vs-rcp-client-performance-analysis.md
```diff
@@ -0,0 +1,94 @@
...
+## Conclusion
+
+The results are not consistent for each different test cases. On average:
+
+1. For experimentOutputs(), the JSON-RCP update improves the speed of communication.
+2. For fetchXYTree(), the JSON-RCP update does not make much difference in the speed of communication.
+3. For fetchTimeGraphStates(), the JSON-RCP update takes more time compared to the HTTP direct implementation.
+
+One of the possible explanation is the difference in the amount of data that needs to be sent to the back end. This is true for the case of the `fetchXYTree()` and the `fetchTimeGraphStates()` tests. Since they both serialize a good amount of data, the performance either stays the same - in the case of `fetchXYTree()` where we have small amount of data - or becomes worse - in the case of `fetchTimeGraphStates()` where we have a decent amount of data.
+
```
marco-miller, line 93: A missing test case might be, to measure the impact on a trailing yet tiny payload request, preceded by a heavy one. Meaning, is the trailing (tiny) request starved in any way by the preceding elephant. We could compare the outcome of this test before versus after json-rpc. This was recently proposed by @MatthewKhouzam IIRC.

So I think he mentioned the tiny request being a trace annotations/markers one or so. -Preceded by some truckload data fetch act. The latter shouldn't starve or prevent the former from showing, IIUC. If this is an issue before changing to json-rpc, then measuring with the latter in turn might become irrelevant. The question could otherwise be, is the json-rpc wrapper (towards the back-ended client use) adding latency to the trailing/tiny request?

marco-miller (changes requested):
Thanks @hoangphamEclipse for this effort; well appreciated. I'd have these comments, some minor edit ones and the others more "fundamental". There could be more work ahead given these potentially opening aspects.

doc/performance/0002-http-vs-rcp-client-performance-analysis.md
```diff
@@ -0,0 +1,94 @@
...
+However, this implementation has some downsides. Details can be found in the issue above. The proposed implementation is:
+
+```
+Theia FE <--JSON RPC--> Theia BE <--TSP over HTTP--> Trace Server
+```
+
+The goal of this document is to compare the performance of the TSP client before and after the change. Since we have an additional link in the communication chain, it is expected that the communication time will either be the same or increase. We would like to know how much is this increased, and how it affects the performance of the Theia Trace Extension.
+
+**Update**: It is confirmed that Theia no longer uses JSON-RCP for it front end (FE)-back end (BE) communication, but they use a faster technology for serializing JSON called `message pack` instead. So the actual implementation would be:
```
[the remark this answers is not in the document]
hoangphamEclipse: I am not sure what you mean by `another amended version`. If you meant that as `the version with message pack`, then:

After much investigation, I found out that previously, Theia used `vscode-ws-jsonrpc` as the binary codec to encode/decode rcp messages. Now it uses a new specialized library called `msgpackr` to encode/decode the rcp-messages. `msgpackr` is an implementation of `msgpack`, which is a binary serialization format. More on this decision can be found [here](https://github.com/eclipse-theia/theia/issues/11159). This change is published in Theia version 1.29.0, and the tests in this PR are done using Theia 1.37.2. Hence, I expected we are already testing with the code that has the `message pack` implementation. 

Even though in the code we use the class `JsonRpcConnectionHandler`, it actually defaults to `Message pack`, if I understand the code correctly.

doc/performance/0002-http-vs-rcp-client-performance-analysis.md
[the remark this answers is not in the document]
hoangphamEclipse: Ack

doc/performance/0002-http-vs-rcp-client-performance-analysis.md
hoangphamEclipse: Marked as resolved because test methodology has changed.

hoangphamEclipse: Marked as resolved because test methodology has changed.

doc/performance/0002-http-vs-rcp-client-performance-analysis.md
hoangphamEclipse: @MatthewKhouzam since you are back, do you think this is a required test case?

marco-miller (commented):
Leaving this review up to @MatthewKhouzam and others if need be, given my recent focus change; thanks.

doc/performance/0002-http-vs-rcp-client-performance-analysis.md
```diff
@@ -0,0 +1,244 @@
...
+    // Store the result in the variable to save a timestamp before returning
+    const result = await RestClient.get(url, undefined, array(OutputDescriptor));
+    // Save the timestamp when the HTTP request is finished
+    const end = Date.now();
+    // Output the timestamps to the debug console
+    console.log('TSPClient:: getAvailableOutputs:: client reached:', start, ', Promised resolved:', end);
+    return result;
+}
+```
+
```
bhufmann, line 175: Could you please add the code for fetchXYTree and fetchTimeGraphStates so that it's easier for someone who wants to reproduce the tests.

doc/performance/0002-http-vs-rcp-client-performance-analysis.md
```diff
@@ -0,0 +1,244 @@
...
+In all test cases, the time takes to reach the TSPClient from the Theia FE is pretty similar. In the first 3 test cases (1, 2, and 3a), it is inconclusive to say that there is a significant difference in speed. In test case 3b, by removing other factors such as parallel running HTTP requests, the time takes to reach the TSPClient in the JSON-RPC implementation is only `2ms` more than the current implementation, which is not really significant. However, payloads going to the Theia BE test here is small. Larger parameters might contribute more latency. However, we don't expect the request payload to become extremely large, such as the case of the response payload.
+
+### TSPClient to Front-end
+
+In all test cases, there is virtually no latency for returning the HTTP response to the front-end in the current implementation. However, there is significant latency when using JSON-RPC.Latency scales with the size of the response. The larger the response is, the longer the latency. When getting out descriptors, where we get a small response payload, the latency added is approximately `50ms`. In test case 3b, where we fetch around `116620 states`, latency can be `150ms`.
```
bhufmann, line 236: There is a space missing after `JSON-RPC.` and before `Latency`

doc/performance/0002-http-vs-rcp-client-performance-analysis.md
```diff
@@ -0,0 +1,244 @@
...
+
+### Total duration
+
+Surprisingly, overall, using JSON-RPC can improve the performance of the Trace Extension. The reason is that the performance gain in the TSPClient processing speed over-compensates for the latency added by JSON-RPC. In all test cases, the JSON-RPC implementation either performs the same as the current implementation (test case 1, 2) or improves the processing speed `up to 1.5 times` (test case 3b). That is a 50% performance gain.
+
+## Conclusion
+
+The implementation of JSON-RPC results in a gain in performance since the performance gained by the TSPClient. Its benefit outweighs the latency added by adding the Theia BE layer in the communication chain. This also implies that moving code to the BE might be an option to improve the performance of the Trace Extension in general.
```
bhufmann, line 244: I would remove the last sentence (`This also implies...`) since it's out of scope of the JSON-RPC investigation.

bhufmann (changes requested):

bhufmann (changes requested):
Thanks for the report of your investigation. It looks good overall from my side, only some small update requests.

doc/performance/0002-http-vs-rcp-client-performance-analysis.md
hoangphamEclipse: Ack

doc/performance/0002-http-vs-rcp-client-performance-analysis.md
hoangphamEclipse: Ack.

bhufmann (approved):
Looks good to me. Thanks!

@MatthewKhouzam  could you please do another round of review since you gave comments initially?

MatthewKhouzam (approved):
Good!

PR #1004, opened 2023-08-22
Add ADR about external analysis API in TSP

bhufmann:

### DIFF
--- a/doc/adr/0011-tsp-analysis-api.md
+++ b/doc/adr/0011-tsp-analysis-api.md
@@ -1,0 +1,273 @@
+# 11. Tsp analysis api
+
+Date: 2023-06-20
+
+## Status
+
+Proposed
+
+## Context
+
+The trace viewer currently is able to visualize trace data provided by a trace server over the trace server protocol (TSP). The Trace Compass server has some built-in analysis view for that. However, it is not possible to side-load analysis and visualization descriptions over the TSP so that end-user can provide some custom views. The Trace Compass supports loading of data-driven analysis and views, e.g. XML driven views or in-and-out anlysis of the Trace Compass incubator project. In the Eclispe-based Trace Compass application, there exists UI primitive to load e.g. XML files or configure custom analysis. While the Trace Compass server has the capablility to understand these defintions, there is no way to side-load this definition over the TSP. This ADR will propose an analysis service within the TSP to facilitate these custom analysis.
+
+### Global analysis service
+
+An analysis service for managing global analyses will be introduced. Global in this context means that the analysis definitions will be handled on the application level. 
+
+    GET /analysis/
+        returns a list of analysis type descriptors: typeId, name, description, expected input parameter list ("path" for file path)
+    GET /analysis/types
+        returns a map typeId -> list of analysis descriptor of existing analyses on server: analysisId, name, description, parameters
+    GET /analysis/types/{typeId}
+        returns a list of analysis descriptors for given typeId
+    POST /analysis/types/{typeId}
+        Upload an analysis definition for given typeId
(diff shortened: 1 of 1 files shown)


doc/adr/0011-tsp-analysis-api.md
```diff
@@ -0,0 +1,268 @@
...
+    parameters: map<string, object>
+}
+```
+
+Where:
+- `name`: Name of the analysis type. Can be shown to the end-user.
+- `description`: The description of the analysis type. Can be shown to the end-user.
+- `id`: Unique id of the analysis type. Used in the application to distinquish analysis types
+- `typeId`: ID of the analysis type.
+- `parameter`: A map of paramter corresponding to the `inputParameter` specified in the analysis type descriptor
```
PatrickTasse, line 73: parameter

doc/adr/0011-tsp-analysis-api.md
```diff
@@ -0,0 +1,268 @@
...
+    inputParameter: string[]
+}
+```
+
+Where:
+
+- `name`: Name of the analysis type. Can be shown to the end-user.
+- `description`: The description of the analysis type. Can be shown to the end-user.
+- `id`: Unique id of the analysis type. Used in the application to distinquish analysis types
+- `scope:` `experiment` for analysis per experiment or `global` for all experiments
```
PatrickTasse, line 51: Trying to understand how the scope would be used. From the sequences below, it seems that the experiment analysis descriptors are accessed using different endpoints with experiment/{expUUID} path. But these sequences do not seem to use any steps to get the AnalysisTypeDescriptor.

doc/adr/0011-tsp-analysis-api.md
```diff
@@ -0,0 +1,268 @@
...
+
+## Decision
+
+The change that we're proposing or have agreed to implement, will be implemented.
+
+## Consequences
+
+### Easier to do
+
+This will introduce new TSP endpoints and it's a a completely new feature for trace viewers supporting supporting these endpoints in the front-end and server back-end. Once implemented, it will greatly enhance the feature capabilities of the whole application. It will help end-users to define their custom analysis and visualization defintions, and allow them to get such features faster than having to write code in the server application, compile and re-deploy the server afterwards. This will reduce trouble-shooting times for the end-users.
```
PatrickTasse, line 259: troubleshooting

PatrickTasse (changes requested):

MatthewKhouzam (dismissed):
Can't find any typos. Also this makes a lot of sense, It will expand the functionality of theia's trace extension.

doc/adr/0011-tsp-analysis-api.md
bhufmann: > To be clear, is this a map where the key is the parameter name and the value(object) is the parameter value?

yes, so for the example of files, the key will be `path` and the value the actual file path.

doc/adr/0011-tsp-analysis-api.md
bhufmann: The AnalysisTypeDescriptor still go through the global end-point, where you will be able to see the types and create analysis descriptors. Then the experiment endpoint is to basically assign an analysis descriptor to an experiment (see post). With this you could assign a given analysis to multiple selected experiments, I think I have to update "Sequence: Create analysis instance for an experiment", because it's not really reflecting that.

doc/adr/0011-tsp-analysis-api.md
bhufmann: Done

PatrickTasse (approved):

MatthewKhouzam (approved):
LGTM!

PR #1006, opened 2023-08-30
Added output data changed signal

ngondalia:
Introduced a new signal to indicate that the data on the server has been updated and therefore the components should refetch the data and rerender the components. This PR handles the signal in table-output-component and timegraph-output-component. We have a use case where data is modified on the server, and we want the components to rerender with the new data. Currently, the user would have to close the active trace panel and reopen it to see the changes.

### DIFF
--- a/packages/base/src/signals/signal-manager.ts
+++ b/packages/base/src/signals/signal-manager.ts
@@ -30,6 +30,7 @@ export declare interface SignalManager {
     fireTraceServerStartedSignal(): void;
     fireUndoSignal(): void;
     fireRedoSignal(): void;
+    fireOutputDataChanged(outputs: OutputDescriptor[]): void;
     fireOpenOverviewOutputSignal(traceId: string): void;
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     firePinView(output: OutputDescriptor, payload?: any): void;
@@ -73,7 +74,8 @@ export const Signals = {
     SAVE_AS_CSV: 'save as csv',
     VIEW_RANGE_UPDATED: 'view range updated',
     SELECTION_RANGE_UPDATED: 'selection range updated',
-    REQUEST_SELECTION_RANGE_CHANGE: 'change selection range'
+    REQUEST_SELECTION_RANGE_CHANGE: 'change selection range',
+    OUTPUT_DATA_CHANGED: 'output data changed'
 };
 
 export class SignalManager extends EventEmitter implements SignalManager {
@@ -142,6 +144,9 @@ export class SignalManager extends EventEmitter implements SignalManager {
     }
     fireRedoSignal(): void {
         this.emit(Signals.REDO);
+    }
+    fireOutputDataChanged(outputs: OutputDescriptor[]): void {
+        this.emit(Signals.OUTPUT_DATA_CHANGED, outputs);
     }
     // eslint-disable-next-line @typescript-eslint/no-explicit-any, @typescript-eslint/explicit-module-boundary-types
     firePinView(output: OutputDescriptor, payload?: any): void {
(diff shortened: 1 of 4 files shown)


packages/base/src/signals/signal-manager.ts
```diff
@@ -143,6 +144,9 @@ export class SignalManager extends EventEmitter implements SignalManager {
     fireRedoSignal(): void {
         this.emit(Signals.REDO);
     }
+    fireOutputDataChanged(): void {
+        this.emit(Signals.OUTPUT_DATA_CHANGED);
```
bhufmann, line 148: Should the signal have a payload to indicate which data provider or data providers were changed? This would allow receivers to filter out signals and only act on it if it is meant for them? 

WDYT?

packages/react-components/src/components/utils/timegraph-container-component.tsx
```diff
@@ -33,14 +33,17 @@ export class ReactTimeGraphContainer extends React.Component<ReactTimeGraphConta
...
     componentWillUnmount(): void {
-        if (this.container) {
-            this.container.destroy();
-        }
         if (this._resizeHandler) {
             this.props.removeWidgetResizeHandler(this._resizeHandler);
         }
     }
 
+    destroyContainer(): void {
```
bhufmann, line 41: I get the following sonar lint error:

Remove this property or method or refactor "ReactTimeGraphContainer", as "destroyContainer" is not used inside component bodysonarlint(typescript:S6441)

see https://rules.sonarsource.com/javascript/RSPEC-6441/

It sounds worth fixing.

bhufmann (changes requested):
Thanks for the contribution. I have some comments.

bhufmann:
> Introduced a new signal to indicate that the data on the server has been updated and therefore the components should refetch the data and rerender the components. This PR handles the signal in table-output-component and timegraph-output-component. We have a use case where data is dynamically modified on the server, and we want the components to rerender with the new data. Currently, the user would have to close the active trace panel and reopen it to see the changes.

@ngondalia Interesting update. It sounds like that it is for some streaming / monitoring use case. If your server is modifying data, do you have some notification channel implemented from your custom server to the front-end? Please let me know.

packages/react-components/src/components/utils/timegraph-container-component.tsx
ngondalia: I had to do this on purpose for the following reasons:
- if the number of row models is 0, the container gets destroyed, which then disposes the layers that were added to it. This becomes a problem, because if the Output_Data_Changed signal is received in the timegraph output component and the row model size is greater than 0, updateChart would throw a bunch of errors because the layers were disposed. So, the way i fixed this for now is to destroy the container when the timegraph-output-component is unmounted.

packages/base/src/signals/signal-manager.ts
ngondalia: Yeah that is a good idea. I will add that in the next patch.

ngondalia:
> > Introduced a new signal to indicate that the data on the server has been updated and therefore the components should refetch the data and rerender the components. This PR handles the signal in table-output-component and timegraph-output-component. We have a use case where data is dynamically modified on the server, and we want the components to rerender with the new data. Currently, the user would have to close the active trace panel and reopen it to see the changes.
> 
> @ngondalia Interesting update. It sounds like that it is for some streaming / monitoring use case. If your server is modifying data, do you have some notification channel implemented from your custom server to the front-end? Please let me know.

Sorry, the word "dynamically" was misleading and wrong here. The data gets modified based on some trace type specific settings/filters coming from our custom webviews within our extension. The front end initiates the data modification, therefore we dont have the need for the server to notify the front-end of anything. It already knows that data has changed, as well as the type of data that has changed.

bhufmann (approved):
It looks good to me. To test I added a command to trigger the refresh. Thanks for the contribution.

PatrickTasse (approved):

PR #1009, opened 2023-09-07
Uplift theia dependencies from v1.37.2 to 1.40.1

bhufmann:
Uplift theia dependencies from v1.37.2 to 1.40.1

Perform a 'yarn upgrade' to keep the yarn.lock fully aligned.

Base this uplifted-to Theia version on community [1]'s below.

Fixed compilation error by making constructors of injectable classes
public.

Align to @postContruct changes of new inversify version [2]. Use Theia's
re-export dependencies for inversify according to [3]

Change electron-main location to ../src-gen/backend/ instead of
../src-gen/frontend due to breaking change described here [4]

[1] https://theia-ide.org/releases/
[2] https://github.com/eclipse-theia/theia/blob/master/doc/Migration.md#inversify-60
[3] https://github.com/eclipse-theia/theia/blob/master/packages/core/README.md#re-exports
[4] https://github.com/eclipse-theia/theia/blob/master/CHANGELOG.md#breaking_changes_1.39.0


marcdumais-work:
There are some breaking changes that occurred in Theia that I noticed have not yet been reflected here. It would probably be good to go through past entries of the [migration guide](https://github.com/eclipse-theia/theia/blob/master/doc/Migration.md) for other things that may have been missed.

1) I think it would be good to use Theia re-exports whenever possible. One example is `inversify`: 

`import { inject, injectable } from 'inversify';`

becomes:

`import { inject, injectable } from '@theia/core/shared/inversify';`

2) (at some point) same as 1, use Theia re-exports in general - there may be other similar cases

Mentioned here specifically about Electron, but it points to the [complete list of re-exports](https://github.com/eclipse-theia/theia/blob/master/packages/core/README.md#re-exports): 
https://github.com/eclipse-theia/theia/blob/master/doc/Migration.md#electron-update

marcdumais-work:
So far, the code looks good!

marcdumais-work:
I see the electron version was bumped, probably by the `yarn upgrade`. It would be good to update the version in  `electron-builder.yml`, to match. (23.3.13)

bhufmann:
> There are some breaking changes that occurred in Theia that I noticed have not yet been reflected here. It would probably be good to go through past entries of the [migration guide](https://github.com/eclipse-theia/theia/blob/master/doc/Migration.md) for other things that may have been missed.
> 
>     1. I think it would be good to use Theia re-exports whenever possible. One example is `inversify`:
> 
> 
> `import { inject, injectable } from 'inversify';`
> 
> becomes:
> 
> `import { inject, injectable } from '@theia/core/shared/inversify';`
> 
>     2. (at some point) same as 1, use Theia re-exports in general - there may be other similar cases
> 
> 
> Mentioned here specifically about Electron, but it points to the [complete list of re-exports](https://github.com/eclipse-theia/theia/blob/master/packages/core/README.md#re-exports): https://github.com/eclipse-theia/theia/blob/master/doc/Migration.md#electron-update

Thanks for pointers. I'll update the PR according to the migration guide. 

About the re-export comment I'll update the `inversify` dependency only in this PR, since there is also a new version. The others we can handle in a separate dedicated PR.

marcdumais-work (approved):
LGTM, thanks Bernd!

PR #1014, opened 2023-09-08
Use 'traceviewer-base/lib' instead of 'traceviewer-base/src'

bhufmann:
Some traceviewer-base imports still use the src directory instead of lib. This commit fixes this.


marcdumais-work (approved):
LGTM, thanks Bernd!

PR #1023, opened 2023-09-26
Move MaterialUI to proper package.json

williamsyang-work:
MaterialUI was installed in the top-level node_modules.  It's moved to the correct one in /packages/react-components/ where it's used.

This may fix eclipse-cdt-cloud/vscode-trace-extension/issues/175 according to this [stackoverflow thread](https://stackoverflow.com/questions/72413194/uncaught-typeerror-cannot-read-properties-of-null-reading-usecontext).

### DIFF
--- a/package.json
+++ b/package.json
@@ -127,7 +127,6 @@
   "dependencies": {
     "@emotion/react": "^11.10.5",
     "@emotion/styled": "^11.10.5",
-    "@mui/material": "^5.10.14",
     "@types/react": "^18.0.15",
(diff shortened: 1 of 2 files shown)


marcdumais-work (approved):
LGTM, thanks for this contribution! 

BTW, it would generally be desirable to use the latest baseline, for future pull-requests, or rebase the patch before posting it for review (e.g. `git fetch origin master && git rebase origin/master`). 

However, I do not think it would change anything in this case.

PR #1033, opened 2023-11-02
Call TSP endpoint to close an experiment when closing it in the UI

### LINKED ISSUE #1024: Close experiment and trace resources when closing an experiment
The TSP allows to open traces and experiments on the server back-end. However, there is now dedicated request available to dispose the back-end resources related to experiments and traces in the back-end. There should be a TSP request to trigger the disposal of resources in the back-end when the front-end close the experiment. 

The the following API would allow this:
`PUT /experiments/{expUUID}:close` 

Consider other alternatives before implement this API.

bhufmann:
With this it's possible to close an experiment on the server. Upon reception of the command the server can dispose allocated resources hence the client doesn't need them anymore.

This change follows the TSP update provided by:
https://github.com/eclipse-cdt-cloud/trace-server-protocol/pull/95

Fixes #1024

### DIFF
--- a/packages/base/src/experiment-manager.ts
+++ b/packages/base/src/experiment-manager.ts
@@ -122,6 +122,17 @@ export class ExperimentManager {
     }
 
     /**
+     * Close the given experiment from the server
+     * @param experimentUUID experiment UUID
+     */
+    async closeExperiment(experimentUUID: string): Promise<void> {
+        const experimentToDelete = this.fOpenExperiments.get(experimentUUID);
+        if (experimentToDelete) {
+            await this.fTspClient.closeExperiment(experimentUUID);
+        }
+    }
+
+    /**
      * Delete the given experiment from the server
      * @param experimentUUID experiment UUID
      */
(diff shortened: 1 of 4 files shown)


bhufmann:
There is on issue with this approach to close the experiment, because it introduces a state in the server and it becomes stateful. For example, if the UI client close an experiment and a CLI client is running an analysis, the analysis will fail. Vice versa, it a cli client close the experiment while the UI client is actually analyzing the experiment, it will fail.

More over, the Trace Compass server, generates row IDs for Timegraph and XY data providers, that are different after closing and reopening a trace. So, if the UI client opens the experiment and a Timegraph view and then the CLI client closes the experiment, even though, the experiment is re-opened when the UI client queries the Timegraph data provider again, the IDs to fetch row data are different and the query will result in empty graph. The UI client would have to close the view and re-open the view to make it work again.

So, to support closing of experiments and release back-end resources, concurrent queries, ongoing queries as well as IDs have to generated correctly.

PR #1034, opened 2023-12-05
Uplift Theia dependencies from v1.40.1 to 1.43.1

bhufmann:
Uplift Theia dependencies from v1.40.1 to 1.43.1

Perform a 'yarn upgrade' to keep the yarn.lock fully aligned.

Base it on Theia community release [1] below.

[1] https://theia-ide.org/releases/


marcdumais-work:
Which node version do you currently use? I got a small difference in `yarn.lock` after building this PR locally - can you check if you get the same, and if so, update the PR?

bhufmann:
> Which node version do you currently use? I got a small difference in `yarn.lock` after building this PR locally - can you check if you get the same, and if so, update the PR?

I'm using v16.20.0.

marcdumais-work (approved):
LGTM - I briefly tested using the tutorial traces.

It's not a huge deal in this case but I suggest you check-in the slightly modified `yarn.lock` - the dependency has not changed but apparently now yarn tags it as being from GitHub. This will prevent this change from showing in `git diff` for other developers.

PR #1036, opened 2023-12-08
docker: fix and improve docker example

frallax:
Before this commit the docker example was not properly building. The issue was introduced when moving from Theia 1.3x to 1.4x, where an extra step in the build process was introduced (i.e. running "yarn" was not enough, an extra "theia" command was needed).

After this commit:
- the docker example is building again
- the docker image generated is stable and reproducible, not based on "next"
- the docker image is much smaller thanks to a multistage build

### DIFF
--- a/examples/docker/Dockerfile
+++ b/examples/docker/Dockerfile
@@ -1,18 +1,38 @@
-FROM node:16-alpine
+FROM node:16.20.2-bookworm-slim as build
 
-RUN mkdir /home/app
-WORKDIR /home/app
-
-RUN apk add --no-cache bash git python3 make pkgconfig libsecret-dev g++
-# Use "bash" as replacement for "sh"
-RUN rm /bin/sh && ln -s /bin/bash /bin/sh
+RUN apt-get update && apt-get install -y \
+    git \
+    python3 \
+    make \
+    pkg-config \
+    libsecret-1-dev \
+    g++ \
+    libx11-dev \
+    libxkbfile-dev
 
 # Build the browser theia-trace-extension application
-COPY example-package.json /home/app/package.json
-RUN npx yarn
+COPY example-package.json /app/tte/package.json
+WORKDIR /app/tte/
+RUN yarn && \
+    npx theia build --app-target=\"browser\" --mode development && \
+    yarn autoclean --init && \
+    echo *.ts >> .yarnclean && \
+    echo *.ts.map >> .yarnclean && \
+    echo *.spec.* >> .yarnclean && \
+    yarn cache clean
 
-COPY startup.sh /home/app
-RUN ["chmod", "+x", "/home/app/startup.sh"]
+FROM node:16.20.2-bookworm-slim
 
-# Start the service
-ENTRYPOINT ["/home/app/startup.sh"]
+COPY --from=build /app/tte /app/tte
+
+RUN apt-get update && apt-get install -y \
+    libx11-6 \
+    libxkbfile1 \
+    libsecret-1-0 \
+    && rm -rf /var/lib/apt/lists/*
+ENV NODE_ENV production
+
+WORKDIR /app/tte
+COPY docker-entrypoint.sh /usr/local/bin
+EXPOSE 4000
+ENTRYPOINT ["docker-entrypoint.sh"]
--- a/examples/docker/example-package.json
+++ b/examples/docker/example-package.json
@@ -15,6 +15,9 @@
     }
   },
   "dependencies": {
+    "@emotion/react": "^11.10.5",
+    "@emotion/styled": "^11.10.5",
+    "@mui/material": "^5.10.14",
     "@theia/core": "1.43.1",
     "@theia/editor": "1.43.1",
     "@theia/filesystem": "1.43.1",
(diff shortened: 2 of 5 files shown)


bhufmann (changes requested):
It looks good and it works. I had to build with parameter "--network host" (`docker build --network host -t tte .`) to avoid a build failure:

#0 45.73 Err:1 http://deb.debian.org/debian bookworm InRelease
#0 45.73   Temporary failure resolving 'deb.debian.org'

Please update the README for that.

bhufmann (approved):
Looks good to me! Thanks for the contribution.

PatrickTasse (approved):

examples/docker/example-package.json
```diff
@@ -27,9 +30,8 @@
     "@theia/preferences": "1.43.1",
     "@theia/process": "1.43.1",
     "@theia/terminal": "1.43.1",
-    "@theia/vsx-registry": "1.43.1",
     "@theia/workspace": "1.43.1",
-    "theia-traceviewer": "next"
+    "theia-traceviewer": "0.2.0-next.20231206220005.ff0943f.0"
```
marcdumais-work, line 34: Have you considered using `"next"` as version? This way it would not be necessary to update that `package.json` as often. If there's even an issue with the latest `next`, one could fix it with a PR, merge it and after it's published rebuild the Docker image.

marcdumais-work: Or leaving it as "next" I guess

examples/docker/Dockerfile
```diff
@@ -1,18 +1,38 @@
-FROM node:16-alpine
+FROM node:16.20.2-bookworm-slim as build
```
marcdumais-work, line 1: I am a bit puzzled by this very specific version being used. What's the thoughts behind this? I guess the "slim" part means it's a streamlined image, which is good. If `16.20.2` is the latest `16` of that image, I think "node:16-bookworm-slim" would resolve to the same image, and leave the door open to eventually pick-up an updated `16.20.3` if there's ever one, without us having to update it manually.

examples/docker/example-package.json
marcdumais-work: I see a related  mention in the description: 

> the docker image generated is stable and reproducible, not based on "next"

Is that referring to the `theia-traceviewer` extension or using the specific version of the base Docker image? Stability has been an issue?

examples/docker/Dockerfile
bhufmann: @frallax could you please answer @marcdumais-work questions?

examples/docker/example-package.json
frallax: I'll try to answer all 3 comments at once, please ping if it is not clear.

Regarding the `"next"` questions:

The goal of this "example/docker" is to provide an example of how a user could create an image of the theia-trace-extension. I think that if the user plans to build and use such image in a stable and controlled product, the user should be able to have a proper way to control which version of the theia-trace-extension he/she is using. Using `"next"` would mean that every time the user builds a image, a new theia-trace-extension version could be included. Hence it would be non reproducible. Note also that:
- with the proposed change, it is not needed to update this docker at all, apart if we want to keep it up to date with the latest and greatest
- the project is not pushing images anywhere

Regarding the "stable and reproducible" topic (which I partly touched above):

Maybe here I used some terms that could be misunderstood. But "reproducible" is connected to the concept above (i.e. not using next) so that a developer could use git history to go back to a dockerfile that generated a specific image. The "stable" term is related to the image build process. Before this commit I often got issues with "vscode-ripgrep" while building the image, due to API rate limit exceed. After this commit, I have not seen this issue so often any longer, and I suspect it is due to the removal of the vsx-registry (which is however not needed for this example)

examples/docker/Dockerfile
frallax: I am not a docker expert. The slim part is as you say (also if this is not so important for the build stage). Regarding the specific version, I just prefer being precise with versions to make things as reproducible as possible.

But the real reason for using this specific image is that I saw a similar project using this image and I tried that. My final goal was to shrink the size of the final image, and by using this multistage approach we do that, compared to what was on master before. I am sure there is room for improvement (both in term of final image size and reproducibility), and if you have any proposal I would be happy to review and test.

examples/docker/example-package.json
marcdumais-work: @frallax thank you for the response. I guess we have he a trade-off between maintenance effort and stability of included building-blocks. I'm ok with what you propose. Note that no matter what, having a `npm` app, we will never have a 100% certainty the same version of every dependency will be used, from one day to the next. This is doubly true in this case where we do not preserve the `yarn.lock` when building the image. 

> Before this commit I often got issues with "vscode-ripgrep" while building the image, due to API rate limit exceed

I can confirm that `@vscode/ripgrep` is no longer pulled, now that `@theia/vsx-registry` is removed. The dependency was indirect, through the Theia plugin system. So long as the resulting Docker appliance does not require running vscode extensions, we're fine.

marcdumais-work (approved):
LGTM, thanks @frallax !

examples/docker/Dockerfile
marcdumais-work: I think using multi-stage build makes a lot of sense - nice improvement.

PR #1037, opened 2023-12-15
[ci] Use GitHub token to avoid ripgrep download issue

marcdumais-work:
In a recent CI run, I noticed this [error](https://github.com/eclipse-cdt-cloud/theia-trace-extension/actions/runs/7223312899/job/19682105853):
"Downloading ripgrep failed: Error: Request failed: 403"

This can randomly happen on CI machines, because GitHub has limits when anonymously accessing its API, which we do when downloading `ripgrep`. There's an easy fix: use an auto-generated GitHub token when building the Theia app.

Also tidied-up the workflow file a little, to make it easier to read.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,4 +1,4 @@
-name: build
+name: CI
 
 on:
   push:
@@ -9,6 +9,7 @@ on:
 jobs:
 
   build-and-test:
+    name: Build and test
     runs-on: ${{ matrix.os }}
 
     strategy:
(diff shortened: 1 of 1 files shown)


marcdumais-work:
While in there, I have cleaned-up the labels used for the workflow, i.e. its name and the build and step names. I think it will look nicer in the PR "checks" section and also in the detailed `Actions` page. Let me know if you do not like.

bhufmann (approved):
Looks good to me.  Thanks!

PR #1047, opened 2024-02-28
Theia 1.45.1

marcdumais-work:
Update the example apps to use the latest Theia community release. 

There was an issue after the update, that required some digging. It turns-out that one of our backend contribution was no longer correctly following the expected life-cycle of a Theia application, causing the app to exit before being fully started. 

TL;DR: our contribution would wait in the "initialize" phase until the application was fully started and one of our frontend contribution would send the trace server port. This has been refactored so that our backend contribution configures itself assuming the default port is used, and corrects this once the application is started (the port is only potentially needed later, when the user opens a trace). 

See commit message for more details.

Note: because of the above, a bit more testing than usual is probably a good idea


examples/docker/example-package.json
```diff
@@ -18,23 +18,23 @@
...
+    "@theia/keymaps": "1.45.1",
+    "@theia/markers": "1.45.1",
+    "@theia/messages": "1.45.1",
+    "@theia/monaco": "1.45.1",
+    "@theia/navigator": "1.45.1",
+    "@theia/preferences": "1.45.1",
+    "@theia/process": "1.45.1",
+    "@theia/terminal": "1.45.1",
+    "@theia/workspace": "1.45.1",
     "theia-traceviewer": "0.2.0-next.20231206220005.ff0943f.0"
```
bhufmann, line 34: This needs to be updated after the a new version is available.

bhufmann (approved):

PR #1049, opened 2024-02-29
[Theia v1.45.1] update trace extension used in docker image

marcdumais-work:
Follow-up, after updating to Theia 1.45.1. Update the specific version of the trace extension that's used in the docker image example.


bhufmann (approved):
I tested the docker example and it works. Thanks for updating this.

marcdumais-work:
Thanks for the review!

PR #1054, opened 2024-03-08
Support opening trace files through a dialog

### LINKED ISSUE #1053: Handle opening of single trace files via dialog
For users that primarily need to analyze single file traces, the current implementation breaks the user flow. If the user wants to add another trace, they would have to switch to explorer view and open it from there. It would be ideal if they can open single traces files in the same manner as trace folders/groups.

There are few options that I can think of to solve this (open to any other suggestions):

1) Have some sort of setting that can control the behaviour of "open trace" button at the extension level - At its core, it would control if the dialog opened will allow selecting files vs folders. I dont prefer this option as it doesnt address the issue for users that may have the need to analyze both trace folders and files.

2) "Open Trace" button handler can invoke a quick pick before showing the dialog so that users can select if they want to open file vs folder. I also am not a fan of this as it just adds more buttons to click to open a trace.

3) (My preference) We can render two buttons, "Open Trace Folder" and "Open Trace File". But additionally, ReactExplorerPlaceholderWidget can render one or both buttons based on its properties. The extension can control what button(s) get rendered based on a setting/api. I think this is a more complete solution as base extension and extending extensions can control the behaviour of what trace resource type will be handled. (Background: we only deal with single trace files and have no need for open trace folders, so i think it would be neat to not have a button for opening trace folder and only for opening trace file)

I would love get some feedback and/or suggestions!

ngondalia:
This commit adds another button for opening trace files through a dialog. Futhermore, the theia-trace-explorer-placeholder-widget can be configured to render either one or both buttons (Open Trace folder, Open Trace File).

Fixes #1053

Signed-off-by: Neel Gondalia <ngondalia@blackberry.com>

### DIFF
--- a/packages/react-components/src/trace-explorer/trace-explorer-placeholder-widget.tsx
+++ b/packages/react-components/src/trace-explorer/trace-explorer-placeholder-widget.tsx
@@ -2,9 +2,15 @@ import { faSpinner } from '@fortawesome/free-solid-svg-icons';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import * as React from 'react';
 
+export enum TraceResourceType {
+    FOLDER = 'Folder',
+    FILE = 'File'
+}
+
 export interface ReactPlaceholderWidgetProps {
     loading: boolean;
-    handleOpenTrace: () => void;
+    handleOpenTrace: (type?: TraceResourceType) => void;
+    renderOnly?: TraceResourceType;
 }
 
 export class ReactExplorerPlaceholderWidget extends React.Component<ReactPlaceholderWidgetProps, unknown> {
@@ -16,19 +22,28 @@ export class ReactExplorerPlaceholderWidget extends React.Component<ReactPlaceho
         return (
             <div className="placeholder-container" tabIndex={0}>
                 <div className="center">{'You have not yet opened a trace.'}</div>
-                <div className="placeholder-open-workspace-button-container">
-                    <button
-                        className="plcaeholder-open-workspace-button"
-                        title="Select a trace to open"
-                        onClick={this.props.handleOpenTrace}
-                        disabled={this.props.loading}
-                    >
-                        {this.props.loading && <FontAwesomeIcon icon={faSpinner} spin style={{ marginRight: '5px' }} />}
-                        {this.props.loading && <span>Connecting to trace server</span>}
-                        {!this.props.loading && <span>Open Trace</span>}
-                    </button>
-                </div>
+                {this.renderButton(TraceResourceType.FOLDER, 'Open Trace Folder')}
+                {this.renderButton(TraceResourceType.FILE, 'Open Trace File')}
             </div>
         );
+    }
+
+    renderButton(type: TraceResourceType, title: string): React.ReactNode {
+        const btnTitle =
+            type === TraceResourceType.FILE ? 'Select a single trace file to open' : 'Select a trace folder to open';
+        return this.props.renderOnly === undefined || this.props.renderOnly === type ? (
+            <div className="placeholder-open-workspace-button-container">
+                <button
+                    className="plcaeholder-open-workspace-button"
+                    title={btnTitle}
+                    onClick={() => this.props.handleOpenTrace(type)}
+                    disabled={this.props.loading}
+                >
+                    {this.props.loading && <FontAwesomeIcon icon={faSpinner} spin style={{ marginRight: '5px' }} />}
+                    {this.props.loading && <span>Connecting to trace server</span>}
+                    {!this.props.loading && <span>{title}</span>}
+                </button>
+            </div>
+        ) : undefined;
     }
 }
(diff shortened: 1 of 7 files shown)


ngondalia:
https://github.com/eclipse-cdt-cloud/theia-trace-extension/assets/113199503/86ab53bb-d24c-4745-a756-46372c044d71

theia-extensions/viewer-prototype/src/browser/trace-viewer/trace-viewer-toolbar-commands.ts
```diff
@@ -44,10 +44,16 @@ export namespace TraceViewerToolbarCommands {
     };
     export const OPEN_TRACE: Command = {
```
hriday-panchasara, line 45: Maybe rename OPEN_TRACE command and id to OPEN_TRACE_FOLDER. I always seem to get confused with all the Open Trace Command references in the codebase

theia-extensions/viewer-prototype/src/browser/trace-viewer/trace-viewer-toolbar-contribution.tsx
```diff
@@ -330,6 +343,12 @@ export class TraceViewerToolbarContribution implements TabBarToolbarContribution
             group: 'navigation',
             onDidChange: this.onMakerSetsChangedEvent
         });
+        registry.registerItem({
+            id: TraceViewerToolbarCommands.OPEN_TRACE_FILE.id,
+            command: TraceViewerToolbarCommands.OPEN_TRACE_FILE.id,
+            tooltip: TraceViewerToolbarCommands.OPEN_TRACE_FILE.label,
+            priority: 7
```
hriday-panchasara, line 350: The item registered with the command 'TraceViewerToolbarCommands.MARKER_SETS.id' also has the same id (7). Although these items won't cause conflict as they are not displayed in the same toolbar, you might consider adjusting the priorities so that its unique for each item

hriday-panchasara (changes requested):
Overall, I like this approach of allowing the extension to control which option gets rendered. The UX would now have to explicitly state 'Open Trace File' or 'Open Trace Folder'
- In `trace-viewer-commands.ts`, could you rename  the label of the OpenTraceCommand to 'Open Trace Folder'? This is so that if a user decides to open a trace through the command palette they know what type of traces they can access. You could even add a command for 'Open Trace File' in this case.

ngondalia:
> Overall, I like this approach of allowing the extension to control which option gets rendered. The UX would now have to explicitly state 'Open Trace File' or 'Open Trace Folder'
> 
> * In `trace-viewer-commands.ts`, could you rename  the label of the OpenTraceCommand to 'Open Trace Folder'? This is so that if a user decides to open a trace through the command palette they know what type of traces they can access. You could even add a command for 'Open Trace File' in this case.

Thanks for reviewing! I have addressed your comments.

theia-extensions/viewer-prototype/src/browser/trace-viewer/trace-viewer-toolbar-contribution.tsx
ngondalia: Ah yes - i missed that! I will make them unique

theia-extensions/viewer-prototype/src/browser/trace-viewer/trace-viewer-toolbar-commands.ts
ngondalia: Yeah it happens to me as well - i will rename them

theia-extensions/viewer-prototype/src/browser/trace-viewer/trace-viewer-contribution.ts
```diff
@@ -71,15 +73,16 @@ export class TraceViewerContribution
...
         let healthResponse;
         try {
             healthResponse = await this.tspClient.checkHealth();
         } catch (err) {
             // continue to start trace server
         }
+        const selectFiles = type && type === TraceResourceType.FILE ? true : false;
         if (healthResponse && healthResponse.isOk() && healthResponse.getModel()?.status === 'UP') {
-            this.openDialog(rootPath);
+            this.openDialog(rootPath, selectFiles);
```
bhufmann, line 85: I think we should change the title of the dialog. Right now the title is `Open Trace` for both cases. It might not be clear for the user. Maybe change the title to `Open Trace File`/`Open Trace Folder` or `Select Trace File to Open`/`Select Trace Folder to Open`.

packages/react-components/src/trace-explorer/trace-explorer-placeholder-widget.tsx
```diff
@@ -16,19 +22,26 @@ export class ReactExplorerPlaceholderWidget extends React.Component<ReactPlaceho
...
             </div>
         );
     }
+
+    renderButton(type: TraceResourceType, title: string): React.ReactNode {
+        return this.props.renderOnly === undefined || this.props.renderOnly === type ? (
+            <div className="placeholder-open-workspace-button-container">
+                <button
+                    className="plcaeholder-open-workspace-button"
+                    title="Select a trace to open"
```
bhufmann, line 36: This should be different for each case to be clear to the user, maybe 'Select a single trace file to open' for file traces and 'Select a trace folder to open`

bhufmann (changes requested):

packages/react-components/src/trace-explorer/trace-explorer-placeholder-widget.tsx
ngondalia: Ok - I will update it in the next patch!

bhufmann:
@hriday-panchasara  could you please update your review?

hriday-panchasara (approved):
Looks good, thanks for the contribution!

PR #1055, opened 2024-03-12
[viewer-prototype] Use Theia re-export rather than importing inversify directly

marcdumais-work:
This way, we re-use inversify through @theia/core, rather than importing it without declaring it as a dependency in our package.json (i.e. implicit dependency).

This also follows Theia's guidance regarding using their re-exports:
https://github.com/eclipse-theia/theia/tree/master/packages/core#re-exports-mechanism


bhufmann (approved):
Looks good to me. Thanks.

marcdumais-work:
Thanks for the review!

PR #1056, opened 2024-03-14
[ci] add infrastructure for automated releases/publish to npm (latest)

marcdumais-work:
This PR adds infrastructure to create GitHub releases (including auto-generated release notes) and publish `latest` npm packages, as part of the CI of this repository. See updated root README for instructions about how to trigger a release.

Note: it's expected that the "GitHub release" job will fail at this point.


.github/workflows/release.yml
```diff
@@ -0,0 +1,37 @@
...
+          release_file: 'RELEASE'
+          # Actions that run using the auto-generated GitHub token are
+          # not allowed to trigger a new workflow run. In this case we want
+          # the tag created by actions-gh-release to re-trigger the main workflow
+          # and result in publishing the package to npm.
+          # The following scopes are required when creating the committer token:
+          #  - repo:status, repo_deployment, public_repo, read:org
+          # See here for more details:
+          # https://docs.github.com/en/actions/security-guides/automatic-token-authentication#using-the-github_token-in-a-workflow
+          token: ${{ secrets.GH_COMMITTER_TOKEN }}
```
marcdumais-work, line 36: I added that secret/token

RELEASE
```diff
@@ -0,0 +1,8 @@
+tag: v0.0.0
```
marcdumais-work, line 1: I created and pushed that tag, on commit `412b849 - Trace viewer prototype application` (the second commit in this repo, right after "Initial commit"). This way, the first generated release notes will go back to the start of the repo's history (we can edit them after as needed)

bhufmann: There is already a 0.1.0 release that was manually pushed to npm and a tag in the repo. Shouldn't you base the new release on the 0.1.0 tag instead?

marcdumais-work: > We could, but then the auto-generated release notes would start at v0.1.0. The plan is, after the next release, I will take the release notes and move those relevant to v0.1.0 in a manually-created GitHub release (from that tag). I did similar for time-line chart: https://github.com/eclipse-cdt-cloud/timeline-chart/releases

marcdumais-work: timeline-chart: original release notes / GH release, that I manually massaged after release:
https://github.com/eclipse-cdt-cloud/timeline-chart/pull/274#issuecomment-1973707287

bhufmann: Right. Forgot about it. Makes sense.

bhufmann:
Is it still a draft or ready for review?

marcdumais-work:
> Is it still a draft or ready for review?

Yes, I think so. I will transition it out of "draft".

RELEASE
marcdumais-work: BTW, I'm not sure I was totally clear - the discussed tag (v0.0.0) will only affect the starting point for the auto-generated release notes, when we proceed with the first release, in another PR. This PR here will not result in a release, just setting the stage.

bhufmann: thanks for clarifying

bhufmann:
> > Is it still a draft or ready for review?
> 
> Yes, I think so. I will transition it out of "draft".

I'm going to review it. It looks good so far.

bhufmann (approved):
Thanks for this contribution. It will make it easy to produce release.

marcdumais-work:
Thanks for the review!

PR #1065, opened 2024-04-04
Add Critical path configuration menu and critical path view

bhufmann:
Note that code is hard-coded. A generic way is needed to configure
the menu and command-handler driven by the server.

Alternatively, this feature could be added in a third-party, trace
compass specific front-end extension(s) (i.e. Theia extension/Vscode
extension).

Requires: https://github.com/eclipse-cdt-cloud/tsp-typescript-client/pull/104​


bhufmann:
Just a thought. The time graph entry metadata for the thread status data provider has already the TID. It misses the hostId. Maybe, it's not necessary to retrieve the context if the host Id is provided as well. 

We could the have only the menu definition (coming from the server, once implemented) indicate that these parameter are meant to be included query parameters for other data providers. 

The only disadvantage is that the hostId is redundant and will increase the size of the data model return from fetchTree().

bhufmann:
PoC only, not needed at the moment.

PR #1071, opened 2024-04-25
Streamline Theia example applications, making them RCP-like

marcdumais-work:
Now that there exists is a full-fledge Theia IDE application [1] that incorporates the theia-trace-viewer, the example applications (browser, electron) in this repo, need not be more than "trace viewer appliances", that are used only for testing purpose. Such applications need contain little more than the trace viewer itself.

The major advantage of streamlining the test apps is that they will build faster, locally and in CI. They will also no longer contain/use outdated vscode built-in extensions, which add time to the build, fetching them from openvsx. Finally, it makes the test applications what they need to be for testing, nothing more, nothing less.

This commit removes most Theia framework extensions for the two example applications, transforming them into something relatively close to appliances. The exact list of retained Theia extensions is up for debate, and I can re-add some if they are deemed useful for testing. 

There remain a few remnants like the outline and problems views, that are still pulled by essential Theia framework extension, that we cannot easily remove.

[1] cdt-cloud blueprint :
https://github.com/eclipse-cdt-cloud/cdt-cloud-blueprint

After this change, here's that the example app looks-like:
Screenshot from 2024-04-25 15-12-09

After starting "clean" (e.g. trigger command "Reset workbench layout"):
Screenshot from 2024-04-25 15-09-27

Potential future work / improvements (not part of the scope of this PR): 
https://github.com/eclipse-cdt-cloud/theia-trace-extension/issues/1072


marcdumais-work:
IP Tickets for failed 3PP license checks:
- https://gitlab.eclipse.org/eclipsefdn/emo-team/iplab/-/issues/14579
- https://gitlab.eclipse.org/eclipsefdn/emo-team/iplab/-/issues/14580
- https://gitlab.eclipse.org/eclipsefdn/emo-team/iplab/-/issues/14581

marcdumais-work:
> We should however, keep the file explorer

+1 (I added it back, and in the docker example too, from which it was removed a while ago)

> Keeping the editor is optional. It would only help opening a config file.

Agreed. FYI, I did not explicitly keep `@theia/editor`, but it's ATM pulled because @theia/monaco depends on it, itself pulled by @theia/preferences that we need/use. I think this dependency chain exists because it's possible through the preferences UI to trigger the underlying .json preferences files to be opened in an editor, like so:

image

> Same for terminal. Is not really needed, but one might want to use a cli client to the trace server to test

Agreed. I have re-added `@theia/terminal` in the test apps. Do you think we should re-add it to the Docker example too?

> Please note that external contributors deployed the example electron app on their server for people to try. It's not maintained anymore, see [here](https://github.com/eclipse-cdt-cloud/theia-trace-extension?tab=readme-ov-file#download-an-external-build-of-the-application). We should probably update the README, since the deployment of the example app is not a mandate of the cdt-cloud project.

I have updated the README. It now points to a downloadable cdt cloud blueprint.

bhufmann:
> 
> Agreed. I have re-added `@theia/terminal` in the test apps. Do you think we should re-add it to the Docker example too?
> 

I think it's ok to leave it our here. Giving user access to the container via terminal is maybe not what we want. If downstream projects require that, then they can add it in their `package.json`.

bhufmann (approved):
Thanks. Makes builds faster and they are no great examples for downstream projects.

marcdumais-work:
Thanks for the review!

PR #1075, opened 2024-05-06
Refactor Item Properties View to use table components

ngondalia:
The current implentation of the item properties view shows properties as plaintext with no styling. This looks ok but can be significantly improved by showing the properties in a table.

 This commit makes the following changes:
- Reuses filter tree component for item properties view.
- Added option to hide filler columns/cells
- Ability to render a table cell value as an element
- Added tooltips for item properties

Signed-off-by: Neel Gondalia <ngondalia@blackberry.com>


ngondalia:
itemProps

bhufmann (approved):
Looks good to me. Thanks!

PR #1087, opened 2024-06-21
Theia release 1.50

bhufmann:


marcdumais-work:
ATM the electron rebuilds fails, when rebuilding the native dependencies. Forcing dependency "nan" to "2.18.0", in the main package.json, seems to resolve the problem:

image

bhufmann:
> ATM the electron rebuilds fails, when rebuilding the native dependencies. Forcing dependency "nan" to "2.18.0", in the main package.json, seems to resolve the problem:
> 

Thanks for finding the fix. As you mentioned, mainline Theia also applied this. I tested it and it works. I'll update this PR so that we have available when doing the next upgrade.

marcdumais-work:
I think we do not need this PR anymore, having skipped over Theia 1.50.0 and gone directly to 1.52.0. Closing.

PR #1090, opened 2024-07-10
Filtering in Timeline-Chart components

williamsyang-work:
Enables filtering in gant-chart views.  Implements UI and functionality.

This change is dependent on changes in the Timeline-Chart repo for the filtering of gap-states.
See: https://github.com/eclipse-cdt-cloud/timeline-chart/pull/285

Signed-off-by: William Yang <william.yang@ericsson.com>

### DIFF
--- a/packages/react-components/src/components/timegraph-output-component.tsx
+++ b/packages/react-components/src/components/timegraph-output-component.tsx
@@ -34,6 +34,7 @@ import { faSpinner } from '@fortawesome/free-solid-svg-icons';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import TextField from '@mui/material/TextField';
 import InputAdornment from '@mui/material/InputAdornment';
+import Chip from '@mui/material/Chip';
 import { debounce } from 'lodash';
 import '../../style/react-contextify.css';
 import { Item, ItemParams, Menu, Separator, Submenu, useContextMenu } from 'react-contexify';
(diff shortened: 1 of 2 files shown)


bhufmann (commented):
Thanks a lot for this contribution. 

I tried contribution and I was able to filter out states in time graph view.  I couldn't review or test more before I'm out-of-office for vacation. But here is what I noticed that should be addressed in this PR or follow-up PRs:

- When rows are empty (no states) then the row should be hidden as well.
- Arrows are still drawn for states that are hidden
- Annotations are not filtered.

The last point might need some back-end updates.

Looking forward having this available.

packages/react-components/src/components/timegraph-output-component.tsx
```diff
@@ -835,6 +839,45 @@ export class TimegraphOutputComponent extends AbstractTreeOutputComponent<Timegr
...
+        );
+    };
+
+    private updateSearchFilters = () => {
+        const filterExpressionsMap: { [key: number]: string[] } = {};
+        if (this.state.searchString) {
+            filterExpressionsMap[1] = [this.state.searchString]; // For greying out
+        }
+        if (this.state.filters.length > 0) {
+            filterExpressionsMap[4] = this.state.filters; // For filtering
```
PatrickTasse, line 871: Could we use constants for these key values?
const DIMMED = 1; // DIMMED filter expression property key
const EXCLUDED = 4; // EXCLUDED filter expression property key

PatrickTasse (changes requested):

packages/react-components/src/components/timegraph-output-component.tsx
williamsyang-work: Done!

PatrickTasse (approved):

bhufmann (approved):
Looks good to me. Thanks.

PR #1106, opened 2024-08-29
Adopt latest timeline-chart and tsp-typescript-client

marcdumais-work:
We have just published new versions of these key dependencies, with fresh "yarn upgrade" performed on them.


bhufmann (approved):
Looks good to me. Thanks for the upgrade.

marcdumais-work:
Thanks for the review!

PR #1109, opened 2024-09-12
Re-enable all React table-renderer-components tests

### LINKED ISSUE #1100: Re-enable table-components.test.tsx tests
In #1094, we upgraded `ag-grid` several major versions. After the update, we noticed that the strategy used in `table-components.test.tsx`, to obtain a `GridApi` object with which to work, seems no longer supported due to the greater encapsulation of the more recent `ag-grid`. We decided to [disable these tests](https://github.com/eclipse-cdt-cloud/theia-trace-extension/pull/1094/commits/41901d05cc11cc517103dc6e7fbff80d1a852bf5) temporarily, so the upgrade could go through. 

The goal of this ticket is to re-enable these tests in some form.

marcdumais-work:
Fixes #1100

Had to mock some ag-grid objects that were used in the newly re-enabled tests.

It turns-out that many of these objects were not really used during our tests
(e.g. api: GridApi, column: Column). So for these, it was enough to mock
them using an empty JS object and use keyword "as" to present them as the real
thing. e.g.:

```JS
<SearchFilterRenderer
    api={{} as GridApi}
    column={{} as Column<any>}
>
```

To run the re-enabled tests against the original jest snapshots, you can checkout the snapshot file's version prior to the ag-grid upgrade:

```bash
$ pwd
<...>/theia-trace-extension

$ git log --oneline 
[...newer commits]
b911f00 Bump ag-grid-community from 28.2.1 to 32.0.1  // **6-deep vs this PR**
167c526 Update 0011-tsp-analysis-api for custom analysis // **7-deep vs this PR**
[... older commits]

$ git checkout HEAD~7 ./packages/react-components/src/components/__tests__/__snapshots__/table-renderer-components.test.tsx.snap
$ yarn && yarn test
```

Expected result: a single test should fail - one that was not disabled: `<TableOutputComponent /> › Renders AG-Grid table with provided props & state`
I believe that test fails/failed because of the newer version of ag-grid generates slightly different HTML code when rendered. 

Confirm that all the newly re-enabled tests passed, using the original snapshot.

### DIFF
--- a/packages/react-components/src/components/__tests__/__snapshots__/table-renderer-components.test.tsx.snap
+++ b/packages/react-components/src/components/__tests__/__snapshots__/table-renderer-components.test.tsx.snap
@@ -1,5 +1,79 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`<TableOutputComponent /> Cell renderer 1`] = `"test cell"`;
+
+exports[`<TableOutputComponent /> Cell renderer with search selection 1`] = `
+Array [
+  <span>
+    <span
+      style={
+        Object {
+          "backgroundColor": "#FFFF00",
+        }
+      }
+    >
+      test
+    </span>
+  </span>,
+  <span>
+     cell
+  </span>,
+]
+`;
+
+exports[`<TableOutputComponent /> Empty search filter renderer 1`] = `
+<div
+  data-testid="search-filter-element-parent"
+  onClick={[Function]}
+  onMouseEnter={[Function]}
+  onMouseLeave={[Function]}
+>
+  <svg
+    aria-hidden="true"
+    className="svg-inline--fa fa-search fa-w-16 "
+    data-icon="search"
+    data-prefix="fas"
+    focusable="false"
+    role="img"
+    style={
+      Object {
+        "marginLeft": "10px",
+      }
+    }
+    viewBox="0 0 512 512"
+    xmlns="http://www.w3.org/2000/svg"
+  >
+    <path
+      d="M505 442.7L405.3 343c-4.5-4.5-10.6-7-17-7H372c27.6-35.3 44-79.7 44-128C416 93.1 322.9 0 208 0S0 93.1 0 208s93.1 208 208 208c48.3 0 92.7-16.4 128-44v16.3c0 6.4 2.5 12.5 7 17l99.7 99.7c9.4 9.4 24.6 9.4 33.9 0l28.3-28.3c9.4-9.4 9.4-24.6.1-34zM208 336c-70.7 0-128-57.2-128-128 0-70.7 57.2-128 128-128 70.7 0 128 57.2 128 128 0 70.7-57.2 128-128 128z"
+      fill="currentColor"
+      style={Object {}}
+    />
+  </svg>
+</div>
+`;
+
+exports[`<TableOutputComponent /> Loading renderer in loading 1`] = `
+<svg
+  aria-hidden="true"
+  className="svg-inline--fa fa-spinner fa-w-16 "
+  data-icon="spinner"
+  data-prefix="fas"
+  focusable="false"
+  role="img"
+  style={Object {}}
+  viewBox="0 0 512 512"
+  xmlns="http://www.w3.org/2000/svg"
+>
+  <path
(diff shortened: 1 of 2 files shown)


bhufmann (commented):
It works fine. I get a new warning when running the test with `yarn test` which was not there before. Do you see the same:
```
traceviewer-react-components:  PASS  src/components/__tests__/table-renderer-components.test.tsx (26.36 s)
traceviewer-react-components:   ● Console
traceviewer-react-components:     console.error
traceviewer-react-components:       Warning: An update to SearchFilterRenderer inside a test was not wrapped in act(...).
traceviewer-react-components:       
traceviewer-react-components:       When testing, code that causes React state updates should be wrapped into act(...):
traceviewer-react-components:       
traceviewer-react-components:       act(() => {
traceviewer-react-components:         /* fire events that update state */
traceviewer-react-components:       });
traceviewer-react-components:       /* assert on the output */
traceviewer-react-components:       
traceviewer-react-components:       This ensures that you're testing the behavior the user would see in the browser. Learn more at https://reactjs.org/link/wrap-tests-with-act
traceviewer-react-components:           at SearchFilterRenderer (/home/eedbhu/git/theia-training/theia-trace-extension/packages/react-components/src/components/table-renderer-components.tsx:88:9)
traceviewer-react-components: 
traceviewer-react-components:       209 |                 result.then(isFound => {
traceviewer-react-components:       210 |                     // need to trigger backend query
traceviewer-react-components:     > 211 |                     this.setState({
traceviewer-react-components:           |                          ^
traceviewer-react-components:       212 |                         isSearching: false,
traceviewer-react-components:       213 |                         isSearchSuccessful: isFound
traceviewer-react-components:       214 |                     });
traceviewer-react-components: 
traceviewer-react-components:       at printWarning (../../node_modules/react-dom/cjs/react-dom.development.js:86:30)
traceviewer-react-components:       at error (../../node_modules/react-dom/cjs/react-dom.development.js:60:7)
traceviewer-react-components:       at warnIfUpdatesNotWrappedWithActDEV (../../node_modules/react-dom/cjs/react-dom.development.js:27628:9)
traceviewer-react-components:       at scheduleUpdateOnFiber (../../node_modules/react-dom/cjs/react-dom.development.js:25547:5)
traceviewer-react-components:       at Object.enqueueSetState (../../node_modules/react-dom/cjs/react-dom.development.js:17925:7)
traceviewer-react-components:       at SearchFilterRenderer.Object.<anonymous>.Component.setState (../../node_modules/react/cjs/react.development.js:354:16)
traceviewer-react-components:       at src/components/table-renderer-components.tsx:211:26
```

marcdumais-work:
> It works fine. I get a new warning when running the test with `yarn test` which was not there before. Do you see the same:
> [...]

Yes, I noticed this warning and tried to debug it. It happens in test `Search filter renderer key interactions`, when the search area is selected and the "Enter" key press is simulated on it, causing the `table-renderer-component` to trigger a (mocked) search for the next element. 

ATM that test does not use "act()" blocks. I've tried adding some, for each line that triggers a state update, but I still got the warning. Looking at the code involved, I think this is because this code-path, though it has a synchronous interface, it ends-up executing the update asynchronously \[1\]. So, I think the async part of this flow can cause the React component state update,  to happen pass the "Act()" block, rendering it useless. 

At least for now, I think that's ok - we might face timing issues if we try adding more robust tests in this area, and that would be a good opportunity to potentially re-consider how this flow is implemented `table-renderer-components.tsx`.


\[1\]: In the tests, when there's a simulated "Enter" keystroke, "table-renderer-components.tsx", sync method `onKeyDownEvent()` is called, and in turn it calls method `this.runSearch()` which does some async operations.

bhufmann:
> > It works fine. I get a new warning when running the test with `yarn test` which was not there before. Do you see the same:
> > [...]
> 
> Yes, I noticed this warning and tried to debug it. It happens in test `Search filter renderer key interactions`, when the search area is selected and the "Enter" key press is simulated on it, causing the `table-renderer-component` to trigger a (mocked) search for the next element.
> 
> ATM that test does not use "act()" blocks. I've tried adding some, for each line that triggers a state update, but I still got the warning. Looking at the code involved, I think this is because this code-path, though it has a synchronous interface, it ends-up executing the update asynchronously [1]. So, I think the async part of this flow can cause the React component state update, to happen pass the "Act()" block, rendering it useless.
> 
> At least for now, I think that's ok - we might face timing issues if we try adding more robust tests in this area, and that would be a good opportunity to potentially re-consider how this flow is implemented `table-renderer-components.tsx`.
> 
> [1]: In the tests, when there's a simulated "Enter" keystroke, "table-renderer-components.tsx", sync method `onKeyDownEvent()` is called, and in turn it calls method `this.runSearch()` which does some async operations.

Thanks for the explanation. I think we can merge it as is.

bhufmann (approved):
Thank you for the contribution.

marcdumais-work:
Thanks for the review!

PR #1119, opened 2024-09-20
Create a new 'Open in Trace Viewer' command, add it to navigator context menu

marcdumais-work:
Starting around Theia v1.50.0, the "open-with" service no longer permitted to open folders from the navigator context menu (only worked for files). It was possible to make it work again, but there were undesirable side-effects. So instead, we decided to register our own command, "Open with Trace Viewer", and add it to the navigator's context menu at the first level. This is very similar to what we have on the `Trace Viewer for VSCode` extension. 

Also updated the README to reflect how a trace can now be be opened from the File Explorer, and recorded a new .gif showing it (will only be visible from the README after the PR is merged).

Note: this PR is currently on top of the upgrade to Theia 1.52.0, that has its own PR. We can merge that one first, or I can re-organize the commits if needed.


theia-extensions/viewer-prototype/src/browser/trace-explorer/trace-explorer-commands.ts
```diff
@@ -17,4 +17,9 @@ export namespace TraceExplorerCommands {
         id: 'trace-explorer:remove-trace',
         label: 'Remove Trace'
     };
+
+    export const OPEN_WITH_TRACE_VIEWER: Command = {
+        id: 'trace-explorer:open-with-trace-viewer',
+        label: 'Open in Trace Viewer'
```
bhufmann, line 23: Should it be called `Open with Trace Viewer`? The vscode-trace-extension is using thas name.

marcdumais-work: First, I am ok going with one or the other. For the PR, I picked the alternative because it seems to make subtly more sense to me, though either one works:
- `open with trace viewer`: seems to imply a choice, that there may be several options to open something with. This made total sense when we were using the "open-with" Theia service, where it was designed to do exactly that - offer options to open a file/folder with.
- `open in trace viewer`: subtly focuses more on what the file will be opened in, the trace viewer. It's _the_ option if you want to open a trace file or folder. 
See what I mean? 

To be frank, what made me even consider this alternative is the naming for opening a file/folder in the integrated terminal, that uses "open in..." :

image

WDYT?

marcdumais-work: About the vscode extension, if we go with "open in" here, I can open a "good first issue" about it in the other repo, to have them use the same.

bhufmann: Both makes sense. For file traces, that would probably make more sense to do `open with` because they could be open by plain text editor as well. For directory traces, there is no equivalent action to open a directory, so `open in` would make fore sense there.

marcdumais-work: > For file traces, that would probably make more sense to do open with because they could be open by plain text editor as well.

I meant when using "open in trace viewer", there is a connotation that there is no alternative trace viewer, a file or folder can be opened _in_. There might be other options to open it _with_. Too subtle to matter?

marcdumais-work: I have changed back to "open with".

README.md
```diff
@@ -466,7 +466,7 @@ The code in this repository is licensed under `MIT` (see root `LICENSE`), except
 [image-icon]: https://raw.githubusercontent.com/eclipse-cdt-cloud/theia-trace-extension/master/doc/images/theia-trace-extension-icon.png
 [image-open-browser]: https://raw.githubusercontent.com/eclipse-cdt-cloud/theia-trace-extension/master/doc/images/theia-trace-extension-open-browser.png
 [image-open-view]: https://raw.githubusercontent.com/eclipse-cdt-cloud/theia-trace-extension/master/doc/images/theia-trace-extension-open-view.gif
-[image-open-with]: https://raw.githubusercontent.com/eclipse-cdt-cloud/theia-trace-extension/master/doc/images/theia-trace-extension-open-with-trace-viewer.gif
+[image-open-with-trace-viewer]: https://raw.githubusercontent.com/eclipse-cdt-cloud/theia-trace-extension/master/doc/images/theia-trace-extension-open-with-trace-viewer.gif
```
bhufmann, line 469: Github caches images. So, even if you change the source file of the image it will still show the old file on the webpage. So, what I usually do is to have a version number on the file, include that new file and delete the old one with the old name (see below for an example).

theia-extensions/viewer-prototype/src/browser/trace-explorer/trace-explorer-commands.ts
```diff
@@ -17,4 +17,9 @@ export namespace TraceExplorerCommands {
         id: 'trace-explorer:remove-trace',
         label: 'Remove Trace'
     };
+
+    export const OPEN_WITH_TRACE_VIEWER: Command = {
+        id: 'trace-explorer:open-with-trace-viewer',
+        label: 'Open with Trace Viewer'
+    };
```
bhufmann, line 24: Not sure where to add the comment. Could you please change the priority of the menu entry of that command so that it's lower in the list, preferably last. We did the same change in the [vscode-trace-extension PR 249](https://github.com/eclipse-cdt-cloud/vscode-trace-extension/pull/249) after receiving user feedback that it's too intrusive (i.e. too high up with the common open commands). See image. Not sure how to do that in Theia but it should be possible. 

BTW, I see now why you suggest to change it to `Open in Trace Viewer`. It would be consistent to the `Open in Terminal` command. So, I would be ok to use that, but we should also change it in the vscode-trace-extension.

image

bhufmann (changes requested):
It works well in both browser and electron app. Small comments though.

README.md
marcdumais-work: I've seen your comment below, about being ok with "open in" - I'll restore that version and it will fix this comment, since the .gif file is named differently.

theia-extensions/viewer-prototype/src/browser/trace-explorer/trace-explorer-commands.ts
marcdumais-work: > change the priority of the menu entry of that command so that it's lower in the list, preferably last

easily done

> [use "open in trace viewer"] we should also change it in the vscode-trace-extension

Created an issue, that we can schedule as we see fit: https://github.com/eclipse-cdt-cloud/vscode-trace-extension/issues/275

marcdumais-work: I pushed a new version that should address the comments. The context menu entry should be the last item in the "Navigation" section (top part) of the menu. We can't control the order of new entries that might get added by others, but with the vanilla entries, added by Theia, we should be consistently at the end of the section.

marcdumais-work: One question I have: are you bugged by the name here, the constant being named "OPEN_WITH..." but presenting the command as "open in..."?:

image

marcdumais-work: If so, I can align everything to use "open in"

bhufmann: It's now or never because after we release it the command could be triggered by a third-party extension.

marcdumais-work: Good point - let's do it then

marcdumais-work: done, new version pushed

bhufmann (approved):
Looks good to me. Tested both browser and electron. Thanks!

PR #1120, opened 2024-09-23
Hide empty rows when filter is applied

PatrickTasse:
Hide rows that have no non-null state in their model, when a filter is applied. Gaps between states in the row model should be make rows visible when strategy is SAMPLED, but not when it is DEEP.

Keep an up-to-date list of empty nodes after fetching row models from the server.

Add emptyNodes and hideEmptyNodes to the properties of EntryTree, Tree and Table. Filter the empty nodes when rendering the Table.

Remove the updateSearchFilters callback when adding or removing a filter as the state change will handle it already.

Removed unused dataRows in TimegraphOuputState

### DIFF
--- a/packages/react-components/package.json
+++ b/packages/react-components/package.json
@@ -37,7 +37,7 @@
         "react-virtualized": "^9.21.0",
         "semantic-ui-css": "^2.4.1",
         "semantic-ui-react": "^0.86.0",
-        "timeline-chart": "^0.3.4",
+        "timeline-chart": "^0.4.0",
         "traceviewer-base": "0.2.7",
         "tsp-typescript-client": "^0.4.3"
     },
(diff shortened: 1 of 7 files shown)


PatrickTasse:
Requires https://github.com/eclipse-cdt-cloud/timeline-chart/pull/296

bhufmann (changes requested):
It works well. I found one issue though. The filter expression (hide or search) filter is not applied anymore after removing or adding annotations (e.g. lost events).

PatrickTasse:
> It works well. I found one issue though. The filter expression (hide or search) filter is not applied anymore after removing or adding annotations (e.g. lost events).

Fixed with https://github.com/eclipse-cdt-cloud/theia-trace-extension/pull/1123

bhufmann (approved):
Looks good to me. I tested it with small and large traces. Thanks! 

I rebased it to current master with before testing. Please do so as well.

PR #1123, opened 2024-09-24
Fix updateChart() missing filter expressions map

PatrickTasse:
When a filter is applied, updateChart() should not be called without the
filter expression map.

Replace method updateSearchFilters(), which computes the filter
expression map and calls updateChart with the map if it applies, with
method filterExpressionsMap() which only computes the map. This method
is now always called to compute the parameter to be passed in
updateChart().

Remove the updateSearchFilters callback when adding or removing a filter
as the state change will handle it already.

Method _debouncedUpdateSearch() is removed, as _debouncedUpdateChart()
can always be used instead.


bhufmann (approved):
Looks good to me. Fixed the issue of not applying filters when removing/adding annotations.

PR #1128, opened 2024-10-10
Use filter table tree for Available Views view implementation

### LINKED ISSUE #1113: req: Add searchable View List/Trace List
Trace Compass's Trace Servers are getting more and more views. This is a positive, but now we must manage the views in a better way than just a simple list.

It would be good to have a search to find the type of view that we want to display.

This also applies to traces. I have 200+ traces in my trace server, so having a way to find which one I was viewing would speed up my workflows.

bhufmann:
- Use FilterTree
- Add tooltips to TreeNode and TableCell
- Enable sorting and filter
- Description of view is part of the tooltip instead of being printed below view name

fixes #1113

vscode-new-available-view2

### DIFF
--- a/packages/react-components/src/trace-explorer/trace-explorer-views-widget.tsx
+++ b/packages/react-components/src/trace-explorer/trace-explorer-views-widget.tsx
@@ -5,7 +5,9 @@ import { OutputDescriptor } from 'tsp-typescript-client/lib/models/output-descri
 import { Experiment } from 'tsp-typescript-client/lib/models/experiment';
 import { ITspClientProvider } from 'traceviewer-base/lib/tsp-client-provider';
 import { ExperimentManager } from 'traceviewer-base/lib/experiment-manager';
-import { AvailableViewsComponent } from '../components/utils/available-views-component';
+import { FilterTree } from '../components/utils/filter-tree/tree';
+import { TreeNode } from '../components/utils/filter-tree/tree-node';
+import { getAllExpandedNodeIds } from '../components/utils/filter-tree/utils';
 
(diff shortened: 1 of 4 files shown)


packages/react-components/src/trace-explorer/trace-explorer-views-widget.tsx
```diff
@@ -44,54 +54,77 @@ export class ReactAvailableViewsWidget extends React.Component<ReactAvailableVie
...
-        this.doHandleOutputClicked(outputDescriptor);
-    protected handleContextMenuEvent = (
-        e: React.MouseEvent<HTMLDivElement>,
-        output: OutputDescriptor | undefined
-    ): void => this.doHandleContextMenuEvent(e, output);
+    private renderOutputs() {
+        if (this.state.treeNodes) {
+            return (
+                <div className="scrollable avail-views-table">
+                    <FilterTree
```
bhufmann, line 66: should we move this into AvailableViewComponent? Then the Overview menu item to select the xy output would have it as well.

PatrickTasse (changes requested):
Now matter how wide the Available Views panel is resized, its scrollable content is just a little bit wider than the available width, so a horizontal scrollbar appears unnecessarily with a very small scrollable range. There might be some margin width unnaccounted for somewhere.

bhufmann:
> Now matter how wide the Available Views panel is resized, its scrollable content is just a little bit wider than the available width, so a horizontal scrollbar appears unnecessarily with a very small scrollable range. There might be some margin width unnaccounted for somewhere.

This should be fixed in the updated version

PatrickTasse (approved):

PR #1130, opened 2024-10-11
Add step to install dependencies on Ubuntu OS to publish workflows

bhufmann:
This makes sure that all necessary dependencies are available during publish.

Bernd Hufmann <bernd.hufmann@ericsson.com>


marcdumais-work (approved):
LGTM - thanks Bernd!

PR #1131, opened 2024-10-17
Fix time graph filtering and collapsing

### LINKED ISSUE #1132: Time graph filtering and collapsing not working properly.
When a filter  is applied, only the matching root elements (below the trace root) are visible. All children of a non-matching parent are ignored.

When a non-matching node is filtered-out, it should be replaced by its matching children, recursively. The children of collapsed nodes should not be considered.

When updating the arrows layer, it is not considering filtered-out nodes so the arrows can start or end at the wrong node position or in blank space.

Rows that change y-position because preceding node is filtered-out or collapsed are remaining at the previous position. They should be updated. If possible avoid refreshing the row from the server if the current model is still valid. If no row needs to be fetched from the server then arrows may still need to be refreshed.

When a filter is applied and a search string is entered, all rows are re-filtered. This should only be done when a filter is added or removed, not when search string is modified.

PatrickTasse:
Update the filtering of empty nodes to include visible nodes even if any of their parent is filtered-out. Recursively replace filtered-out nodes by their visible children, if any, but only if the filtered-out parent node is not collapsed.

In TimeGraphOutputComponent update the arrows layer after empty nodes have been determined, and filter out empty nodes from the rowIds passed to the arrows layer.

Fixes #1132

### DIFF
--- a/packages/react-components/src/components/utils/filter-tree/table-body.tsx
+++ b/packages/react-components/src/components/utils/filter-tree/table-body.tsx
@@ -4,8 +4,8 @@ import { TableRow } from './table-row';
 
 interface TableBodyProps {
     nodes: TreeNode[];
-    selectedNode?: number;
-    multiSelectedNodes?: number[];
+    selectedRow?: number;
+    multiSelectedRows?: number[];
     collapsedNodes: number[];
     isCheckable: boolean;
     isClosable: boolean;
@@ -34,7 +34,6 @@ export class TableBody extends React.Component<TableBodyProps> {
         if (!this.props.nodes) {
             return undefined;
         }
-
         return <tbody>{this.renderRows()}</tbody>;
     }
 }
--- a/packages/react-components/package.json
+++ b/packages/react-components/package.json
@@ -37,7 +37,7 @@
         "react-virtualized": "^9.21.0",
         "semantic-ui-css": "^2.4.1",
         "semantic-ui-react": "^0.86.0",
-        "timeline-chart": "^0.4.0",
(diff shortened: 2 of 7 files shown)


bhufmann (changes requested):
Please fix linting error.  Meanwhile I'll review the change

bhufmann:
I noticed some issues when selection rows when filtering is ongoing and after all rows are filtered out.

When filtering is ongoing, and I click on the tree, it seems that the tree is scrolling to a different location:

timeline-chart-filter-selection-bug2


When filtering is finished, the selection is not synced correctly from tree to graph. See video when selecting with left mouse button:
timeline-chart-filter-selection-bug

bhufmann (changes requested):
I tested the latest updates and the selection is working now. The original bug for collapsed nodes is fixed as well.

Please fix the linter error then I can approved it.

packages/react-components/src/components/utils/filter-tree/table-body.tsx
```diff
@@ -34,7 +34,7 @@ export class TableBody extends React.Component<TableBodyProps> {
         if (!this.props.nodes) {
             return undefined;
         }
-
+        console.log('TableBody.render() selectedRow:'+this.props.selectedRow+' multiSelectedRows:'+this.props.multiSelectedRows);
```
bhufmann, line 37: remove print-out

bhufmann (changes requested):

bhufmann (approved):
LGTM. Thanks!

PR #1139, opened 2024-11-01
Fix filter tree filter text box width

PatrickTasse:
Use box-sizing: border-box setting so that the provided width includes the filter text box padding and border.

Use the reference element's right position as provided width so that it can include the reference element's left margin.

### DIFF
--- a/packages/react-components/src/components/utils/filter-tree/filter.tsx
+++ b/packages/react-components/src/components/utils/filter-tree/filter.tsx
@@ -19,7 +19,7 @@ export class Filter extends React.Component<FilterProps, FilterState> {
         };
         this.resizeObserver = new ResizeObserver(entries => {
             this.setState({
-                width: entries[0].contentRect.width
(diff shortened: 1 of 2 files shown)


packages/react-components/src/components/utils/filter-tree/filter.tsx
```diff
@@ -19,7 +19,7 @@ export class Filter extends React.Component<FilterProps, FilterState> {
         };
         this.resizeObserver = new ResizeObserver(entries => {
             this.setState({
-                width: entries[0].contentRect.width
+                width: entries[0].contentRect.right
```
bhufmann, line 22: Is this really needed? It's seems enough with just the css change below. WDYT?

PatrickTasse: Then the filter is not as wide as it should be. The width of the filter text box should not be influenced by the left margin of the tree column.

PR #1141, opened 2024-11-11
Adopt tsp-typescript-client version 0.5.1 and pin inversify to 6.0.3

bhufmann:
- Use resolution to pin inversify to v6.0.3 to avoid docker build errors
  The docker example doesn't have yarn.lock file and hence the builds starts fresh. The new inversify version v6.1.x cause a build error.
- Adopt tsp-typescript-client version 0.5.1

### DIFF
--- a/examples/docker/example-package.json
+++ b/examples/docker/example-package.json
@@ -14,6 +14,7 @@
       }
     }
   },
+  
   "dependencies": {
     "@emotion/react": "^11.10.5",
     "@emotion/styled": "^11.10.5",
@@ -35,6 +36,7 @@
     "node": ">=16 <19"
   },
   "resolutions": {
-    "msgpackr": "^1.10.1"
+    "msgpackr": "^1.10.1",
+    "inversify": "6.0.3" 
   }
 }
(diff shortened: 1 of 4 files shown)


bhufmann:
Licences check fails. Ticket has been submitted for the new version: https://gitlab.eclipse.org/eclipsefdn/emo-team/iplab/-/issues/17159

marcdumais-work:
I think the Docker CI failure is most likely unrelated to this PR and probably caused by pulling newer versions of some dependencies. This happens because that example has no lockfile.

PatrickTasse (approved):

bhufmann:
> I think the Docker CI failure is most likely unrelated to this PR and probably caused by pulling newer versions of some dependencies. This happens because that example has no lockfile.

I added a commit to pin the inversify version to avoid the error.

examples/docker/example-package.json
```diff
@@ -14,6 +14,7 @@
       }
     }
   },
+
```
marcdumais-work, line 17: added newline on purpose?

PR #1143, opened 2024-11-12
Import style in trace-explorer-views-widget.tsx

bhufmann:
With this it's guaranteed that the styles are loaded when using this react component outside this repository (e.g. in vscode-trace-extension)


MatthewKhouzam (approved):
Yes, it's harmless to bring it here and safer!

PR #1146, opened 2024-11-12
Change mouse cursor to pointer when hovering over an output

bhufmann:
The implementation of the available view has recently change and the cursor was not updated when hovering over an output (view). This commit changes that.

image


marcdumais-work (approved):
LGTM, thanks @bhufmann 
Good catch! Nice, minimal fix too - I confirmed it works.

PR #1147, opened 2024-11-13
Remove gridlines in Available Views view

bhufmann:
The gridlines are not needed because there are visual effects like mouse pointer and background change when hovering over a line, which provide visual clues for the user.


marcdumais-work (approved):
LGTM - no more gridlines on that view!

PR #1152, opened 2024-11-21
Show column in TimegraphOutputComponent if back-end provides them

bhufmann:
No header is drawn with the commit.

Fixes #1149

Signed-off-by: Bernd Hufmann <bernd.hufmann@ericsson.com>

PR #1154, opened 2024-11-27
Improve display of vertical tree gridlines in timegraph views

bhufmann:
v0.4.2 introduced vertical gridlines for the trees of timegraph views (see #1152) . However, 2 vertical lines were drawn so that they appeared thicker. This commit fixes that.

Besides, the vertical line on the right of the last column are not drawn anymore. For that, the hideFiller flag needed to be added to the EntryTree props so that that flag can be set to true, propagated and hence the filler column can be hidden.

Here how it looks like with this PR:

image


MatthewKhouzam (approved):
Looks great, thanks!

marcdumais-work (approved):
LGTM!

PR #1168, opened 2025-01-30
Tooltip refactor

williamsyang-work:
# Refactor Tooltip Components

## Overview
This PR modernizes our tooltip implementation by replacing the `react-tooltip` dependency with a custom tooltip component and introducing a more React-idiomatic state management approach.

## Key Changes
- Removes dependency on `react-tooltip` in favor of a custom implementation
- Replaces direct DOM manipulation via refs with proper React state management
- Introduces an abstract component pattern for consistent tooltip behavior
- Removes the separate XY tooltip implementation in favor of a unified approach
- Improves type safety and component reusability

### New Tooltip Implementation
The new tooltip component uses React's built-in state management and Portal API, providing better control over positioning and animations:

```typescript
export const TooltipComponent = ({ content = '⏳', visible = false, fadeTransition = 500 }: TooltipProps) => {
    const [state, setState] = useState<TooltipState>({
        content,
        visible,
        zIndex: visible ? 99999 : -99999
    });

    // Clean React-based positioning logic
    const calculateAndSetPosition = () => {
        if (!tooltipRef.current) return;
        
        const viewportWidth = window.innerWidth;
        const viewportHeight = window.innerHeight;
        const tooltipRect = tooltipRef.current.getBoundingClientRect();
        
        // Position calculation logic
    };
}
```

Compared to the old implementation which relied on external library and callbacks:

```typescript
<ReactTooltip
    className="react-tooltip"
    id="tooltip-component"
    effect="float"
    type="info"
    place="bottom"
    html={true}
    delayShow={500}
    delayUpdate={500}
    afterShow={() => {
        if (this.state.content === undefined) {
            this.fetchContent(this.state.element);
        }
    }}
/>
```

### Abstract Component Pattern
Enhanced the existing `AbstractOutputComponent` with tooltip management capabilities:

```typescript
interface AbstractOutputState {
    tooltipContent?: React.ReactNode;
    tooltipVisible?: boolean;
}

export abstract class AbstractOutputComponent<P extends AbstractOutputProps, S extends AbstractOutputState> extends React.Component<P, S> {
    render(): JSX.Element {
        return (
            <div onMouseLeave={() => this.closeTooltip()}>
                {/* Component content */}
                <TooltipComponent 
                    content={this.state.tooltipContent} 
                    visible={this.state.tooltipVisible} 
                />
            </div>
        );
    }

    protected setTooltipContent(content: React.ReactNode): void {
        this.setState({ tooltipContent: content, tooltipVisible: true });
    }

    protected closeTooltip(): void {
        this.setState({ tooltipVisible: false });
    }
}
```

This replaces the old callback-based approach:

```typescript
setElement<T>(element: T, func?: (element: T) => MaybePromise<{ [key: string]: string } | undefined>): void {
    if (element !== this.state.element && this.state.element) {
        if (this.state.content) {
            if (this.timerId === undefined) {
                this.timerId = setTimeout(() => {
                    if (this.state.element !== element) {
                        ReactTooltip.hide();
                        this.setState({ content: undefined });
                    }
                }, 500);
            }
        }
    }
    this.setState({ element, func });
}
```

### Removed Complexity
- Eliminated callback-based tooltip content generation
- Removed manual DOM event handling for positioning
- Simplified tooltip state management

## Migration Examples
Here are examples of how components are being updated to use the new tooltip pattern:

### XY Chart Tooltip
```typescript
// ...
if (points.length > 0) {
    this.setTooltipContent(this.generateXYComponentTooltip(tooltipData));
} else {
    this.closeTooltip();
}
// ...

private generateXYComponentTooltip = (tooltipData: TooltipData) => (
    <>
        <p style={{ margin: '0 0 5px 0' }}>{tooltipData?.title}</p>
        <ul style={{ padding: '0' }}>
            {tooltipData.dataPoints?.map(
                (point: { color: string; background: string; label: string; value: string }, i: number) => (
                    <li key={i} style={{ listStyle: 'none', display: 'flex', marginBottom: 5 }}>
                        <div
                            style={{
                                height: '10px',
                                width: '10px',
                                margin: 'auto 0',
                                border: 'solid thin',
                                borderColor: point.color,
                                backgroundColor: point.background
                            }}
                        ></div>
                        <span style={{ marginLeft: '5px' }}>
                            {point.label} {point.value}
                        </span>
                    </li>
                )
            )}
        </ul>
        {tooltipData.zeros > 0 && (
            <p style={{ marginBottom: 0 }}>
                {tooltipData.zeros} other{tooltipData.zeros > 1 ? 's' : ''}: 0
            </p>
        )}
    </>
);
```

### Table Tooltip
```typescript
// ...
const tooltipContent = this.generateTooltipTable(tooltipObject);
this.setTooltipContent(tooltipContent);
// ...

private generateTooltipTable(tooltip: { [key: string]: string }) {
    return (
        <table>
            {Object.keys(tooltip).map(key => (
                <tr key={key}>
                    <td style={{ textAlign: 'left' }}> {key} </td>
                    <td style={{ textAlign: 'left' }}> {tooltip[key]} </td>
                </tr>
            ))}
        </table>
    );
}
```

## Testing Instructions
- Verify tooltip positioning works correctly near viewport edges
- Check fade animations on show/hide
- Ensure tooltips close properly on scroll/mouse leave
- Test with different content types (text, elements, async content)
- Compatability with `vscode-trace-extension`
  - You will run into this error linking the packages.  https://github.com/eclipse-cdt-cloud/vscode-trace-extension/issues/175

## Future Considerations
- Potential for additional customization options (colors, sizes, animations)
- Performance optimizations for frequent tooltip updates
- Accessibility improvements


bhufmann:
@williamsyang-work thanks for this update. I see that it's still a draft PR. Let us know when it's ready for review.

williamsyang-work:
> @williamsyang-work thanks for this update. I see that it's still a draft PR. Let us know when it's ready for review.

@bhufmann I just cleaned up the PR.  It's ready for review!

packages/react-components/src/components/tooltip-component.tsx
```diff
@@ -1,121 +1,194 @@
...
+        return () => {
+            window.removeEventListener('scroll', handleScroll, options);
+            document.removeEventListener('scroll', handleScroll, options);
+        };
+    }, []);
 
-    private getContent() {
-        if (this.state.content) {
-            return this.state.content;
+    // Handle content and visibility changes
```
PatrickTasse, line 128: I am not sure where this happens or can be fixed, but when you hover over the tooltip and do a text range selection, some range selection remains when you close and reopen the tooltip, even when the text content changes.

PatrickTasse (changes requested):

packages/react-components/src/components/timegraph-output-component.tsx
```diff
@@ -816,6 +822,19 @@ export class TimegraphOutputComponent extends AbstractTreeOutputComponent<Timegr
         return fullTooltip;
     }
 
+    private generateTooltipTable(tooltip: { [key: string]: string }) {
+        return (
+            <table>
+                {Object.keys(tooltip).map(key => (
+                    <tr key={key}>
+                        <td style={{ textAlign: 'left' }}> {key} </td>
+                        <td style={{ textAlign: 'left' }}> {tooltip[key]} </td>
```
PatrickTasse, line 831: Seems odd formatting to add spaces around the {...} ? Does it get stripped in the html?

packages/react-components/src/components/tooltip-component.tsx
```diff
@@ -1,121 +1,194 @@
...
-    content?: string;
+interface TooltipState {
+    content: React.ReactNode;
+    visible: boolean;
+    zIndex: number;
 }
 
-export class TooltipComponent extends React.Component<unknown, TooltipComponentState> {
-    private static readonly HOURGLASS_NOT_DONE = '&#x23f3;';
+export const TooltipComponent = ({ content = '⏳', visible = false, fadeTransition = 500 }: TooltipProps) => {
```
PatrickTasse, line 18: I get error/warning: Missing return type on function.

packages/react-components/src/components/abstract-xy-output-component.tsx
```diff
@@ -751,7 +770,7 @@ export abstract class AbstractXYOutputComponent<
             rightPos = xLocation;
         }
 
-        const tooltipData = {
+        const tooltipData: TooltipData = {
             title: timeLabel,
             dataPoints: points,
             top: topPos,
```
PatrickTasse, line 776: The four positioning fields are not used, so the position calculations above are also not needed, and the x,y function parameters are not needed.

packages/react-components/src/components/abstract-xy-output-component.tsx
```diff
@@ -64,6 +64,25 @@ export type AbstractXYOutputState = AbstractTreeOutputState & {
...
+    value: string;
+}
+
+interface TooltipData {
+    title?: string;
+    dataPoints?: Point[];
+    top?: number;
+    bottom?: number;
+    right?: number;
+    left?: number;
```
PatrickTasse, line 80: The four position fields are not used.

packages/react-components/src/components/abstract-xy-output-component.tsx
```diff
@@ -64,6 +64,25 @@ export type AbstractXYOutputState = AbstractTreeOutputState & {
...
+    opacity?: number;
+    transition?: string;
```
PatrickTasse, line 82: The transition field is not used.

yarn.lock
```diff
@@ -12407,14 +12407,6 @@ react-test-renderer@^18.2.0:
...
 
-react-tooltip@4.2.14:
-  version "4.2.14"
-  resolved "https://registry.yarnpkg.com/react-tooltip/-/react-tooltip-4.2.14.tgz#8e06b5926fdf6672e78d8ccadaa16bef40d131d7"
-  integrity sha512-hS2kAlpjyH5MXL9DaGKsdmEFCIEuMD2RZXkEJeNjmDe05dHpqj93o5JgpmczAgQFk099+JSsnHUDo7pIOuyMDQ==
-  dependencies:
-    prop-types "^15.7.2"
-    uuid "^7.0.3"
-
 react-tooltip@^4.2.21:
```
PatrickTasse, line 12410: Just checking, is this one still needed?

marcdumais-work: It's a dependency of `@theia/core` - Indeed, I do not see directly used in this repo here, but it might be used in the Theia examples in this repo.

```
$ yarn why react-tooltip
yarn why v1.22.22
[1/4] Why do we have the module "react-tooltip"...?
[2/4] Initialising dependency graph...
[3/4] Finding dependency...
[4/4] Calculating file sizes...
=> Found "react-tooltip@4.5.1"
info Reasons this module exists
   - "_project_#browser-theia-trace-example-app#@theia#core" depends on it
   - Hoisted from "_project_#browser-theia-trace-example-app#@theia#core#react-tooltip"
info Disk size without dependencies: "956KB"
info Disk size with unique dependencies: "1.48MB"
info Disk size with transitive dependencies: "1.6MB"
info Number of shared dependencies: 6
```

packages/react-components/src/components/tooltip-component.tsx
williamsyang-work: I don't know what you mean exactly...  Can you post a GIF?

packages/react-components/src/components/abstract-xy-output-component.tsx
williamsyang-work: Thanks.  I went ahead and removed the function inputs `(x, y)` as well as the following keys from `TooltipData`: `top, bottom, right, left` and `opacity, transition`.

Function calls of `tooltip` have also been fixed in `xy-output-component` and `xy-output-overview-component`.

packages/react-components/src/components/tooltip-component.tsx
williamsyang-work: Added typing for the function.

yarn.lock
williamsyang-work: I did check this out before submitting this PR.  I made sure react-tooltip was removed from all of the codebase as well as all `package.json` files.  I came to the same conclusion as @marcdumais-work, that `@theia` was using it somewhere.

williamsyang-work:
@PatrickTasse sorry for the delay.  I addressed all of your comments except the first one, which I need some clarification on.  Thank you!

yarn.lock
marcdumais-work: In any case, this dependency will not be part of the published npm packages (since they do not depend on it), so we should be good.

packages/react-components/src/components/timegraph-output-component.tsx
PatrickTasse: I mean around {key} and {tooltip[key]} inside the <td></td> html tag.

williamsyang-work:
After some thinking I decided to refactor this into a React Class Component.  It will be easier to maintain.

packages/react-components/src/components/tooltip-component.tsx
PatrickTasse: I hover on a state on the kworker row, then mouse over the tooltip and select the End time value:
tooltip1

Then I move my mouse and hover over a state on the swapper row, and in the tooltip the End time value (a different string) is already selected:
tooltip2

This example is neat, but when the tooltip content is very different, then the persisted selection can be anything, starting and ending in the middle of words from different properties.

williamsyang-work: This has been addressed!

PatrickTasse (changes requested):
Please check and fix Lint build errors/warnings, some unused vars and formatting (yarn format:write to auto-fix)

packages/react-components/src/components/tooltip-component.tsx
```diff
@@ -151,6 +151,7 @@ export class TooltipComponent extends React.Component<TooltipProps, TooltipState
      * 1. If mouse is over the tooltip, cancels the update to prevent hiding while user is interacting
      * 2. When hiding the tooltip, maintains content and z-index during fade-out animation
      * 3. When showing the tooltip, updates position and immediately sets z-index to visible
+     * 4. When changing content, remove user slected text (fixes a weird bug)
```
PatrickTasse, line 154: selected

packages/react-components/src/components/tooltip-component.tsx
```diff
@@ -159,24 +160,32 @@ export class TooltipComponent extends React.Component<TooltipProps, TooltipState
...
             const weAreHidingTooltip = prevState.visible === true && newState.visible === false;
+            const textIsChanging = prevState.content !== newState.content;
 
             if (this.isMouseOver) {
                 this.debouncedUpdate.cancel();
                 return prevState;
-            } else if (weAreHidingTooltip) {
+            }
+            
+            if(textIsChanging) {
```
PatrickTasse, line 170: space before (

hriday-panchasara (approved):
LGTM, thanks!

marcdumais-work:
@williamsyang-work I am also having a look at this contribution. It should not take too long.

marcdumais-work:
Trying this PR in the vscode extension (*) - I see some errors/warnings that I do not see testing using the theia trace extension (in the local example application), or when I instead use the "master" version of `traceviewer-react-components` in :vscode trac extension. i.e. this seems new. Not sure how "bad" these are, but probably worth having a look-into?

When I trigger the first tooltip on a newly opened trace:
image

I get this additional one when triggering the first tooltip e.g. in the Thread Status view:
image

(*): WIP PR that facilitates using locally-build Trace components in the vscode trace extension build. https://github.com/eclipse-cdt-cloud/vscode-trace-extension/pull/310

marcdumais-work:
> Compatability with vscode-trace-extension
You will run into this error linking the packages. https://github.com/eclipse-cdt-cloud/vscode-trace-extension/issues/175

William, if you want I can help you get a local setup working, based on (WIP/Draft) https://github.com/eclipse-cdt-cloud/vscode-trace-extension/pull/310

PatrickTasse (changes requested):
I also see the two errors found by Marc

williamsyang-work:
This 

> Trying this PR in the vscode extension (*) - I see some errors/warnings ~that I do not see testing using the theia trace extension (in the local example application), or~ when I ~instead~ use the "master" version of `traceviewer-react-components` in :vscode trac extension. i.e. this seems new. Not sure how "bad" these are, but probably worth having a look-into?
> 
> _update: correction - I do see the two errors in the Theia example app - not sure why I did not before. So probably no need to use the vscode extesion to reproduce_
> 
> When I trigger the first tooltip on a newly opened trace: image
> 
> I get this additional one when triggering the first tooltip e.g. in the Thread Status view: image
> 
> (*): WIP PR that facilitates using locally-build Trace components in the vscode trace extension build. [eclipse-cdt-cloud/vscode-trace-extension#310](https://github.com/eclipse-cdt-cloud/vscode-trace-extension/pull/310)

This error was caused because `this.calculateAndSetPosition` was being called inside of a `this.setState` call in the `this.updateState` method.

I solved this by pulling the logic out of `this.setState` and  comparing `updatedState` to current state instead of the `prevState`.

The old code has been removed but I have it here

```ts
private updateState = (newState: Partial<TooltipState>): void => {
       // All of the logic is inside this setState call.
        this.setState(prevState => {
            const updatedState = { ...prevState, ...newState };

            const weAreHidingTooltip = prevState.visible === true && newState.visible === false;
            const textIsChanging = prevState.content !== newState.content;

            if (this.isMouseOver) {
                this.debouncedUpdate.cancel();
                return prevState;
            }

            if (textIsChanging) {
                this.clearSelection();
            }

            if (weAreHidingTooltip) {
                // Don't update the content
                updatedState.content = prevState.content;
                // Keep z-index high during fade out
                updatedState.zIndex = prevState.zIndex;
                return updatedState;
            }

            this.calculateAndSetPosition();
            // Update z-index immediately when showing
            if (newState.visible) {
                updatedState.zIndex = 99999;
            }
            return updatedState;
        });
    };
```

New code
```ts
private updateState = (newState: Partial<TooltipState>): void => {
        const currentState = this.state;
        const updatedState = { ...currentState, ...newState };

        const weAreHidingTooltip = currentState.visible === true && newState.visible === false;
        const textIsChanging = currentState.content !== newState.content;

        if (this.isMouseOver) {
            this.debouncedUpdate.cancel();
            return;
        }

        if (weAreHidingTooltip) {
            // Don't update the content
            updatedState.content = currentState.content;
            // Keep z-index high during fade out
            updatedState.zIndex = currentState.zIndex;
            this.setState(updatedState);
            return;
        }

        if (textIsChanging) {
            this.clearSelection();
        }

        // Update position before showing
        const { top, left } = this.calculateAndSetPosition();
        // Update the position here so it doesn't move back
        updatedState.top = top;
        updatedState.left = left;

        // Update z-index immediately when showing
        if (newState.visible) {
            updatedState.zIndex = 99999;
        }

        this.setState(updatedState);
    };
```

marcdumais-work:
Both error are gone - thanks @williamsyang-work

marcdumais-work (approved):
LGTM - thanks for this nice contribution @williamsyang-work !

PatrickTasse (approved):

PR #1169, opened 2025-02-05
Add filter to Opened Traces widget

PatrickTasse:
### What it does

Add filter container to ReactOpenTracesWidget. Add filter string to the widget's state.

When filter is applied, hide rows that don't match the filter for neither the experiment name nor any of the experiment's traces' name. Rows are hidden by returning 0 as row height and rendering to undefined.

Rename CSS class 'input-filter-tree' to generic 'input-filter-text'.

### How to test

Open a few traces and experiments with multiple traces. Type filter strings that match experiment names, trace names, both or none. Repeat with a selected experiment.

### Follow-ups

N/A

### Review checklist

### DIFF
--- a/packages/react-components/src/components/utils/filter-tree/__tests__/__snapshots__/entry-tree.test.tsx.snap
+++ b/packages/react-components/src/components/utils/filter-tree/__tests__/__snapshots__/entry-tree.test.tsx.snap
@@ -1022,7 +1022,7 @@ Array [
       id="input-filter-icon"
     />
     <input
-      id="input-filter-tree"
+      id="input-filter-text"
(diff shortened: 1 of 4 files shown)


bhufmann (approved):
It looks good to me. When upgrading the `vscode-trace-extension` you need to supply the CSS for the filter. One way is to add the following import to : https://github.com/eclipse-cdt-cloud/vscode-trace-extension/blob/master/vscode-trace-webviews/src/trace-explorer/opened-traces/vscode-trace-explorer-opened-traces-widget.tsx: 

`import 'traceviewer-react-components/style/output-components-style.css';`

PR #1170, opened 2025-02-06
Have example apps require at least node 18

marcdumais-work:
### What it does

Modify example applications "engines.node" to require using at least node.js 18 to build. Node 16 has been EoL for a long while now, so this should not be controversial. If someone wants to still use an older node version, they can use a yarn option to ignore this setting, but note that we do not support older node.js versions officially.

### How to test

Verify that CI still passes. Optionally try building locally using node 16 and confirm yarn reports an error about it.

### Follow-ups

N/A

### Review checklist


bhufmann (approved):
Looks good to me. Thanks!

marcdumais-work:
Closing in favor of https://github.com/eclipse-cdt-cloud/theia-trace-extension/pull/1171

PR #1172, opened 2025-02-10
Show timegraph tree column headers

PatrickTasse:
### What it does

Show headers in the timegraph tree if columns are set.

Remove default 'Name' column header and leave columns empty if unset.

Change columns sortable property to false.

Synchronize the top margin of the timegraph content with the timegraph tree column header height. Change its and its parent's heights to 'auto'. Reduce the timegraph chart container height by the value of the top margin.

Make the timegraph tree column header sticky and adjust its height including borders and margins to be equal to the table rows height.

### How to test

Open timegraph output with columns (e.g. Thread Status) and without columns (e.g. Resources Status). Check scrolling and resizing.

### Follow-ups

Consider enabling sortable columns and showing the column header tooltip, when those are applicable in the ColumnHeader model.

### Review checklist


packages/react-components/src/components/timegraph-output-component.tsx
```diff
@@ -704,7 +717,7 @@ export class TimegraphOutputComponent extends AbstractTreeOutputComponent<Timegr
                         ev.preventDefault();
                         ev.stopPropagation();
                     }}
-                    style={{ height: this.props.style.height }}
+                    style={{ height: 'auto' }}
```
bhufmann, line 720: The height property is ignored now, is this change needed?

bhufmann (commented):
Looks good to me. Thanks

bhufmann (approved):

PR #1176, opened 2025-02-12
Enable sorting of timegraph tree columns

PatrickTasse:
### What it does

Change the sortable property to true for all defined columns in a
timegraph. Add a sortable default column with empty header when no
column headers are defined in the model.

Change the FilterTree onOrderChange callback to include all ordered ids,
including collapsed and hidden nodes.

Add a handler for the onOrderChange callback for a timegraph. When the
callback is handled, reorder the timegraphTree entries according to the
specified order. Update the timegraphTree in the state to trigger
synchronization of the timegraph chart.

### How to test

Open a timegraph with defined columns (e.g. Thread Status) or no defined columns (e.g. Resources Status) and click the column headers to sort the tree. Verify that the timegraph chart is also reordered and that row synchronization still works.

### Follow-ups

Consider adding support to enable or disable the sortable property of each individual column from an updated TSP model.

### Review checklist


bhufmann (approved):
Looks good to me. Thanks for this contribution.

bhufmann (approved):
Looks good to me. Nice improvement. Thanks!

PR #1177, opened 2025-02-13
Enable reset of entry order in entry trees

PatrickTasse:
### What it does

Add onSortReset callback to table, and add onOrderReset callbacks to
tree and entry tree.

Trigger the onSortReset callback from table when the first column is
cycled back to the default sort order.

Store the default ordered ids when fetching the tree in timgraph output
component, xy output component and datatree output component. When the
onOrderReset callback is handled, trigger the ordering of the tree using
the stored default order.

Remove the unused orderedNodes from the state of xy output component and
datatree output component.

### How to test

Cycle through the sort order of the first column in a timegraph, xy chart and datatree.

### Follow-ups

N/A

### Review checklist


bhufmann (dismissed):
Allowing the sorting of first column with the possibility to go back to original sort order provided by the back-end makes sense. Showing the default sorting icon in the head after it's reset indicates that's it's unsorted in the FE.

Thanks for the contribution.

bhufmann (approved):

PR #1189, opened 2025-04-14
Update for Theia community release v1.58.5

marcdumais-work:
### What it does

Update to Theia community release v1.58.5. 

Also, this PR includes using the bundled Electron backend, which is a simple webpack config update. If we later notice issues with this, we can roll-back to use the non-bundled backend in the next community release, which will have the fix above. See commit message for more details about this and why it was necessary.

Note about the webpack config update: as suggested, I have deleted the webpack configs and run "theia build" to have them re-created. Then I re-added the small change we need for the time range context menu. The result is a webpack config that's more aligned with the one Theia generates, for both our example Theia apps.

### How to test

Make sure the trace viewer still works as expected. Also, since the webpack config was updated, also confirm that the "time range" context menu still works as expected, e.g. on views like "Futex Contention analysis - Latency Statistics".

### Follow-ups

N/A

### Review checklist

### DIFF
--- a/examples/browser/package.json
+++ b/examples/browser/package.json
@@ -15,14 +15,14 @@
         }
     },
     "dependencies": {
-        "@theia/core": "1.55.0",
-        "@theia/navigator": "1.55.0",
-        "@theia/preferences": "1.55.0",
-        "@theia/terminal": "1.55.0",
+        "@theia/core": "1.58.5",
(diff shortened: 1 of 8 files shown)


bhufmann (approved):
Looks good to me. Thanks!

marcdumais-work:
thanks for the review!

PR #1194, opened 2025-06-05
update to theia community release 1.61.0

marcdumais-work:
### What it does

Update the Theia dependencies to the latest community release, v1.61.0. Also, add the context to the object passed to `contextMenuRenderer.render()`. This fixes the new compile errors, in 3 places, though I do not think the context object is used in our case (related to the "extract to secondary window" feature, that's not used in the trace viewer.

This change is needed to adapt to the following breaking Theia upstream change:
https://github.com/eclipse-theia/theia/pull/14982

### How to test

Perform some light manual testing of the viewer to confirm all looks ok. 
Related to the added context object when creating context menus, it's a good idea to confirm that the they still appear at the correct place when triggered:
- in the "Opened Traces" panel on the left side, right-click on an opened trace and a context menu should appear with items "open trace", "close trace" and "remove trace". Make sure it appears at the correct place vs the trace you picked to trigger the menu
- Setup some "marker set" and trigger the two related context menus:
   - the one that serves to enable a set, including any set you added and "none"
   - Once a set is enabled, the one that permits to enable/disable contained elements or "select all"

### Follow-ups

N/A

### Review checklist


bhufmann (approved):
Looks good to me. Thanks!

marcdumais-work:
Thanks for the review! 

I will not merge yet, because of the 2 dependencies failing the licence check, at least until we have a chance to discuss. Both are test dependencies and not present at runtime, BTW.

marcdumais-work:
Thanks for the review!

PR #1200, opened 2025-06-12
Update for Theia 1.61.1 and later (from 1.61.0)

marcdumais-work:
### What it does

Because of how Theia resolves its dependencies, Theia apps cannot successfully include more than one version of Theia extensions (e.g. @theia/core), even if they are compatible.

Since a new butfix Theia release happened after we updated the Theia extension in this repo to use Theia 1.61.0, we need to follow. This time I took the precaution of requesting a version range that will include potential future 1.61.x bugfix Theia releases, using "~". e.g.:

"@theia/core": "~1.61.1"

There's a small risk that a future Theia v1.61.x might break for
 us, but OTOH if we do not do this, the theia-traceviewer extension
will not work with these anyway - so this seems like a worthy trade-off.

### How to test

Confirm CI passes

### Review checklist

### DIFF
--- a/examples/browser/package.json
+++ b/examples/browser/package.json
@@ -15,14 +15,14 @@
         }
     },
     "dependencies": {
-        "@theia/core": "1.61.0",
-        "@theia/navigator": "1.61.0",
-        "@theia/preferences": "1.61.0",
-        "@theia/terminal": "1.61.0",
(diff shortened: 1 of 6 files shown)


bhufmann (approved):
Looks good to me. Tested on my laptop. Thanks!

PR #1205, opened 2025-06-20
POC: Enable flame graph for user space traces

aaanh:
### What it does

- Adds a modified duplication of `timegraph-output-component.tsx` called `flamegraph-output-component.tsx` for the lack of a better name.
    -  Mostly adjustments in nomenclature for classes and properties
- Adds a GANTT_CHART case to handle the ProviderType in `trace-context-component.tsx` which returns the `FlamegraphOutputComponent`
    - The component takes in a time range of `[0,very big number)` to fetch all the states from the experiment. Currently, it is set to a value: start + end of range
- Adds a new script in `package.json` called `start:electronmon` and the electronmon dependency which will help with development by auto-reloading the Electron application.

### How to test

1. Simply pull this PR branch to local system.
2. Run `yarn` to install the new dependencies
3. Run `yarn start:electronmon` to start the Electron app

### Follow-ups

This is an early stage investigative proof of concept to show that we are able to get flame graph working on Webview leveraging the existing time graph implementation to render the flame graph.

Upcoming works would include, in such order:
- Disconnect the signal handling logic so that the flame graph view is independent from the experiment syncs (e.g. zooming) by implementing a custom UnitController
- Render the x-axis as an arbitrary numerical range instead of timestamps
- Complete separation from the time graph logic and render the data as a weighted tree
- Clean up and refactoring

### Review checklist

### DIFF
--- a/packages/react-components/src/components/trace-context-component.tsx
+++ b/packages/react-components/src/components/trace-context-component.tsx
@@ -758,6 +758,7 @@ export class TraceContextComponent extends React.Component<TraceContextProps, Tr
                                 />
                             );
                         case ProviderType.DATA_TREE:
+                        case 'GANTT_CHART':
                             return (
                                 <DataTreeOutputComponent
                                     key={output.id}
--- a/packages/react-components/src/components/trace-context-component.tsx
+++ b/packages/react-components/src/components/trace-context-component.tsx
@@ -716,12 +716,6 @@ export class TraceContextComponent extends React.Component<TraceContextProps, Tr
                         pinned: this.state.pinnedView ? this.state.pinnedView === output : undefined
                     };
 
-                    const flamegraphRange = new TimeRange(
-                        BigInt(0),
-                        this.state.currentRange.getEnd() + this.state.currentRange.getStart(),
-                        BigInt(0)
-                    );
-
                     switch (responseType) {
                         case 'OVERVIEW':
                             return (
@@ -774,6 +768,11 @@ export class TraceContextComponent extends React.Component<TraceContextProps, Tr
                                 />
                             );
                         case 'GANTT_CHART':
+                            const flamegraphRange = new TimeRange(
+                                BigInt(0),
+                                this.state.currentRange.getEnd() + this.state.currentRange.getStart(),
+                                BigInt(0)
+                            );
                             return (
                                 <FlamegraphOutputComponent
                                     key={output.id}
--- a/packages/react-components/src/components/trace-context-component.tsx
+++ b/packages/react-components/src/components/trace-context-component.tsx
@@ -27,6 +27,7 @@ import { cloneDeep } from 'lodash';
 import { UnitControllerHistoryHandler } from './utils/unit-controller-history-handler';
 import { TraceOverviewComponent } from './trace-overview-component';
 import { TimeRangeUpdatePayload } from 'traceviewer-base/lib/signals/time-range-data-signal-payloads';
+import { AggregatedgraphOutputComponent } from './aggregatedgraph-output-component';
 
 const ResponsiveGridLayout = WidthProvider(Responsive);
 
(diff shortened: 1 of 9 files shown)


packages/react-components/src/components/trace-context-component.tsx
```diff
@@ -180,7 +183,9 @@ export class TraceContextComponent extends React.Component<TraceContextProps, Tr
                 cursorColor: 0x259fd8,
                 lineColor: this.props.backgroundTheme === 'light' ? 0x757575 : 0xbbbbbb
             },
-            backgroundTheme: this.props.backgroundTheme
+            backgroundTheme: this.props.backgroundTheme,
+            flamegraphViewRanges: {},
+            flamegraphResetZoomKey: 0
```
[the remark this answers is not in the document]
aaanh: Yep, same as your comment on the ranges. We'll need to think of a way to handle multiple charts.

packages/react-components/src/components/gantt-chart-output-component.tsx
```diff
@@ -0,0 +1,1712 @@
...
+
+        const strategy = additionalProperties?.filter_query_parameters?.strategy;
+        const ids = rowIds ? rowIds : this.getTimegraphRowIds().rowIds;
+        const { start, end } = range;
+        const newRange: TimelineChart.TimeGraphRange = range;
+        const nbTimes = Math.ceil(Number(end - start) / resolution) + 1;
+        const timeGraphData: TimelineChart.TimeGraphModel = await this.tspDataProvider.getData(
+            ids,
+            this.state.ganttChartTree,
+            fetchArrows,
```
bhufmann, line 1143: can we set it to false for ganttCharts by default?

packages/react-components/src/components/gantt-chart-output-component.tsx
```diff
@@ -0,0 +1,1712 @@
+import * as React from 'react';
```
bhufmann, line 1: this component mostly the same as the timegraph-output-component. Could we abstract it to a common base class and have gantt and regular timegraph extend it?

bhufmann (changes requested):
I have couple more comments.

packages/react-components/src/components/trace-context-component.tsx
```diff
@@ -57,6 +58,8 @@ export interface TraceContextState {
     style: OutputComponentStyle;
     backgroundTheme: string;
     pinnedView: OutputDescriptor | undefined;
+    ganttChartRanges?: Record<string, { start: bigint; end: bigint }>;
+    ganttChartResetZoomKey?: number;
```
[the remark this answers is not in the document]
aaanh: Oh, the comment before was marked as outdated when I changed the name of the file. Below was my reply to it.

I have no solutions that come to mind for this case currently. Maybe a Map<ChartID, Record<T>> to store multiple flame graphs' ranges? If I understand the case correctly.

packages/react-components/src/components/gantt-chart-output-component.tsx
aaanh: Not sure I understand which attribute are you referring to?

aaanh:
I'm thinking we should tackle the multiple zoom states and refactoring to a base component in another PR since it's going to be a large change

aaanh:
Okay, so I have squashed and rebased the branch to consolidate the working tree. Should be cleaner now :sparkles: 

And I'll work on the tsp-typescript-client to get it to be merged and published. After that, we'll update the dependencies and use the new ProviderType.

bhufmann (changes requested):
It works fine. Could you please squash it into one commit with proper commit message? After that it's ready to merge from my point of view.

bhufmann (changes requested):
Please update the commit title: Add support for Flame Graphs

aaanh:
Okay, I have updated the PR description for the follow-up

MatthewKhouzam (approved):
Looks good to me, I approved the refactor too. They didn't break any legacy corner cases I tried.

MatthewKhouzam:
> BTW, I'm ok to merge it as is and use GATNT_CHART

That would be explosive!

PR #1206, opened 2025-06-20
Use TimeGraphContainer forceCanvasRenderer option

marcdumais-work:
### What it does

There's a relatively low limit of WebGL contexts available for a Chrome-based browser tab (16). Each time a "TimeGraphContainer" is created, one such context is consumed. We use these contexts for a few things, not just time-line charts (where it makes most sense).

This commit takes advantage of the new option and declines that a WebGL context be used for containers we create, except for proper timeline chart views and marker containers. This means that we will now be able to create about 8 timeline-chart views per session (each uses 2 WebGL contexts), spread over however many experiments.

~~Note: Depends on the following timeline-chart PR, where the `forceCanvasRenderer` option is added in the `TimeGraphContainerOptions` interface:~~

https://github.com/eclipse-cdt-cloud/timeline-chart/pull/318

update: merged and released

### How to test

The easiest would be to wait until the PR above is merged and released. in the meantime, `yarn link` could be used.

### Review checklist

### DIFF
--- a/packages/react-components/package.json
+++ b/packages/react-components/package.json
@@ -34,7 +34,7 @@
         "react-grid-layout": "1.2.0",
         "react-modal": "^3.8.1",
         "react-virtualized": "^9.21.0",
-        "timeline-chart": "^0.4.1",
+        "timeline-chart": "^0.4.2",
         "traceviewer-base": "0.8.0",
         "tsp-typescript-client": "^0.6.0"
     },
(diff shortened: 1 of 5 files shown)


marcdumais-work:
I wanted to mention that this PR is a first attempt to mitigate the limited `WebGL` contexts that are available, to use used by time-graph containers. There is the possibility of making follow-ups to further mitigate. A couple of ideas:
- keep track of assigned `WebGL` contexts currently used (updated each time one is allocated or freed) by the react-components and switch to requesting software rendering only, pass a certain threshold, such that we avoid running-out no matter how many trace views are opened
- add a preference that permits using only software renderer - this would arguably be preferable for users that expect to regularly bust the number of available WebGL contexts

marcdumais-work:
Note: the license check passed locally - the workflow seems to fail because of too many requests towards `clearlydefined `from GitHub.

bhufmann (approved):
Looks good to me. Thanks!

marcdumais-work:
Thanks for the review!

Issue #133, opened 2020-09-26
Allow time graph rows to have different row heights


bhufmann:
All rows in time graph view have the same height. It shall be possible to have different row heights based on a style configuration provided by the data provider for each row. I would suggest to provide a height factor instead of a absolute height number. Height factor would be optional. If omitted the height factor 1. Some rows could provide height factors (e.g. 0.7) to make the row smaller. 

The height adjustment will need to applied in the tree as well as the timeline chart.

PR #1209, opened 2025-06-27
Fix timegraph misalignment when entry label is blank

### LINKED ISSUE #1208: Timegraph misalignment when entry label is blank
### Bug Description:

When a time graph entry has no label, its tree entry has no height (except padding) because there is no text to render, while the corresponding time graph row has the standard height. This causes misalignment with this and all following entries.

### Steps to Reproduce:

1. Open a kernel trace and open the Resources status view.
2. Scroll to a separator entry between CPUs
3. Observe the misalignment

Image

### Additional Information

- Operating System:
- Theia Version: v0.9.0

PatrickTasse:
### What it does

Set the height for timegraph td in .css stylesheet since the line-height for timegraph tr does not apply if there is no text.

Fixes #1208

### How to test

Open kernel trace and Resources status view. Check that separator rows between CPUs have same height in tree and timegraph.

### Follow-ups

Consider supporting variable height rows. See issue #133.

### Review checklist

### DIFF
--- a/packages/react-components/style/output-components-style.css
+++ b/packages/react-components/style/output-components-style.css
@@ -312,7 +312,6 @@ canvas {
 }
 
 .timegraph-tree tr {
-    /* TODO: Fix row alignment, this number is arbitrary, it works [on my machine], but it should match line height in timeline-chart */
     line-height: 18px;
 }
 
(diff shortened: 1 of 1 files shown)


bhufmann (approved):
It fixes the referenced bug. Thanks for the contribution.

PR #1211, opened 2025-07-02
Add electronmon as the auto-reloader and relevant scripts

aaanh:
### What it does

Current in local development environment, running the examples/electron application is done with `yarn start:electron`. But whenever there are code changes, the application does not automatically reload with the new changes and developers need to manually restart the running instance.

The PR introduces a dev dependency `electronmon` that would watch for changes in directories that match certain patterns. It also updates README and scripts in `package.json` to support this feature.

### How to test

Pull and refresh your dependencies with `yarn`.

Start the monitored electron app with `yarn start:electronmon`

Inspect the developer console in the electron app before any changes then add some `console.log()` statements somewhere, save the file and see the statements being printed after the app is reloaded.

### Follow-ups

There is a slight chance that the electron reload finishes before other running watchers finish recompiling so the changes are not there yet. Workaround it by simply sending a Save command in the code editor which triggers the reload again or <kbd> Ctrl + Shift + P</kbd> in the electron app itself, search for Reload and hit <kbd>Enter</kbd>

### Review checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,6 +148,12 @@ Or, run
 yarn start:electron
 ```
 
+Or, with auto-reloading when code changes are detected
+
+```bash
+yarn start:electronmon
+```
+
 and use the Electron application.
 
 If there are errors that occurred while starting the app, see [Troubleshooting](#troubleshooting) for known issues.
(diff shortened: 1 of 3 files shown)


aaanh:
Closing, as electronmon could be run as a global npm package, no need to be added as a dev dependency

marcdumais-work:
> Closing, as electronmon could be run as a global npm package, no need to be added as a dev dependency

Ok, thanks for giving this a try. In addition, I will point-out that `electronmon` does not seem to be updated much since April 2024, when the last "cluster" of commits were merged. The latest officially supported Node.js version is 18, which is already EoL.

PR #1214, opened 2025-07-16
RFC: Refactor flame graph and gantt chart to extend from a new base gantt component class

aaanh:
### What it does

Add a new BaseGanttOutputComponent which carries most of the functionalities of the existing TImegraph to be the parent class that Timegraph and GanttChart can extend from.

The new BaseGantt has some improvements to make it more generic to use, such as parameterizing IDs and names.

Finally, the Timegraph and GanttChart components extends the BaseGantt and implements their own `renderTree()` method.

Note, `renderTree` is not made abstract due to the specific CSS classes used. Also, it is to maintain some customizabilities for the children classes

### How to test

Check out this PR commit and start local dev with existing instructions in README

### Follow-ups

This would probably need to be merged before starting work on adding the group by props and time range selection features.

### Review checklist


bhufmann (changes requested):
Thanks for the contribution. There seems to be some regression regarding styling. The serach box is now in the middle instead of left aligned. Also the handle of the vertical resize is in the search box and not below. Also the left handle seems to be missing

image

bhufmann:
l

> > > Thanks for the contribution. There seems to be some regression regarding styling. The serach box is now in the middle instead of left aligned. Also the handle of the vertical resize is in the search box and not below. Also the left handle seems to be missing
> > > image
> > 
> > 
> > I think that was already introduced in the parent PR #1205.
> 
> Interesting... I thought it was originally aligned center when I started the flame graph implementation. But now that I checked the current prod vscode extension, it is indeed aligned left/start.

I double-checked it for #1205 and it was left aligned. So, the change to center-aligned was introduced in this version.

aaanh:
> l
> 
> > > > Thanks for the contribution. There seems to be some regression regarding styling. The serach box is now in the middle instead of left aligned. Also the handle of the vertical resize is in the search box and not below. Also the left handle seems to be missing
> > > > image
> > > 
> > > 
> > > I think that was already introduced in the parent PR #1205.
> > 
> > 
> > Interesting... I thought it was originally aligned center when I started the flame graph implementation. But now that I checked the current prod vscode extension, it is indeed aligned left/start.
> 
> I double-checked it for #1205 and it was left aligned. So, the change to center-aligned was introduced in this version.

My latest push should restore the left-alignment for the search bar
image

bhufmann (changes requested):
Please squash the commits into a single commit and remove RFC from the the title.

MatthewKhouzam:
LGTM, didn't break any legacy I saw either.

bhufmann (approved):
LGTM. Thanks!

PR #1233, opened 2025-10-22
Update lerna to version independent for the components

bhufmann:
### What it does

Update lerna to version independent for the components

With this each component will be updated accordingly respecting major, minor and patch. The traceviewer-libs that were added as git subtree have different versions than the vscode-trace-extension components. Using independent version they will increase according to their current version.

Add comment in README.md to update the commit message when releasing.

### How to test

Try a version update (e.g. yarn version:patch) and verify all updated versions in the changed files the release commit.

### Follow-ups

N/A

### Review checklist


marcdumais-work (approved):
LGTM, thanks!

